PR #7, opened 2021-01-26
 Fix show heating state when away

### LINKED ISSUE #6: When the preset mode is away, heat_demand is ignored
When the preset mode of the controller is set to away and a zone is showing heat demand because the temperature is lower then the setpoint, this isn't visible in the climate entity. 

The state of that entity(zone) should not be "auto" but should be "head", as there is heat demand. 
As for the state of the controller is also showing "heat_demand" > 0 , so the state should also reflect that
[home-assistant.log](https://github.com/zxdavb/evohome_cc/files/5871719/home-assistant.log)

phdelodder:
This should fix #6

### DIFF
--- a/custom_components/evohome_cc/climate.py
+++ b/custom_components/evohome_cc/climate.py
@@ -125,6 +125,8 @@ class EvoZone(EvoZoneBase, ClimateEntity):
         if self._evo_device._evo.mode["system_mode"] == "heat_off":
             return HVAC_MODE_OFF
         if self._evo_device._evo.mode["system_mode"] == "away":
+            if self._evo_device.heat_demand is not None and self._evo_device.heat_demand > 0:
+                return HVAC_MODE_HEAT
             return HVAC_MODE_AUTO
 
         if self._evo_device.mode is None:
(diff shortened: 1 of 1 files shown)


custom_components/evohome_cc/climate.py
```diff
@@ -243,7 +245,10 @@ def hvac_mode(self) -> Optional[str]:
             return
         if self._evo_device.mode["system_mode"] == "heat_off":
             return HVAC_MODE_OFF
-        if self._evo_device.mode["system_mode"] == "away":
+        if self.i_evo_device.mode["system_mode"] == "away":
```
zxdavb, line 248: Is `i_evo_device.mode` a typo?

zxdavb:
Have you confused `CURRENT_HVAC_*` with `HVAC_MODE_*`?

custom_components/evohome_cc/climate.py
phdelodder: Yes it's a typo

phdelodder:
Rethinking about it

PR #8, opened 2021-01-31
Show evozone preset modes

phdelodder:


phdelodder:
There is one small issue:

- Set a zone to a different setpoint, the preset is set to temporary
- Set controller to eco
- Set controller back to default (none)
--> the preset mode/value isn't change of the zone that had the different setpoint(temporary preset)

PR #10, opened 2021-02-20
Service calls

phdelodder:


phdelodder:
Currently has one know issue for me is that the until being set by evohome_rf is completely ignored by the controller, resulting in ignoring the complete setpoint request

iMiMx:
Seems to work pretty well for me so far, many thanks - this is'was the main aspect I'm still relying on the internet gateway for, the ability to set a time based override using (evohome_cc.clear_zone_override).

I did notice that the example data seems to be missing, so when you click 'FILL EXAMPLE DATA' on the Developer Tools Services, it doesn't automatically populate defaults for easy testing as the other evohome.* service calls do.

phdelodder:
@iMiMx I have update the services.yaml to be able to use the auto populate the fields with the example button. 

For the setpoint status, I don't know if the info is already available through evohome_rf.

iMiMx:
I think, service_calls might be broken with the merged latest release - for example the below does not set an override:

```
service: evohome_cc.set_zone_override
data:
  entity_id: climate.office_2
  setpoint: 5
  duration:
    minutes: 135
```

I don't have much/any logging enabled by default, will perform some more tests later.

phdelodder:
@iMiMx commit 35a8951 should fix it!

iMiMx:
Is/was there some 'force refresh after change' that might also be missing/broken with recent changes?  I see some commits above for this.

On the earlier 0.5.10 commits the UI seemed to update reasonably quickly, on the latest versions when you set an override using your service_calls (or simply manually make a change to 'temporary' via the UI), it takes some time to update.  With the manual present change, it flicks 'temporary', then back to 'None', before it then changes back to 'Temporary'.

phdelodder:
The refresh(using self._refresh()) still remains the same, it hasn't changed since the previous code. I guess first needs to get this merge so it's being tested by more people

zxdavb (changes requested):
Can you rebase please?

custom_components/evohome_cc/__init__.py
```diff
@@ -158,8 +195,108 @@ async def load_system_config(store) -> Optional[Dict]:
...
+                vol.Optional(ATTR_DURATION_HOURS): vol.All(
+                    cv.time_period,
+                    vol.Range(min=timedelta(hours=0), max=timedelta(hours=24)),
+                ),
+            }
+        )
+        system_mode_schemas.append(schema)
+
+    # These modes are set for a number of days (or indefinitely): use this schema
+    temp_modes = [m for m in SYSTEM_MODE_LOOKUP if m == EVO_MODE_AWAY or m == EVO_MODE_CUSTOM or m == EVO_MODE_DAY_OFF]
```
zxdavb, line 265: Why not:
```
temp_modes = [m for m in SYSTEM_MODE_LOOKUP if m in (EVO_MODE_AWAY, EVO_MODE_CUSTOM, EVO_MODE_DAY_OFF)]
```
or better:
```
temp_modes = (EVO_MODE_AWAY, EVO_MODE_CUSTOM, EVO_MODE_DAY_OFF)
```
... or - as it appears to be a constant;
```
TEMP_MODES = (EVO_MODE_AWAY, EVO_MODE_CUSTOM, EVO_MODE_DAY_OFF)
```

custom_components/evohome_cc/__init__.py
```diff
@@ -158,8 +195,108 @@ async def load_system_config(store) -> Optional[Dict]:
...
+    if temp_modes:  # any of: EVO_MODE_AWAY, EVO_MODE_CUSTOM, EVO_MODE_DAY_OFF, permanent or for 1-99 days
```
zxdavb, line 266: If using `TEMP_MODES = ` (i.e. it's a constant), then the `if TEMP_MODES:` isn't needed at all?

custom_components/evohome_cc/__init__.py
```diff
@@ -158,8 +195,108 @@ async def load_system_config(store) -> Optional[Dict]:
...
+        schema = vol.Schema(
+            {
+                vol.Required(ATTR_SYSTEM_MODE): vol.In(temp_modes),
```
zxdavb, line 269:
```
vol.Required(ATTR_SYSTEM_MODE): vol.In(TEMP_MODES),
```

custom_components/evohome_cc/__init__.py
```diff
@@ -158,8 +195,108 @@ async def load_system_config(store) -> Optional[Dict]:
...
+        }
+
+        async_dispatcher_send(hass, DOMAIN, payload)
+
+    hass.services.async_register(DOMAIN, SVC_REFRESH_SYSTEM, force_refresh)
+
+    system_mode_schemas = []
+
+    # Not all systems support "AutoWithReset": register this handler only if required
+    if [m for m in SYSTEM_MODE_LOOKUP if m == EVO_MODE_RESET]:
```
zxdavb, line 247: I think you've copied this from the **evohome** integration without understanding the nuances?

custom_components/evohome_cc/__init__.py
```diff
@@ -46,6 +56,33 @@
...
+ATTR_ZONE_TEMP = "setpoint"
+ATTR_DURATION_UNTIL = "duration"
+
+SVC_REFRESH_SYSTEM = "refresh_system"
+SVC_SET_SYSTEM_MODE = "set_system_mode"
+SVC_RESET_SYSTEM = "reset_system"
+SVC_SET_ZONE_OVERRIDE = "set_zone_override"
+SVC_RESET_ZONE_OVERRIDE = "clear_zone_override"
+
+from .const import EVO_MODE_AWAY, EVO_MODE_CUSTOM, EVO_MODE_ECO, EVO_MODE_DAY_OFF, EVO_MODE_RESET, EVO_MODE_AUTO, EVO_MODE_HEAT_OFF
```
zxdavb, line 72: These `const`s should come from **evohome_cc.const** _or_ **evohome_rf.const**, but not both, see above.

custom_components/evohome_cc/__init__.py
```diff
@@ -18,16 +18,26 @@
 except (ImportError, ModuleNotFoundError):
     import evohome_rf
 
+from evohome_rf.const import (
+    SYSTEM_MODE_MAP,
+    SYSTEM_MODE_LOOKUP
```
zxdavb, line 23: These consts should come from **evohome_cc.const** or **evohome_rf.const**, but not both, see below.

zxdavb (commented):
Here's a start - you have a few of these:
```
temp_modes = [m for m in SYSTEM_MODE_LOOKUP if m == EVO_MODE_ECO]
```
 - they make sense for **evohome**, but not here - they need to go.

custom_components/evohome_cc/__init__.py
zxdavb: I think I'll refactor to come from **evohome_rf**.

zxdavb:
All this is in 0.6.5 - so, close

PR #20, opened 2021-03-14
Update services.yaml

iMiMx:
Add date format example for until


zxdavb:
I have manually merged these

PR #25, opened 2021-03-24
Small Tweaks

phdelodder:


phdelodder:
There are 2 changes within this proposed PR:

- Change MODE_TO_ZONE to use the MODE_ZONE_TO_HA instead of the PRESET_TCS_TO_HA as otherwise the hvac_modes of a zone are empty and with this it has selectable values. 
- Change in preset_mode is to remove the SystemMode.AUTO, as otherwise it doesn't reach the rest of the code. Thus not setting the correct preset mode

PR #42, opened 2022-06-04
Update sensor.py

janvkem:
Added sensors for 31DA struct

### DIFF
--- a/custom_components/evohome_cc/sensor.py
+++ b/custom_components/evohome_cc/sensor.py
@@ -314,4 +314,49 @@ SENSOR_ATTRS = {
         DEVICE_CLASS: SensorDeviceClass.CO2,
         DEVICE_UNITS: CONCENTRATION_PARTS_PER_MILLION,
     },
+    "exhaust_fan_speed": {
+        DEVICE_UNITS: PERCENTAGE,
+    },
+    "air_quality": {
+        DEVICE_UNITS: PERCENTAGE,
+    },
+    "air_quality_base": {
+        DEVICE_UNITS: PERCENTAGE,
+    },
+    "outdoor_humidity": {
+        DEVICE_CLASS: SensorDeviceClass.HUMIDITY,
+        DEVICE_UNITS: PERCENTAGE,
+    },
+    "exhaust_temperature": {
+        DEVICE_CLASS: SensorDeviceClass.TEMPERATURE,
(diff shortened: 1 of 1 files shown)


zxdavb:
> Next up is sending commands :) Could you explain to me simply where the send_packet command as a service is called from and where in evohome_cc and ramses_rf the commands are handled? 

You shoudl not be using `send_packet`. You can use a setter (preferred) if one exists:
https://github.com/zxdavb/ramses_cc/blob/d36ca1629295d68f6a1db37b3c8d580afb482558/custom_components/evohome_cc/climate.py#L259-L262

... or a service (see below). It starts with **services.yaml**, for example:
https://github.com/zxdavb/ramses_cc/blob/d36ca1629295d68f6a1db37b3c8d580afb482558/custom_components/evohome_cc/services.yaml#L91-L111

Then you can look at the **`register_service_functions()`** in **\_\_init\_\_.py** or in (say) **climate.py** (you'll use **fan.py**):
https://github.com/zxdavb/ramses_cc/blob/d36ca1629295d68f6a1db37b3c8d580afb482558/custom_components/evohome_cc/climate.py#L142-L143
...and...
https://github.com/zxdavb/ramses_cc/blob/d36ca1629295d68f6a1db37b3c8d580afb482558/custom_components/evohome_cc/climate.py#L290-L299

zxdavb:
BTW, if you want to test/dev bespoke commands, use the **client.py** tool:
```
python client.py monitor /dev/ttyACM0 -x "RQ 37:123456 2411 0000FF"
```

zxdavb:
Surely some of these should be in **fan.py**? Maybe start with them in **sensor.py** and move them over at some stage? See what you think.

custom_components/evohome_cc/sensor.py
```diff
@@ -314,4 +314,57 @@ async def async_update(self) -> None:
         DEVICE_CLASS: SensorDeviceClass.CO2,
         DEVICE_UNITS: CONCENTRATION_PARTS_PER_MILLION,
     },
+    "exhaust_fan_speed": {
```
zxdavb, line 317: Please import the constants from **ramses_rf** into **const.py** and import into here.

zxdavb (changes requested):

zxdavb:
some of the above has been implemented - others are waiting for a change in the schema

PR #44, opened 2022-06-16
Attempt to add servicecalls and move all into fan.py

janvkem:


zxdavb:
@janvkem Thanks for your work... There's a couple of things you have to make a permanent part of your workflow:
 - de-linting (using tools like black, isort, etc.
 - rebasing (if you submits a PR, but I subsequently change the code base)

janvkem:
> @janvkem Thanks for your work... There's a couple of things you have to make a permanent part of your workflow:
> 
>     * de-linting (using tools like black, isort, etc.
> 
>     * rebasing (if you submits a PR, but I subsequently change the code base)

Thanks @zxdavb. I'll look into the de-linting and will pay better attention to the rebasing.

PR #89, opened 2023-12-04
Apply HA linter suggestions and neaten up some registrations

trvrnrth:
I'm interested in adding Config Flow support so have begun investigation by setting up a dev environment and working through the HA linter warnings.

The changes here also include some minor neatening up of entity/service registrations to more closely follow the conventions in most of the core integrations with a view to splitting out the schemas in a follow-up PR.

PR #90, opened 2023-12-04
Lint suggestions and ruff isort settings

trvrnrth:
This applies suggestions from #89 and adopts the isort config from HA core.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,10 @@
 
 
 [tool.ruff.lint.isort]
-  combine-as-imports = true  # default is false
-
+  force-sort-within-sections = true
+  known-first-party = [
+      "homeassistant",
+  ]
+  combine-as-imports = true
+  split-on-trailing-comma = false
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -43,6 +43,10 @@
 
 
 [tool.ruff.lint.isort]
-  combine-as-imports = true  # default is false
-
+  force-sort-within-sections = true
+  known-first-party = [
+      "homeassistant",
+  ]
```
trvrnrth, line 49: I'm not sure if this is a good idea or not. I can't find any generally accepted convention for custom components.

zxdavb:
If you have time, here is a challenge for you. 

These two bits of code do the same thing:
https://github.com/zxdavb/ramses_cc/blob/1cdb2499516682026747d6d8b220f82439d3307a/custom_components/ramses_cc/__init__.py#L63-L68

... and:
https://github.com/zxdavb/ramses_cc/blob/1cdb2499516682026747d6d8b220f82439d3307a/custom_components/ramses_cc/coordinator.py#L111-L113

I think the former is best practice.

You need to get the co-ordinator squared away before you add config flow.

PR #92, opened 2023-12-07
Service and entity registration refactor

trvrnrth:
This splits out the service schemas into their respective platforms and refactors the entity registration and controller/broker/coordinator update handing.

⚠️ I have not done any extensive testing of this so there are bound to be issues, but I wanted to get a PR created for initial review ASAP.


custom_components/ramses_cc/climate_heat.py
trvrnrth, line 1: Merged into `climate.py` for simplicity to avoid circular dependencies.

custom_components/ramses_cc/climate_hvac.py
trvrnrth, line 1: Merged into `climate.py` for simplicity to avoid circular dependencies.

custom_components/ramses_cc/__init__.py
```diff
@@ -52,258 +90,337 @@
...
     @callback
-    def async_handle_dispatch(self, *args) -> None:  # TODO: remove as unneeded?
-        """Process a dispatched message.
-
-        Data validation is not required, it will have been done upstream.
-        This routine is threadsafe.
-        """
-        if not args:
-            self.update_ha_state()
+    def async_register_platform(
```
trvrnrth, line 365: This can probably be simplified once config entries are in place as I believe it will then be possible to await the platform setup before starting the controller.

custom_components/ramses_cc/__init__.py
```diff
@@ -52,258 +90,337 @@
...
+        for system in self._client.systems:
+            rf_entities.append(system)
+            if isinstance(system, MultiZone):
+                rf_entities.extend([z for z in system.zones if z.name])
+
+        have_new_entities = False
+        for rf_entity in rf_entities:
+            unique_id = f"{rf_entity.__class__.__module__}.{rf_entity.__class__.__qualname__}|{rf_entity.id}"
+            if unique_id in self._rf_entity_ids:
+                async_dispatcher_send(self._hass, f"RAMSES_RF_UPDATE_{rf_entity.id}")
```
trvrnrth, line 355: This uses an entity specific message as it may be nice in future to be able to check a last-updated timestamp here to limit the work done on each call. That would be a complicated endeavour though so for now we always trigger updates.

custom_components/ramses_cc/__init__.py
```diff
@@ -52,258 +90,337 @@
...
 
-        func(*args, **kwargs)
-        self.update_ha_state()
+        rf_entities: list[RamsesRFEntity] = []
+        for device in self._client.devices:
+            rf_entities.append(device)
+        for system in self._client.systems:
+            rf_entities.append(system)
+            if isinstance(system, MultiZone):
+                rf_entities.extend([z for z in system.zones if z.name])
```
trvrnrth, line 349: This defers zone entity creation until name discovery is complete to ensure that the HA entity is first created with an appropriate entity ID.

zxdavb:
Trevor, thanks for your work!

It is a massive PR - is there any chance you could break it into smaller pieces that are a little easier to fit my head around?

custom_components/ramses_cc/manifest.json
```diff
@@ -3,11 +3,20 @@
...
+    "version": "0.0.2",
+    "config_flow": true,
+    "usb": [
+      {
+        "vid": "1b4f",
+        "pid": "9206",
+        "manufacturer": "SparkFun",
+        "description": "evofw3 atmega32u4"
+      }
+    ]
```
zxdavb: Can we leave config flow until the other refactoring (e.g. co-ordinator) is fully squared away?

custom_components/ramses_cc/water_heater.py
```diff
@@ -93,145 +185,118 @@ class EvohomeDHW(EvohomeZoneBase, WaterHeaterEntity):
...
-        self.svc_set_dhw_mode(
+        await self.async_set_dhw_mode(
             mode=MODE_HA_TO_EVO[operation_mode], active=active, until=until
         )
 
-    def set_temperature(self, temperature: float = None, **kwargs) -> None:
+    async def async_set_temperature(self, temperature: float = None, **kwargs) -> None:
         """Set the target temperature of the water heater."""
-        self.svc_set_dhw_params(setpoint=temperature)
+        self.async_set_dhw_params(setpoint=temperature)
```
zxdavb, line 257: why make it `async def` rather than `def`? 

Have a look at `@callback` decorator, I think.

custom_components/ramses_cc/__init__.py
zxdavb: IIRC, as long as it has a unique_id, then the entity_id can be set as required - I can't recall the api off the top of my head.

If we wait for the zone name, and the packet isn't available, then there is an entity that exists, but won't be registered.

You can't use the zone name as the unique_id, because zones can be renamed.

custom_components/ramses_cc/__init__.py
```diff
@@ -109,7 +109,7 @@ class RamsesController:
 
     _port_name: str
     _client_config: dict
-    _client: Gateway
+    _client: Gateway | None
```
zxdavb, line 112: if this is never None, other than at instantiation, and never subsequently tested for None-ness, if woudl leave it as is, and set a mypy type hint

zxdavb:
OK...

I am not an expert on writing HA integrations (although I have written at least 4), so don't be afraid to question it, it you feel I have miss-advised you.

All these are OK:
 - merging heat/hvac into climate
 - pulling all the constants into const.py
 - renaming service methods to `async_svc_*` (it doesn't mean the code is async, just that the code is safe to run in the main thread)
 - pulling schemas/service registration into platforms

All of these could be separate/smaller PRs.

Why did you want to stop using the service dicts, like `SVCS_SENSOR`? Unless there's a need, I'd be inclined to leave it the way it was.

Can we leave config flow until last? Need to get the co-ordinator settled first.

I am reluctant to say more - it is simply too big a PR - I cannot get my head around it. The HA team, for example, would always reject such a large PR.

Other than that, is there a specific question you want an answer too?

custom_components/ramses_cc/manifest.json
trvrnrth: Oops. I didn't mean to include this here. Consider it a sneak preview!

custom_components/ramses_cc/water_heater.py
trvrnrth: Thanks, I'll look into that a bit more closely.

custom_components/ramses_cc/__init__.py
trvrnrth: The problem is that for the climate entities the entity ID is implicit and generated from the name at first creation so you can end up with a zone with an entity ID reflective of the zone ID rather than the more desirable zone name. It's something that can be addressed/discussed separately.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -89,13 +92,17 @@ def entity_factory(broker, device, attr, *, entity_class=None, **kwargs):
...
+
+    platform.async_register_entity_service(
+        SVC_PUT_PRESENCE_DETECT,
+        cv.make_entity_service_schema(
+            {
+                vol.Required(SZ_PRESENCE_DETECTED): cv.boolean,
+            },
+            "svc_put_presence_detect",
+        ),
+    )
```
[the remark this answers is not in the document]
trvrnrth: With the final changes in the PR each platform is set up once on first init and then registers a callback with the controller to handle adding new ramses_rf entities. This should work well with the flow required to set up config entries and matches the approach taken in, for example, the core zwave components.

custom_components/ramses_cc/climate.py
```diff
@@ -35,26 +78,104 @@ def entity_factory(entity_class, broker, device):  # TODO: deprecate
...
+                    cv.positive_float,
+                    vol.Range(min=5, max=21),
+                ),
+                vol.Optional(CONF_LOCAL_OVERRIDE, default=True): cv.boolean,
+                vol.Optional(CONF_OPENWINDOW, default=True): cv.boolean,
+                vol.Optional(CONF_MULTIROOM, default=True): cv.boolean,
+            }
+        ),
+        "svc_set_zone_config",
+    )
```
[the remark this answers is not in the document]
trvrnrth: It could go either way really, but this appeared to be the more common approach in the core components and I liked the simple flexibility it offered to toggle/adjust service availability in future depending on config.

custom_components/ramses_cc/__init__.py
trvrnrth: The client is only created later in `_create_client`. It's tested for None-ness in `_async_create_update_tasks` so that we only create the update tasks once all platforms have registered *and* we have created a client.

trvrnrth:
Thanks for the comments, they are helpful. It sounds like there are no great concerns with the overall approach so I'll start spinning out smaller PRs and we can discuss the specifics in isolation. I'll close out this PR now as it has served it's purpose for broad initial review.

PR #93, opened 2023-12-08
Remove timedelta alias

trvrnrth:
This is now consistent within the component and with core.


zxdavb (approved):

zxdavb:
I think HA also keep `datetime` instead of `datetime as dt` (if you're willing to submit another PR).

PR #96, opened 2023-12-08
Remove unused fan platform

trvrnrth:


zxdavb:
Yes, let's remove - will one day replace remotes with fans, but not in the short term.

PR #99, opened 2023-12-08
Remove entity domain migration

trvrnrth:
The move to version 0.30 seems like an appropriate time to remove this.


zxdavb:
spot on

PR #102, opened 2023-12-08
Rename water heater entity class

trvrnrth:
This conforms to the core component naming conventions.

### DIFF
--- a/custom_components/ramses_cc/water_heater.py
+++ b/custom_components/ramses_cc/water_heater.py
@@ -70,7 +70,10 @@ async def async_setup_platform(
     broker = hass.data[DOMAIN][BROKER]
 
     async_add_entities(
-        [entity_factory(EvohomeDHW, broker, dhw) for dhw in discovery_info["dhw"]]
+        [
+            entity_factory(RamsesWaterHeater, broker, dhw)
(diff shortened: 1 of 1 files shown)


custom_components/ramses_cc/water_heater.py
```diff
@@ -82,7 +85,7 @@ def entity_factory(entity_class, broker, device):  # TODO: deprecate
             platform.async_register_entity_service(name, schema, f"svc_{name}")
 
 
-class EvohomeDHW(EvohomeZoneBase, WaterHeaterEntity):
+class RamsesWaterHeater(EvohomeZoneBase, WaterHeaterEntity):
```
zxdavb, line 88: May as well call this **RamsesZoneBase** too.

zxdavb (changes requested):

custom_components/ramses_cc/water_heater.py
trvrnrth: Yep. I was going to come back to a few other renames once #94 is merged.

zxdavb: ok, lets merge this for now

PR #104, opened 2023-12-10
Tidy entity classes

trvrnrth:
This is a first pass at some more tidy up of the entity classes.


custom_components/ramses_cc/__init__.py
```diff
@@ -162,47 +162,25 @@ async def svc_send_packet(call: ServiceCall) -> None:
...
 
     @property
     def extra_state_attributes(self) -> dict[str, Any]:
         """Return the integration-specific state attributes."""
-        attrs = {
-            a: getattr(self._device, a)
-            for a in self._entity_state_attrs
-            if hasattr(self._device, a)
+        return {
+            ATTR_ID: self.rf_entity.id,
```
trvrnrth, line 182: This seemed like a useful attribute for all entities to have. `_entity_state_attrs` was unused.

custom_components/ramses_cc/__init__.py
```diff
@@ -264,44 +242,11 @@ def __init__(
...
-        """Return the current temperature."""
-        return self._device.temperature
-
-    @property
-    def extra_state_attributes(self) -> dict[str, Any]:
-        """Return the integration-specific state attributes."""
-        return {
-            **super().extra_state_attributes,
-            "schema": self._device.schema,
-            "params": self._device.params,
```
trvrnrth, line 301: The relevant config/params are exposed separately by the relevant entities so this was redundant and somewhat confusing. Technically a breaking change, but it would seem unlikely that anyone is relying on it.

custom_components/ramses_cc/__init__.py
```diff
@@ -264,44 +242,11 @@ def __init__(
...
-    def current_temperature(self) -> float | None:
-        """Return the current temperature."""
-        return self._device.temperature
-
-    @property
-    def extra_state_attributes(self) -> dict[str, Any]:
-        """Return the integration-specific state attributes."""
-        return {
-            **super().extra_state_attributes,
-            "schema": self._device.schema,
```
trvrnrth, line 300: This seemed slightly redundant with the existence of the schema binary sensor which includes the full system schema. Happy to add back the zone-level fragment though if it's considered useful.

custom_components/ramses_cc/__init__.py
```diff
@@ -264,44 +242,11 @@ def __init__(
...
 
     @property
     def name(self) -> str:
         """Return the name of the binary_sensor/sensor."""
-        if not hasattr(self._device, "name") or not self._device.name:
-            return f"{self._device.id} {self._state_attr}"
-        return f"{self._device.name} {self._state_attr}"
-
-
-class RamsesZoneBase(RamsesEntity):  # for: climate & water_heater
```
trvrnrth, line 277: This is almost entirely unused so it seemed most appropriate to just move anything relevant into the individual entity classes as it made things easier to understand.

custom_components/ramses_cc/__init__.py
```diff
@@ -264,44 +242,11 @@ def __init__(
...
-    def extra_state_attributes(self) -> dict[str, Any]:
-        """Return the integration-specific state attributes."""
-        return {
-            **super().extra_state_attributes,
-            "schema": self._device.schema,
-            "params": self._device.params,
-        }
-
-    @property
-    def name(self) -> str | None:
```
trvrnrth, line 305: For now this is copy-pasted into a couple of places but will be removed in future when device info is added at base entity level as the entity names will then just become `None` to reflect the fact they are the primary entity for the HA device.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -108,20 +108,13 @@ def __init__(
...
-
-        _LOGGER.info("Found a Binary Sensor for %s: %s", device.id, state_attr)
-
-        super().__init__(
-            broker,
-            device,
-            state_attr,
-            device_class=device_class,
-        )
+        _LOGGER.info("Found %r: %s", device, state_attr)
```
trvrnrth, line 111: I've shortened these logs and standardised on use of RF entity repr as the logger name already indicates type.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -96,32 +99,25 @@ def entity_factory(broker, device, attr, *, entity_class=None, **kwargs):
...
-
-        _LOGGER.info("Found a Binary Sensor for %s: %s", device.id, state_attr)
-
-        super().__init__(
-            broker,
-            device,
-            state_attr,
-            device_class=device_class,
-        )
+        _LOGGER.info("Found %r: %s", rf_entity, state_attr)
```
trvrnrth, line 114: I've shortened these logs and standardised on use of RF entity repr as the logger name already indicates type.

zxdavb:
@trvrnrth Please do what you can to keep your PRs as small as possible, whilst remaining atomic.

Have a look at: https://developers.home-assistant.io/docs/review-process/#creating-the-perfect-pr

For example, your renamed the `_device` attr to `rf_device`. That can easily be a separate PR.

I can turn over a larger number of smaller PRs much faster that small numbers of large PRs (assuming they make the same changes).

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -152,22 +147,22 @@ class RamsesFaultLog(RamsesBinarySensor):
...
     def extra_state_attributes(self) -> dict[str, Any]:
         """Return the integration-specific state attributes."""
-        return {
-            "active_fault": self._device.tcs.active_fault,
-            "latest_event": self._device.tcs.latest_event,
-            "latest_fault": self._device.tcs.latest_fault,
+        return super().extra_state_attributes() | {
+            "active_fault": self.rf_entity.tcs.active_fault,
+            "latest_event": self.rf_entity.tcs.latest_event,
+            "latest_fault": self.rf_entity.tcs.latest_fault,
```
zxdavb, line 159: I am afraid I don't see the value in renaming this attr to `rf_entity`. 

In addition, it should also start with an underscore as it is not for HA, only for self.

I suggest leaving it as `_device`.

custom_components/ramses_cc/__init__.py
```diff
@@ -241,67 +219,34 @@ def update_ha_state(self, delay=3) -> None:
...
-class RamsesDeviceBase(RamsesEntity):  # for: binary_sensor & sensor
-    """Base for any RAMSES II-compatible entity (e.g. BinarySensor, Sensor)."""
+class RamsesSensorBase(RamsesEntity):
+    """Base for any Ramses sensor entity."""
 
     def __init__(
         self,
         broker,
-        device,
+        rf_entity: RamsesRFEntity,
```
zxdavb, line 228: I think we can leave this as `device`. The word `entity` is very HA, and so I avoid using that term if I can.

custom_components/ramses_cc/__init__.py
zxdavb: OK, can go - I'm not fussed either way (we'll see if there are complaints, soon enough)

Ideally, this data should not go to the state machine, and should be extracted via a service call. This is equally true for the system schema.

custom_components/ramses_cc/__init__.py
```diff
@@ -7,16 +7,16 @@
 import logging
 from typing import Any
 
+from ramses_rf.entity_base import Entity as RamsesRFEntity
```
zxdavb, line 10: This (**`Entity`**) isn't likely to be stable... that part of **ramses_rf** is a real mess. OK for now, but something to think about in future.

custom_components/ramses_cc/__init__.py
zxdavb: Nice one.

custom_components/ramses_cc/__init__.py
zxdavb: Yep.

custom_components/ramses_cc/__init__.py
zxdavb: Not sure - I'd need to have a think. 

Leave in for now.

custom_components/ramses_cc/__init__.py
zxdavb: OK, just beware some (climate) entities can be renamed.

custom_components/ramses_cc/binary_sensor.py
zxdavb: OK.

custom_components/ramses_cc/climate.py
```diff
@@ -209,11 +214,11 @@ def hvac_action(self) -> str | None:
...
         """Return the Controller's current operating mode of a Controller."""
 
-        if self._device.system_mode is None:
+        if self.rf_entity.system_mode is None:
             return  # unable to determine
-        if self._device.system_mode[CONF_SYSTEM_MODE] == SystemMode.HEAT_OFF:
+        if self.rf_entity.system_mode[CONF_SYSTEM_MODE] == SystemMode.HEAT_OFF:
             return HVACMode.OFF
-        if self._device.system_mode[CONF_SYSTEM_MODE] == SystemMode.AWAY:
+        if self.rf_entity.system_mode[CONF_SYSTEM_MODE] == SystemMode.AWAY:
```
zxdavb, line 221: I'll have a look at this file, after `rf_entity` is reverted back to old name - will make thsi PR much smaller.

zxdavb (commented):
Pease rename `rf_entity` back to `_device`; this will make it much easier to review this PR.

trvrnrth:
> @trvrnrth Please do what you can to keep your PRs as small as possible, whilst remaining atomic.
> 
> Have a look at: https://developers.home-assistant.io/docs/review-process/#creating-the-perfect-pr
> 
> For example, your renamed the `_device` attr to `rf_device`. That can easily be a separate PR.
> 
> I can turn over a larger number of smaller PRs much faster that small numbers of large PRs (assuming they make the same changes).

As I mentioned in the PR description I did split this into two distinct commits to make it easier to review. I can submit the two commits as separate PRs if you prefer.

zxdavb:
> As I mentioned in the PR description I did split this into two distinct commits to make it easier to review. I can submit the two commits as separate PRs if you prefer.

Yes please.

custom_components/ramses_cc/__init__.py
trvrnrth: OK, noted. It was this naming that suggested to me that rf_entity would make sense, but if it's subject to change anyway then leaving it as _device makes as much sense as anything else.

custom_components/ramses_cc/__init__.py
trvrnrth: That should be fine, but will of course need testing properly once the HA devices are in place.

custom_components/ramses_cc/sensor.py
```diff
@@ -225,42 +218,6 @@ def extra_state_attributes(self) -> dict[str, Any]:
         return attrs
 
 
-class RamsesFaultLog(RamsesDeviceBase):
```
trvrnrth, line 228: Removed as unused. I assume the binary sensor will be the only place this is exposed.

custom_components/ramses_cc/climate.py
```diff
@@ -510,13 +520,6 @@ def current_temperature(self) -> float | None:
         """Return the current temperature."""
         return self._device.indoor_temp
 
-    @property
-    def extra_state_attributes(self) -> dict[str, Any]:
```
trvrnrth, line 514: Note that I haven't added `params` here as the attribute wasn't previously present (RamsesHvac did not extend RamsesZoneBase).

custom_components/ramses_cc/climate.py
```diff
@@ -287,31 +293,40 @@ def svc_set_system_mode(self, mode, period=None, days=None) -> None:
...
+    def current_temperature(self) -> float | None:
+        """Return the current temperature."""
+        return self._device.temperature
+
     @property
     def extra_state_attributes(self) -> dict[str, Any]:
         """Return the integration-specific state attributes."""
-        return {
+        return super().extra_state_attributes | {
+            "params": self._device.params,
```
trvrnrth, line 325: I'll raise a separate PR for discussion on removal of this.

custom_components/ramses_cc/climate.py
```diff
@@ -287,31 +293,40 @@ def svc_set_system_mode(self, mode, period=None, days=None) -> None:
...
-class RamsesZone(RamsesZoneBase, ClimateEntity):
-    """Base for a Honeywell TCS Zone."""
+class RamsesZone(RamsesEntity, ClimateEntity):
+    """Representation of a Ramses zone."""
+
+    _device: Zone
 
     _attr_icon: str = "mdi:radiator"
     _attr_hvac_modes: list[str] = list(MODE_TO_ZONE)
+    _attr_precision: PRECISION_TENTHS
```
zxdavb, line 303: The precision is 0.01, not 0.1.

People like this greater precision, and will complain if it's removed.

I would leave this 0.01 for now, then we'll make it optional via **configuration.yaml**.

custom_components/ramses_cc/climate.py
```diff
@@ -287,31 +293,40 @@ def svc_set_system_mode(self, mode, period=None, days=None) -> None:
...
     _attr_supported_features: int = (
         ClimateEntityFeature.PRESET_MODE | ClimateEntityFeature.TARGET_TEMPERATURE
     )
     _attr_target_temperature_step: float = PRECISION_TENTHS
+    _attr_temperature_unit: str = UnitOfTemperature.CELSIUS
 
     def __init__(self, broker, device) -> None:
-        """Initialize a TCS Zone."""
-        _LOGGER.info("Found a Zone: %r", device)
+        """Initialize a Ramses zone."""
```
zxdavb, line 312: Leave as 'TCS zone'.

custom_components/ramses_cc/climate.py
zxdavb: This is only because HVAC hasn't gotten the love it deserves - OK for now.

custom_components/ramses_cc/sensor.py
zxdavb: This needs sorting, one day.

zxdavb (commented):
See comments - there is an rf_entry left over.

Make sure you check you don't need to rebase.

custom_components/ramses_cc/__init__.py
```diff
@@ -241,8 +219,8 @@ def update_ha_state(self, delay=3) -> None:
         )  # HACK: call_soon_threadsafe should not be needed
 
 
-class RamsesDeviceBase(RamsesEntity):  # for: binary_sensor & sensor
-    """Base for any RAMSES II-compatible entity (e.g. BinarySensor, Sensor)."""
+class RamsesSensorBase(RamsesEntity):
+    """Base for any Ramses sensor entity."""
```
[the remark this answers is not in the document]
zxdavb: """Base for any Ramses sensor/binary_sensor entity."""

custom_components/ramses_cc/water_heater.py
```diff
@@ -127,6 +128,11 @@ def is_away_mode_on(self) -> bool | None:
         except TypeError:
             return
 
+    @property
+    def name(self) -> str | None:
+        """Return the name of the zone."""
+        return self.rf_entity.name
```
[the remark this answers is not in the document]
trvrnrth: Fixed.

custom_components/ramses_cc/__init__.py
```diff
@@ -162,47 +162,25 @@ async def svc_send_packet(call: ServiceCall) -> None:
...
         self.hass = broker.hass
         self._broker = broker
         self._device = device
 
-        self._attr_should_poll = False
-
-        self._entity_state_attrs = ()
-
-        # NOTE: this is bad: self.update_ha_state(delay=5)
+        self._attr_unique_id = device.id
```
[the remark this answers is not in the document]
trvrnrth: This is to retain the existing IDs per the current implementation for [climate](https://github.com/zxdavb/ramses_cc/blob/3aa5f06ab9d0e759918801297c62753f6b5759be/custom_components/ramses_cc/__init__.py#L288), [water heater ](https://github.com/zxdavb/ramses_cc/blob/3aa5f06ab9d0e759918801297c62753f6b5759be/custom_components/ramses_cc/water_heater.py#L92) and [remote](https://github.com/zxdavb/ramses_cc/blob/3aa5f06ab9d0e759918801297c62753f6b5759be/custom_components/ramses_cc/remote.py#L79).

custom_components/ramses_cc/__init__.py
```diff
@@ -162,47 +162,25 @@ async def svc_send_packet(call: ServiceCall) -> None:
...
     @property
     def extra_state_attributes(self) -> dict[str, Any]:
         """Return the integration-specific state attributes."""
-        attrs = {
-            a: getattr(self._device, a)
-            for a in self._entity_state_attrs
-            if hasattr(self._device, a)
+        return {
+            ATTR_ID: self._device.id,
         }
```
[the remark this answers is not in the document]
trvrnrth: `_entity_state_attrs` is never set, so this previously always returned an empty dict. The child entity classes handle the extra params as required and it didn't seem like there was any real need to change that.

custom_components/ramses_cc/climate.py
trvrnrth: I copied this [as-is from RamsesZoneBase](https://github.com/zxdavb/ramses_cc/blob/3aa5f06ab9d0e759918801297c62753f6b5759be/custom_components/ramses_cc/__init__.py#L280).

custom_components/ramses_cc/__init__.py
trvrnrth: I haven't checked yet what will be required to update the associated device name.

custom_components/ramses_cc/__init__.py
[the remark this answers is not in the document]
trvrnrth: I'm not sure what would be involved in a migration to a better set of unique IDs so left it alone for now.

custom_components/ramses_cc/__init__.py
zxdavb: I've had complaints that **ramses_cc** doesn't handle renaming of zones very well.

It would be great to start writing tests for this at some stage. Sigh... there is a lot to do.

trvrnrth: Would I be correct in guessing that the complaints are that the entity ID is not updated to reflect the new name?

zxdavb:
OK, I have change base to **refactor**

custom_components/ramses_cc/__init__.py
zxdavb: Yes.

trvrnrth: I believe it would be easy enough to provide a fixed entity ID as is [done for the sensors already](https://github.com/zxdavb/ramses_cc/blob/7fcf48ae23a7c5073addbfd25fd606c061ef3457/custom_components/ramses_cc/__init__.py#L259).

custom_components/ramses_cc/__init__.py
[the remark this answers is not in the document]
zxdavb: You probably already know this...

The unique ids should be a function of a fixed id, such as the device's id, e.g. `04:123456`.

Thus, for sensors: device_id (any device) and the corresponding attr. For remotes: device_id.

Finally, the 'child' entities:
 - the TCS controller (unique ID is device_id, `01_123456`), 
 - the zones (device_id, zone_idx, e.g. `01_123456_02`), and 
 - the domains (controller device_id + domain_id, e.g. `01:123456_FC`)

in future, there'll be other 'child' entities, such as UFH controller circuits.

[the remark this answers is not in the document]
trvrnrth: I'm fuzzy on the domains but this should be doing everything else you describe.

custom_components/ramses_cc/__init__.py
zxdavb: I am not sure where we are with this, but I feel the `entity_id` should be the zone name.

That way, you can destroy/recreate your zone (it may get a new `zone_idx`, and thus a new `unique_id`). But if it has the same `entity_id`, then you don't have to re-write your automations.

custom_components/ramses_cc/__init__.py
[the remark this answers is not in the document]
zxdavb: Domains are a bit hard to explain - they 'called' for heat (or cold) by the TCS.
```python
DOMAIN_TYPE_MAP: dict[str, str] = {
    F6: "cooling_valve",  # cooling
    F7: "domain_f7",
    F8: "domain_f8",
    F9: DEV_ROLE_MAP[DevRole.HT1],  # Heating Valve
    FA: DEV_ROLE_MAP[DevRole.HTG],  # HW Valve (or UFH loop if src.type == UFC?)
    FB: "domain_fb",  # also: cooling valve?
    FC: DEV_ROLE_MAP[DevRole.APP],  # appliance_control
    FD: "domain_fd",  # seen with hometronics
    # "FE": ???
    # FF: "system",  # TODO: remove this, is not a domain
}  # "21": "Ventilation", "88": ???
```

We can ignore them for now.

custom_components/ramses_cc/__init__.py
trvrnrth: At the moment the entity id will be implicitly generated from the entity name. This is fine for the most part but has some issues:
- if discovery does not find the zone name before the entity is created the entity id will reflect the fallback name
- if the zone name is changed the entity id is not regenerated

So on balance it is probably better to explicitly provide an entity id, but I'll approach that separately as I didn't really want to change any behaviour in this PR.

custom_components/ramses_cc/climate.py
trvrnrth: My understanding was that the precision increase was from 0.5 as is "normally" available to 0.1 (ie. PRECISION_TENTHS rather than PRECISION_HALVES).

zxdavb: HA never allowed 0.01 for climate, water_heater... SO OK for that. But aim to keep for sensors.

We have a separate branch now - we can be more adventurous.

PR #105, opened 2023-12-11
Simplify entity state saves

trvrnrth:
- Removes unused dispatch handlers.
- Tidies up dispatch of update signal.
- Unwrap device function calls so that we can make use of type hinting. We now explicitly call to write state afterwards.


custom_components/ramses_cc/__init__.py
```diff
@@ -207,38 +206,16 @@ def extra_state_attributes(self) -> dict[str, Any]:
...
-        Data validation is not required, it will have been done upstream.
-        This routine is threadsafe.
-        """
-        if not args:
-            self.update_ha_state()
+        self.async_on_remove(
+            async_dispatcher_connect(
+                self.hass, SIGNAL_UPDATE, self.async_write_ha_state
+            )
+        )
```
zxdavb, line 213: Curious why this is added - **ramses_cc** (and **ramses_rf**) has no concept of removing devices (i.e. device_id, TCS or zone), only discovering new devices.

custom_components/ramses_cc/__init__.py
```diff
@@ -207,38 +206,16 @@ def extra_state_attributes(self) -> dict[str, Any]:
     async def async_added_to_hass(self) -> None:
         """Run when entity about to be added to hass."""
         self._broker._entities[self.unique_id] = self
-        async_dispatcher_connect(self.hass, DOMAIN, self.async_handle_dispatch)
```
zxdavb, line 210: Is it right to remove this?

custom_components/ramses_cc/__init__.py
```diff
@@ -207,38 +206,16 @@ def extra_state_attributes(self) -> dict[str, Any]:
...
     @callback
-    def update_ha_state(self, delay=3) -> None:
-        """Update HA state after a short delay to allow system to quiesce.
-
-        This routine is threadsafe.
-        """
-        args = (delay, self.async_schedule_update_ha_state)
-        self.hass.loop.call_soon_threadsafe(
-            self.hass.helpers.event.async_call_later, *args
-        )  # HACK: call_soon_threadsafe should not be needed
```
zxdavb, line 241: So, IIRC, _should_ not be needed, but _is_ (well, was)... Have you tested this?

We have no test suite. :(

custom_components/ramses_cc/climate.py
```diff
@@ -425,22 +408,25 @@ def svc_put_zone_temp(
         self._device.sensor._make_fake()
         self._device.sensor.temperature = temperature
         self._device._get_temp()
-        self.update_ha_state()
+        self.async_write_ha_state()
```
zxdavb, line 411: So, `async_write_ha_state_delayed()` after a read, and `async_write_ha_state()` after a write?

Note that the new version of **ramses_rf** can invoke async API calls (have a look at **src/ramses_tx/command.py** for these APIs), and wait for the result (i.e. send a RQ & wait for the RP, send an W & wait for the I).

Before (0.2x.x), these calls were simply cast.

custom_components/ramses_cc/coordinator.py
```diff
@@ -294,9 +294,5 @@ async def async_update(self, *args, **kwargs) -> None:
...
             self.hass.helpers.event.async_call_later(5, self.async_save_client_state)
 
-        self.async_dispatcher_send()
-
-    def async_dispatcher_send(self, *args, **kwargs) -> None:
-        """Inform the devices that their state data may have changed."""
-
-        async_dispatcher_send(self.hass, DOMAIN, *args, **kwargs)
+        # Trigger state updates of all entities
+        async_dispatcher_send(self.hass, SIGNAL_UPDATE)
```
zxdavb, line 298: I can't recall if the new entities do an update after instantiation - IIRC this is configurable.

Other than that, when do we need to do this? E.g. look at name of zones - it just pulls the attr from the underlying **ramses_rf** entity when required.

custom_components/ramses_cc/water_heater.py
```diff
@@ -132,20 +132,6 @@ def target_temperature(self) -> float | None:
...
-    @callback
-    def async_handle_dispatch(self, *args: Any) -> None:
-        """Process a service request (system mode) for a controller."""
-        if not args:
-            self.update_ha_state()
-            return
-
-        payload = args[0]
-        if payload.get(UNIQUE_ID) != self.unique_id:
-            return
```
zxdavb, line 144: Why did I put this in, I wonder...

zxdavb (commented):
Nice.

You may need to rebase.

custom_components/ramses_cc/__init__.py
trvrnrth: It's replaced with the more explicit `SIGNAL_UPDATE`.

custom_components/ramses_cc/__init__.py
trvrnrth: As you say, entity removal is unlikely but it still seemed like it was worth putting the handling in place to remove the update listener just in case.

custom_components/ramses_cc/coordinator.py
trvrnrth: This will trigger the calls to `async_write_ha_state` which I understood to be necessary as the entities are not set up to be polled (ie. `_attr_should_poll` is false) so they must push state to HA. Please correct me if I'm wrong!

zxdavb:
I've just had to add this hack to **__init__.py**:
```python
    def _call_client_api(self, func, *args, **kwargs) -> None:
        """Wrap client APIs to make them threadsafe."""
        self.hass.loop.call_soon_threadsafe(
            partial(func, *args, **kwargs)
        )  # HACK: call_soon_threadsafe should not be needed, but is!

        # func(*args, **kwargs)
        self.update_ha_state()
```

trvrnrth:
> ```python
> _call_client_api
> ```

I don't believe anything that has been merged down has touched that area. What was the issue?

zxdavb:
```
2023-12-13 14:26:25.119 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [139713416868160] Non-thread-safe operation invoked on an event loop other than the current one
Traceback (most recent call last):
  File "/home/dbonnes/clients/hass/homeassistant/components/websocket_api/commands.py", line 238, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/hass/homeassistant/core.py", line 2133, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/home/dbonnes/clients/hass/homeassistant/core.py", line 2170, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/hass/homeassistant/helpers/entity_component.py", line 272, in handle_service
    return await service.entity_service_call(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/hass/homeassistant/helpers/service.py", line 880, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/hass/homeassistant/helpers/service.py", line 950, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/home/dbonnes/clients/hass/homeassistant/components/climate/__init__.py", line 508, in async_set_hvac_mode
    await self.hass.async_add_executor_job(self.set_hvac_mode, hvac_mode)
  File "/usr/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/ramses_cc/custom_components/ramses_cc/climate.py", line 399, in set_hvac_mode
    self.svc_reset_zone_mode()
  File "/home/dbonnes/clients/ramses_cc/custom_components/ramses_cc/climate.py", line 439, in svc_reset_zone_mode
    self._call_client_api(self._device.reset_mode)
  File "/home/dbonnes/clients/ramses_cc/custom_components/ramses_cc/__init__.py", line 221, in _call_client_api
    func(*args, **kwargs)
  File "/home/dbonnes/clients/ramses_rf/src/ramses_rf/system/zones.py", line 737, in reset_mode
    return self.set_mode(mode=ZON_MODE_MAP.FOLLOW)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/ramses_rf/src/ramses_rf/system/zones.py", line 751, in set_mode
    return self._send_cmd(cmd)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/ramses_rf/src/ramses_rf/entity_base.py", line 144, in _send_cmd
    return self._gwy.send_cmd(cmd)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dbonnes/clients/ramses_rf/src/ramses_rf/gateway.py", line 818, in send_cmd
    task = self._loop.create_task(self.async_send_cmd(cmd, callback=callback))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 436, in create_task
    task = tasks.Task(coro, loop=self, name=name, context=context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 763, in call_soon
    self._check_thread()
  File "/usr/lib/python3.11/asyncio/base_events.py", line 800, in _check_thread
    raise RuntimeError(
RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one
```

zxdavb:
Hang on, there are other issues...

zxdavb:
@trvrnrth I think we need to create your own branch for this work, so I can work on stabilising **ramses_cc** without introducing bugs from your refactoring.

ASAP, I will look to move current **master** to **stable**, then only do bug fixes to stable. 

I will create a branch called **refactor** for you, can look to merge that into **master** when we have a new **stable**.

OK?

zxdavb:
I'm not blaming your work - the issue is where you're starting from (i.e. my code).

trvrnrth:
> @trvrnrth I think we need to create your own branch for this work, so I can work on stabilising the new **ramses_cc** without introducing bugs from your refactoring.
> 
> ASAP, I will look to move current **master** to **stable**, then only do bug fixes to stable.
> 
> I will create a branch called **refactor** for you, can look to merge that into **master** when we have a new **stable**.
> 
> OK?

Sounds fine to me.

zxdavb:
OK, I have change base to **refactor**

custom_components/ramses_cc/climate.py
zxdavb: Yes leave to wait - that reasoning isn't completely sound.

Invoking a command on a faked remote (which, BTW is technically stateless - other faked devices may not be, e.g. actuator) will cause the USB dongle to cast a packet on its behalf.

That `RQ`, `W` or `I` will 

Of course, this isn't a faked device, it's a zone, but teh same applies in a way.

With the new transport, which has QoS, you will be able to update without a delay. The QoS code is there, but is disabled by default.

If you want to play with it:
```yaml
ramses_cc:
  ramses_rf:
    disable_qos: false
```

custom_components/ramses_cc/coordinator.py
zxdavb: OMG - I think I was/am polling, IIRC (oeriodic call of aync_update is scheduled). But it could move to fully event-driven with callbacks from **ramses_rf** - that wont happen soon.

trvrnrth: Entity polling is [currently disabled in master](https://github.com/zxdavb/ramses_cc/blob/7fcf48ae23a7c5073addbfd25fd606c061ef3457/custom_components/ramses_cc/__init__.py#L190) so they will only save out their state when the broker `async_update` is called on the `CONF_SCAN_INTERVAL` schedule by way of listening for the `DOMAIN` message to be dispatched.

I haven't intentionally changed the fundamentals of this setup in this PR so if there's anything here that is not maintaining the same point of saving state as at present it was not deliberate.

I had the same thought about events from ramses_rf but also figured it wouldn't be coming soon.

trvrnrth:
> RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one

I guess the question there is how the RF Gateway/Engine  is ending up with a different event loop as I thought it was [passed through on init](https://github.com/zxdavb/ramses_cc/blob/7fcf48ae23a7c5073addbfd25fd606c061ef3457/custom_components/ramses_cc/coordinator.py#L135C37-L135C41).

zxdavb:
> > RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one
> 
> I guess the question there is how the RF Gateway/Engine is ending up with a different event loop as I thought it was [passed through on init](https://github.com/zxdavb/ramses_cc/blob/7fcf48ae23a7c5073addbfd25fd606c061ef3457/custom_components/ramses_cc/coordinator.py#L135C37-L135C41).

HA runs multiple threads. So my understanding is was a call made from a different thread?

I am not sure - I am not a python programmer ad its are area I was never 100% confident in. 

FWIW, there are a few `asyncio.get_running_loop()` scattered in there.

trvrnrth:
> HA runs multiple threads. So my understanding is was a call made from a different thread?
Well that likely makes sense then if the service call thread/loop differs from the one at time of Gateway/Engine init. I'll do some digging around and see about coming up with a neat way to deal with it.

>I am not a python programmer
Me either!

custom_components/ramses_cc/__init__.py
zxdavb: So, this is the first update after adding the entity.

custom_components/ramses_cc/climate.py
zxdavb: Yep - and async is best!

zxdavb:
If you rebase, looks like we can merge?

zxdavb (approved):

trvrnrth:
> If you rebase, looks like we can merge?

Sounds good. We'll need to check/revisit the threadsafe wrapper later but I need to do some proper reading about that first and any re-addition is probably going to be easier to review in isolation.

PR #106, opened 2023-12-14
Tidy up constants

trvrnrth:


custom_components/ramses_cc/water_heater.py
```diff
@@ -96,15 +96,15 @@ def __init__(self, broker, device) -> None:
...
         try:
-            mode = self._device.mode[CONF_MODE]
+            mode = self._device.mode[SZ_MODE]
         except TypeError:
             return
         if mode == ZoneMode.SCHEDULE:
             return STATE_AUTO
         elif mode == ZoneMode.PERMANENT:
-            return STATE_ON if self._device.mode[CONF_ACTIVE] else STATE_OFF
+            return STATE_ON if self._device.mode["active"] else STATE_OFF
```
trvrnrth, line 105: I couldn't find an appropriate constant to import from the library for this.

custom_components/ramses_cc/__init__.py
```diff
@@ -27,15 +26,17 @@
...
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator
 
-from .const import BROKER, DOMAIN, SIGNAL_UPDATE
-from .coordinator import RamsesBroker
-from .schemas import (
-    SCH_DOMAIN_CONFIG,
+from .const import (
+    BROKER,
+    CONF_ADVANCED_FEATURES,
+    CONF_MESSAGE_EVENTS,
```
zxdavb, line 32: So, my practice is for constants that are strings to start with `SZ_`.

HA does use `CONF_` for strings used in config (and voluptuous schemas), or `ATTR_`, etc.

So when should it be `CONF_*` (or `ATTR_`), and when it be `SZ_*`?

You seem to have it about right.

custom_components/ramses_cc/water_heater.py
zxdavb: This doesn't look right anyway - is `_device` a 
 - **BdrSwitch** in `ramses_rf.device.heat`, or
 - **StoredHw** in `ramses_rf.system.heat`

zxdavb: Woops - its **DhwZone** in `ramses_rf.system.zones`!
```python
if payload[2:4] != "FF":
    result["active"] = {"00": False, "01": True, "FF": None}[payload[2:4]]
```

zxdavb: So, I've added:
```python
SZ_ACTIVE: Final[str]  = "active"
```

to `ramses_tx.const` (although that is no good for you today.

zxdavb: Perhaps change **manifest.json** to point to the master branch of zxdavb/ramses_rf (something like):
```json
"requirements": [ "git+https://github.com/zxdavb/ramses_rf" ],
```

zxdavb (approved):
Looks good!

zxdavb:
Woops - should be `SZ_ACTIVE`!

custom_components/ramses_cc/water_heater.py
trvrnrth: I've got ramses_rf installed from master locally for testing anyway so all good :)

PR #109, opened 2023-12-15
Tidy up climate mode/preset maps

trvrnrth:
These definitions are now more long-form but easier to understand.

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -34,61 +34,70 @@ from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from homeassistant.helpers.typing import ConfigType, DiscoveryInfoType
 
 from . import RamsesEntity
-from .const import BROKER, DOMAIN, SystemMode, ZoneMode
+from .const import (
+    BROKER,
+    DOMAIN,
+    PRESET_CUSTOM,
+    PRESET_PERMANENT,
+    PRESET_TEMPORARY,
+    SystemMode,
+    ZoneMode,
+)
 from .coordinator import RamsesBroker
 from .schemas import SVCS_CLIMATE_EVO_TCS, SVCS_CLIMATE_EVO_ZONE
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


custom_components/ramses_cc/climate.py
```diff
@@ -34,61 +34,70 @@
...
-    ZoneMode.TEMPORARY: "temporary",
-    ZoneMode.PERMANENT: "permanent",
+    ZoneMode.TEMPORARY: PRESET_TEMPORARY,
+    ZoneMode.PERMANENT: PRESET_PERMANENT,
+}
+PRESET_TO_ZONE = {
+    PRESET_NONE: ZoneMode.SCHEDULE,
+    PRESET_TEMPORARY: ZoneMode.TEMPORARY,
+    PRESET_PERMANENT: ZoneMode.PERMANENT,
 }
```
zxdavb, line 100: How about:
```
PRESET_HA_TO_ZONE = {v: k for k, v in PRESET_ZONE_TO_HA.items()}
```

zxdavb: Ah! It's below!

I think leave it this way - it says clearly that there is a 1-1 mapping, both ways, no?

zxdavb (changes requested):

custom_components/ramses_cc/climate.py
trvrnrth: I guess that makes sense for this one since it actually is a straight mapping.

I'm not sure if it was intentional in your suggestion, but I also think that `PRESET_HA_TO_ZONE` is a clearer name than `PRESET_TO_ZONE` so will change that as well for all of them.

zxdavb: I am of the opinion that this mapping is wrong in any case: Zones should come up as HA **Auto** if they are in **FollowSchedule**, and - IIRC - they don't.

trvrnrth: These are the presets. If you look a bit further up then schedule does map to auto for the mode.

Issue #70, opened 2023-11-05
Cannot set system mode with period or duration


raldred:
I am unable to use the `set_system_mode` service with any of the addition options the service defines with my evohome system.
If I provide duration or period, I receive an error and the service call fails.
Additionally I can only get it working for eco_boost, all the other modes do nothing.

HA version: 2023.10.3
Ramses_cc version: 0.21.40

I can use the standard climate.set_preset_mode service but this does not allow you to set non default periods/durations.

Using the native set_system_mode like so...
eg,
```
tap_action:
  action: call-service
  service: ramses_cc.set_system_mode
  data:
    mode: eco_boost
    duration: 120
    entity_id: climate.controller
```
This returns an error
<img width="628" alt="Screenshot 2023-11-05 at 15 05 57" src="https://github.com/zxdavb/ramses_cc/assets/37096/21330d19-420a-4b89-86b5-cf473eb0baa5">

If omit duration, it will set the system mode to eco for the rest of the day.  
But other modes don't seems to be settable, for example:

```
action: call-service
  service: ramses_cc.set_system_mode
  data:
    mode: away
    period: 2
    entity_id: climate.controller
```
This does nothing, no error, system mode remains in current mode

raldred:
Fixing in #110

PR #110, opened 2023-12-15
Fix system mode period/duration args

trvrnrth:
I believe this is the intended behaviour based on [services.yaml](https://github.com/zxdavb/ramses_cc/blob/7fcf48ae23a7c5073addbfd25fd606c061ef3457/custom_components/ramses_cc/services.yaml#L150).

The naming of the arguments seems a bit odd to me, so perhaps a future candidate for deprecation/renaming?


zxdavb (commented):
is this closing an issue? #70

PR #111, opened 2023-12-16
Tidy up service constants

trvrnrth:
Moves a couple of name constants missed in #106 and standardises on `SCHEDULE` rather than `SCHED` for all constants.


zxdavb (approved):

zxdavb:
thanks!

PR #112, opened 2023-12-16
Fix mode change service descriptions

trvrnrth:
This is the likely source of the payload issues in #70

### DIFF
--- a/custom_components/ramses_cc/schemas.py
+++ b/custom_components/ramses_cc/schemas.py
@@ -93,7 +93,7 @@ SVC_SET_SYSTEM_MODE = "set_system_mode"
 
 SCH_SYSTEM_MODE = _SCH_ENTITY_ID.extend(
     {
-        vol.Required(CONF_MODE): vol.In(SYSTEM_MODE_LOOKUP),  # incl. DAY_OFF_ECO
+        vol.Required(CONF_MODE): vol.In(SYSTEM_MODE_LOOKUP),
     }
 )
 SCH_SYSTEM_MODE_HOURS = _SCH_ENTITY_ID.extend(
@@ -107,7 +107,12 @@ SCH_SYSTEM_MODE_HOURS = _SCH_ENTITY_ID.extend(
 SCH_SYSTEM_MODE_DAYS = _SCH_ENTITY_ID.extend(
     {
         vol.Required(CONF_MODE): vol.In(
-            [SystemMode.AWAY, SystemMode.CUSTOM, SystemMode.DAY_OFF]
+            [
+                SystemMode.AWAY,
+                SystemMode.CUSTOM,
+                SystemMode.DAY_OFF,
+                SystemMode.DAY_OFF_ECO,
+            ]
         ),
         vol.Optional(CONF_DURATION_DAYS, default=timedelta(days=0)): vol.All(
             cv.time_period, vol.Range(min=timedelta(days=0), max=timedelta(days=99))
--- a/custom_components/ramses_cc/services.yaml
+++ b/custom_components/ramses_cc/services.yaml
@@ -129,37 +129,32 @@ set_system_mode:
             - eco_boost
             - away
             - day_off
+            - day_off_eco
             - heat_off
             - custom
 
     period:
-      name: Days
+      name: Period (days)
       description: >-
-        Optional. A period of time in days; used only with away, day_off, or custom.
+        Optional. A period of time in days; used only with away, day_off, day_off_eco or custom.
         The system will revert to auto at midnight (up to 99 days, today is day 1).
-      default: 1
+      default:
+        days: 28
       example: '{"days": 28}'
       selector:
-        number:
-          min: 1
-          max: 99
-          step: 1
-          unit_of_measurement: days
-          mode: slider
+        duration:
+          enable_day: true
 
     duration:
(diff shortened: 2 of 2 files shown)


custom_components/ramses_cc/services.yaml
```diff
@@ -133,33 +133,27 @@ set_system_mode:
...
       description: >-
-        Optional. A period of time in days; used only with away, day_off, or custom.
+        Optional. A period of time in days; used only with away, day_off, day_off_eco or custom.
         The system will revert to auto at midnight (up to 99 days, today is day 1).
-      default: 1
+      default:
+        days: 28
       example: '{"days": 28}'
       selector:
-        number:
```
trvrnrth, line 143: As the service schema uses the standard time period schema for this field, any number entered will be interpreted in seconds not days. Rather than change the schema and potentially break existing calls which take advantage of this it seemed safer to instead switch to the duration UI selector (even if the hours/minutes/seconds are ignored). 

An alternative to this would be the `object` selector which may be preferable as that may make it clearer that only days are accepted?

custom_components/ramses_cc/services.yaml
```diff
@@ -133,33 +133,27 @@ set_system_mode:
...
       description: >-
         Optional. The duration in hours/minutes; used only with eco_boost (up to 24h).
-      default: 60
-      example: '{"hours": 18}'
+      default:
+        hours: 1
+        minutes: 30
+      example: '{"hours": 1, "minutes": 30}'
       selector:
-        number:
```
trvrnrth, line 157: Same reasoning as above for period.

custom_components/ramses_cc/services.yaml
```diff
@@ -133,33 +133,27 @@ set_system_mode:
...
-          max: 99
-          step: 1
-          unit_of_measurement: days
-          mode: slider
+        duration:
+          enable_day: true
 
     duration:
-      name: Eco/Boost Duration
+      name: Duration (hours/minutes)
```
trvrnrth, line 148: Renamed to make it clearer what this field is when switching between the UI/YAML and referencing descriptions.

custom_components/ramses_cc/services.yaml
```diff
@@ -133,33 +133,27 @@ set_system_mode:
             - custom
 
     period:
-      name: Days
+      name: Period (days)
```
trvrnrth, line 136: Renamed to make it clearer what this field is when switching between the UI/YAML and referencing descriptions.

custom_components/ramses_cc/schemas.py
```diff
@@ -107,7 +107,12 @@
 SCH_SYSTEM_MODE_DAYS = _SCH_ENTITY_ID.extend(
     {
         vol.Required(CONF_MODE): vol.In(
-            [SystemMode.AWAY, SystemMode.CUSTOM, SystemMode.DAY_OFF]
+            [
+                SystemMode.AWAY,
+                SystemMode.CUSTOM,
+                SystemMode.DAY_OFF,
+                SystemMode.DAY_OFF_ECO,
```
trvrnrth, line 114: I am making an assumption that this was not intentionally omitted here.

custom_components/ramses_cc/services.yaml
```diff
@@ -443,23 +444,23 @@ set_dhw_mode:
         boolean:
 
     duration:
-      name: Override duration
+      name: Duration
```
trvrnrth, line 447: Renamed so the references in the description regarding mutual exclusivity make sense.

custom_components/ramses_cc/services.yaml
```diff
@@ -443,23 +444,23 @@ set_dhw_mode:
...
-          min: 15
-          max: 1440
-          step: 15
-          unit_of_measurement: mins
-          mode: slider
+        duration:
 
     until:
-      name: Override end time
+      name: Until
```
trvrnrth, line 458: Renamed so the references in the description regarding mutual exclusivity make sense.

custom_components/ramses_cc/services.yaml
```diff
@@ -349,7 +350,7 @@ get_dhw_schedule:
 
   fields:
     entity_id: &entity_id_dhw
-      name: Stored HW
+      name: Stored DHW
```
trvrnrth, line 353: Renamed for consistency.

zxdavb (changes requested):

custom_components/ramses_cc/services.yaml
trvrnrth: > I am very willing to make breaking changes, if that is for the greater good.

It would probably make most sense to have the same `duration` and `until` fields here as for `set_zone_mode`. Perhaps something for the refactor branch? Keeping backwards compat for `period` probably wouldn't be too hard either but it sounds like it's perhaps not worth the effort.

> I could never get that UI selector to work - have you tested it? I think we shoudl use it if we can.

Seemed fine to me. If there are fewer limitations in the underlying system it probably makes perfect sense to expose full flexibility anyway.

zxdavb: Yes - keep breaking changes minimal for **master** - I am working towards making it the new **stable** branch (dodgy workflow).

But be happy to do the 'right' thing for **refactor** - even if breaking change (keep a track of them).

PR #114, opened 2023-12-17
Merge master to refactor

trvrnrth:
Not sure how you'd prefer to handle these merges.

I imagine there are likely to be more changes in this area but getting the two branches in sync first seemed wise.

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -213,12 +213,11 @@ class RamsesEntity(Entity):
     # @callback  # TODO: WIP
     def _call_client_api(self, func, *args, **kwargs) -> None:
         """Wrap client APIs to make them threadsafe."""
-        
+
         self.hass.loop.call_soon_threadsafe(
             partial(func, *args, **kwargs)
         )  # HACK: call_soon_threadsafe should not be needed, but is!
 
-        # func(*args, **kwargs)
         self.update_ha_state()
 
     @callback
--- a/custom_components/ramses_cc/manifest.json
+++ b/custom_components/ramses_cc/manifest.json
@@ -3,11 +3,11 @@
     "name": "RAMSES RF",
     "documentation": "https://github.com/zxdavb/ramses_cc",
     "requirements": [
-      "ramses-rf==0.30.5",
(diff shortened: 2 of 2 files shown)


custom_components/ramses_cc/services.yaml
```diff
@@ -302,14 +298,19 @@ set_zone_mode:
           mode: slider
 
     duration:
+      name: Duration
```
zxdavb, line 301: Shall it be `Duration (hours/minutes)` too?

custom_components/ramses_cc/services.yaml
```diff
@@ -443,23 +444,23 @@ set_dhw_mode:
         boolean:
 
     duration:
-      name: Override duration
+      name: Duration
```
zxdavb, line 447: See above.

custom_components/ramses_cc/services.yaml
```diff
@@ -443,23 +444,23 @@ set_dhw_mode:
...
-          min: 15
-          max: 1440
-          step: 15
-          unit_of_measurement: mins
-          mode: slider
+        duration:
 
     until:
-      name: Override end time
+      name: Until
```
zxdavb, line 458: `Until (datetime)` ?

zxdavb:
> Not sure how you'd prefer to handle these merges.

We can be adventurous with **refactor**, but keeping sync'd is desirable.

zxdavb:
I am aware my review is not internally consistent,

zxdavb (approved):

custom_components/ramses_cc/services.yaml
trvrnrth: I didn't change this one before as there isn't the peculiarity of having two separate fields that do more or less the same thing for this service.

trvrnrth: From what you were saying is it also the case that the underlying will actually respect durations to the second as well? I haven't checked.

custom_components/ramses_cc/services.yaml
trvrnrth: I was figuring that the selector and example were already doing do a good enough job of making it clear.

PR #115, opened 2023-12-18
Move service definitions into their respective modules

trvrnrth:
It's worth noting that I have left the register calls separate rather than iterating over a map for each platform as there are services which are candidates for supporting a response, or requiring features. Keeping the calls separate will allow for these enhancements to made more easily later without the need for a more complicated interstitial data structure.


custom_components/ramses_cc/climate.py
```diff
@@ -104,39 +227,56 @@ async def async_setup_platform(
...
+        )
+        platform.async_register_entity_service(
+            SVC_SET_ZONE_MODE, SVC_SET_ZONE_MODE_SCHEMA, "svc_set_zone_mode"
+        )
+        platform.async_register_entity_service(
+            SVC_RESET_ZONE_MODE, {}, "svc_reset_zone_mode"
+        )
+        platform.async_register_entity_service(
+            SVC_GET_ZONE_SCHEDULE, {}, "svc_get_zone_schedule"
+        )
```
trvrnrth, line 260: This service is a candidate for supporting a response in future.

custom_components/ramses_cc/water_heater.py
```diff
@@ -50,29 +117,51 @@ async def async_setup_platform(
...
+        )
+        platform.async_register_entity_service(
+            SVC_SET_DHW_PARAMS, SVC_SET_DHW_PARAMS_SCHEMA, "svc_set_dhw_params"
+        )
+        platform.async_register_entity_service(
+            SVC_RESET_DHW_PARAMS, {}, "svc_reset_dhw_params"
+        )
+        platform.async_register_entity_service(
+            SVC_GET_DHW_SCHEDULE, {}, "svc_get_dhw_schedule"
+        )
```
trvrnrth, line 147: This service is a candidate for supporting a response in future.

custom_components/ramses_cc/remote.py
```diff
@@ -37,23 +70,30 @@ async def async_setup_platform(
...
+            SVC_SEND_COMMAND, SVC_SEND_COMMAND_SCHEMA, "async_send_command"
+        )
+        platform.async_register_entity_service(
+            SVC_DELETE_COMMAND, SVC_DELETE_COMMAND_SCHEMA, "async_delete_command"
+        )
 
-    for name, schema in SVCS_REMOTE.items():
-        platform.async_register_entity_service(name, schema, f"svc_{name}")
+    if discovery_info is None:
+        return
```
zxdavb, line 91: Shouldn't this (`discovery_info`) be at the top, so the services are not registered, unless a remote exists?

zxdavb (changes requested):

custom_components/ramses_cc/remote.py
trvrnrth: I believe the check is redundant as the platform setup only gets called by the broker if there is discovery info available. But I thought I'd leave complete removal of it until I PR changes relating to entity addition.

zxdavb: IIRC, it has something to do with config flow too.

trvrnrth: For config flow, the entry gets forwarded to each platform. If we want to maintain only creating services if we have supported devices (which would be nicer) then the approach taken in the core zwave JS component will probably work quite nicely.

zxdavb: > I believe the check is redundant as the platform setup only gets called by the broker if there is discovery info available. 

IIRC, when I submitted a PR to HA (I wrote the **evohome** integration, among others), a core dev had insisted I put it back... But, TBH, I can't be sure.

It will cost us nothing to leave it in, please (I am willing to be talked around, if you can e.g. show me an existing integration that is missing it).

I note new entity detection in in the broker/co-ordinator. I often wondered if it would be better to do that in the platform... probably not.

> If we want to maintain only creating services if we have supported devices (which would be nicer)

I agree - we *should not* create a service unless there is an entity for it to run against. That means, that some service for a platform will be registered, and other not.

> then the approach taken in the core zwave JS component will probably work quite nicely.

I have no capacity to research that atm. I would be pleased to accept a PR from you.

trvrnrth: I'll just shift the check back to the start for this PR; it will be more relevant to talk about it in the context of later changes and as you say does no harm for now.

trvrnrth: > IIRC, when I submitted a PR to HA (I wrote the **evohome** integration, among others), a core dev had insisted I put it back... But, TBH, I can't be sure.

For reference: https://github.com/home-assistant/core/pull/26042#discussion_r319737004 which I suppose makes sense whilst we're still using discovery.

zxdavb: > For reference: https://github.com/home-assistant/core/pull/26042#discussion_r319737004 which I suppose...

You found it!

custom_components/ramses_cc/remote.py
```diff
@@ -37,23 +70,30 @@ async def async_setup_platform(
...
-    platform = entity_platform.async_get_current_platform()
+    if not broker._services.get(PLATFORM):
+        broker._services[PLATFORM] = True
 
-    for name, schema in SVCS_REMOTE.items():
-        platform.async_register_entity_service(name, schema, f"svc_{name}")
+        platform = entity_platform.async_get_current_platform()
+
+        platform.async_register_entity_service(
+            SVC_LEARN_COMMAND, SVC_LEARN_COMMAND_SCHEMA, "async_learn_command"
```
zxdavb, line 84: I like consistency...

This service is called `async_*`, but elsewhere they're called `svc_*`.

Make them all one or the other?

custom_components/ramses_cc/schemas.py
```diff
@@ -26,313 +25,22 @@
...
 
-#
-# Generic services for Integration/domain
-SCH_FAKE_DEVICE = vol.Schema(
-    {
-        vol.Required(SZ_DEVICE_ID): cv.matches_regex(r"^[0-9]{2}:[0-9]{6}$"),
-        vol.Optional("create_device", default=False): vol.Any(None, cv.boolean),
-        vol.Optional("start_binding", default=False): vol.Any(None, cv.boolean),
-    }
-)
```
zxdavb, line 100: I'm happy to see this go, but I thought this was so elegant:
```python
services = {k: v for k, v in locals().items() if k.startswith("svc")}
```

Regardless, why not leave all the schema definitions in **schemas.py** - seems tidier to me.

Not a biggie - up to you.

zxdavb (changes requested):

custom_components/ramses_cc/remote.py
trvrnrth: These are necessarily `async_` as they are what the core service will call and I removed the redundant wrapper. I did consider renaming the others to match so will go that route.

custom_components/ramses_cc/schemas.py
trvrnrth: I'd have perhaps been inclined to leave it, but this did not seem so elegant:
```
if not broker.config[SZ_ADVANCED_FEATURES].get(SVC_SEND_PACKET):
        del domain_service[SVC_SEND_PACKET]
```

I favoured the "don't make me think" approach and split them out as it was easier to follow that way.

custom_components/ramses_cc/climate.py
```diff
@@ -104,39 +227,56 @@ async def async_setup_platform(
...
     new_entities = []
 
     if discovery_info.get("fans"):
-        if not broker._services.get(f"{PLATFORM}_hvac"):
-            broker._services[f"{PLATFORM}_hvac"] = True
-            # for name, schema in SVCS_CLIMATE_HVAC.items():
-            #     platform.async_register_entity_service(name, schema, f"svc_{name}")
-
         for fan in discovery_info["fans"]:
             new_entities.append(RamsesHvac(broker, fan))
```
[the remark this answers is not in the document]
trvrnrth: I'll adjust this in a separate PR that addresses entity creation more specifically if that's OK.

custom_components/ramses_cc/climate.py
```diff
@@ -104,39 +227,56 @@ async def async_setup_platform(
...
-            for name, schema in SVCS_CLIMATE_EVO_ZONE.items():
-                platform.async_register_entity_service(name, schema, f"svc_{name}")
-
         for tcs in discovery_info.get("ctls", []):
-            new_entities.append(entity_factory(RamsesController, broker, tcs))
+            new_entities.append(RamsesController(broker, tcs))
 
         for zone in discovery_info.get("zons", []):
-            new_entities.append(entity_factory(RamsesZone, broker, zone))
+            new_entities.append(RamsesZone(broker, zone))
```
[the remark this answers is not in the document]
trvrnrth: Will address in a separate PR. I should have left these lines alone entirely here really as they're not relevant to the service changes.

custom_components/ramses_cc/schemas.py
trvrnrth: On the note of `schemas.py` when looking through the core components from what I could see the convention was to have the schemas defined either inline or as a const in the same module so it seemed fairly sensible to follow suit. I suspect it will probably also make more sense later when the main config is fitted into a config flow as it may need to be split up a bit.

custom_components/ramses_cc/remote.py
trvrnrth: Renamed. It reads better this way for the calls that do not originate from a service call so worth doing.

zxdavb (approved):

PR #116, opened 2023-12-19
Tidy up broker

trvrnrth:
Tidies up broker start and polling.

Removes redundant DataUpdateCoordinator as the broker already handles updates on an interval. This is more suited to a single API call that returns data and is intended for use with entities that extend `CoordinatorEntity`. Without such entities in place there are no listeners registered with the coordinator and it never does any polling. It seems like a better idea to stick with the existing broker setup as ideally in future it would be possible to transition to listening to event-based updates from ramses_rf which the DataUpdateCoordinator framework is not suited to.

### DIFF
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -1,4 +1,4 @@
-"""Coordinator for RAMSES integration."""
+"""Broker for RAMSES integration."""
 from __future__ import annotations
 
 from datetime import datetime as dt, timedelta
@@ -22,6 +22,7 @@ from homeassistant.const import CONF_SCAN_INTERVAL, Platform
 from homeassistant.core import HomeAssistant, callback
 from homeassistant.helpers.discovery import async_load_platform
 from homeassistant.helpers.dispatcher import async_dispatcher_send
+from homeassistant.helpers.event import async_call_later, async_track_time_interval
 from homeassistant.helpers.storage import Store
 from homeassistant.helpers.typing import ConfigType
 
@@ -47,11 +48,11 @@ class RamsesBroker:
         self._store: Store = hass.helpers.storage.Store(STORAGE_VERSION, STORAGE_KEY)
 
         self.hass_config = hass_config
+        _LOGGER.debug("Config = %s", hass_config)
         self._ser_name, self._client_config, self.config = normalise_config(
             hass_config[DOMAIN]
         )
 
-        # self.status = None
         self.client: Gateway = None  # type: ignore[assignment]
         self._remotes: dict[str, str] = None  # type: ignore[assignment]
 
@@ -67,8 +68,6 @@ class RamsesBroker:
         self._rems = []
         self._zons = []
 
-        self.loop_task = None
-        self._last_update = dt.min
         self._sem = Semaphore(value=1)
 
         self.learn_device_id = None  # TODO: remove me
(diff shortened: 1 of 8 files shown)


custom_components/ramses_cc/broker.py
```diff
@@ -129,7 +124,7 @@ def _create_client(
         if not schema_is_minimal(config_schema):  # move this logic into ramses_rf?
             _LOGGER.warning("The config schema is not minimal (consider minimising it)")
 
-        if cached_schema and (merged := merge_schemas(cached_schema, config_schema)):
+        if cached_schema and (merged := merge_schemas(config_schema, cached_schema)):
```
trvrnrth, line 127: These were switched which I imagine was causing issues with the cached schema taking precedence over the config schema.

custom_components/ramses_cc/broker.py
```diff
@@ -143,33 +138,28 @@ def _create_client(
...
+            and pkt[41:45] not in msg_code_filter
+        }
+
     async def _async_load_storage(self) -> dict:
         """May return an empty dict."""
 
         app_storage = await self._store.async_load()  # return None if no store
         return app_storage or {}
 
-    async def _async_load_client_state(self) -> None:
```
trvrnrth, line 152: This was unused.

custom_components/ramses_cc/broker.py
```diff
@@ -67,8 +68,6 @@ def __init__(self, hass: HomeAssistant, hass_config: ConfigType) -> None:
         self._rems = []
         self._zons = []
 
-        self.loop_task = None
```
trvrnrth, line 70: This was unused.

custom_components/ramses_cc/broker.py
```diff
@@ -67,8 +68,6 @@ def __init__(self, hass: HomeAssistant, hass_config: ConfigType) -> None:
...
-        self._last_update = dt.min
```
trvrnrth, line 71: This was set on every update but not otherwise used so removing accordingly.

custom_components/ramses_cc/broker.py
trvrnrth, line 1: The alternative would be to rename to `RamsesCoordinator` but I think I actually prefer `RamsesBroker` as it's shorter and easier to type! So on that basis I just renamed the file to match.

zxdavb:
> listening to event-based updates from **ramses_rf** which the DataUpdateCoordinator framework is not suited to

FYI: this is a long way away - lots of refactoring required.

custom_components/ramses_cc/broker.py
zxdavb: Hmmm... I'll have to look at this - can't do it until after work.

If so, we should cherry-pick for **master** branch.

trvrnrth:
> > listening to event-based updates from **ramses_rf** which the DataUpdateCoordinator framework is not suited to
> 
> FYI: this is a long way away - lots of refactoring required.

Yup. I wasn't imagining it would be happening any time soon. Just didn't want to lock the component into a paradigm that didn't really suit long-term.

custom_components/ramses_cc/broker.py
zxdavb: a remnant from test/dev - I usually mark these with `# TODO`

zxdavb:
>  Just didn't want to lock the component into a paradigm that didn't really suit long-term

I just had an idea that a co-ordinator was a pre-requisite for config flow?

custom_components/ramses_cc/broker.py
zxdavb: can you split it out as a separate PR? SO I can cherry-pick?

zxdavb (changes requested):

trvrnrth:
> > Just didn't want to lock the component into a paradigm that didn't really suit long-term
> 
> I just had an idea that a co-ordinator was a pre-requisite for config flow?

Not that I'm aware of.

custom_components/ramses_cc/broker.py
trvrnrth: https://github.com/zxdavb/ramses_cc/pull/117

zxdavb (approved):

PR #118, opened 2023-12-21
Simplify sensor and other entity creation

trvrnrth:
I suspect there may be issues lurking in here as I've not had a chance to do much testing yet, but wanted to open this up for review to check the approach (mainly with respect to the initial discovery in the broker).


custom_components/ramses_cc/binary_sensor.py
```diff
@@ -1,34 +1,75 @@
...
+)
+
+
+@dataclass(kw_only=True)
+class RamsesBinarySensorEntityDescription(
+    RamsesEntityDescription, BinarySensorEntityDescription
+):
+    """Class describing Ramses binary sensor entities."""
+
+    attr: str | None = None
```
trvrnrth, line 63: This isn't currently used but I added it to allow the description key and device attribute to differ.

This would then allow for a HA entity id of `..._temperature` (taken from key) and a ramses_rf device attribute of `..._temp` (taken from attr). I see this has happened in the past but as far as I could tell is not currently the case anywhere.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -1,34 +1,75 @@
...
+    entity_class: RamsesBinarySensor | None = None
+    rf_class: type | UnionType | None = RamsesRFEntity
+    entity_category: EntityCategory | None = EntityCategory.DIAGNOSTIC
```
trvrnrth, line 66: This seemed like a sensible default category for the sensors.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -39,124 +80,214 @@ async def async_setup_platform(
...
+    """Representation of a fault log."""
 
     @property
     def available(self) -> bool:
         """Return True if the device has been seen recently."""
         if msg := self._device._msgs.get("0418"):
             return dt.now() - msg.dtm < timedelta(seconds=1200)
 
-    @property
-    def extra_state_attributes(self) -> dict[str, Any]:
```
trvrnrth, line 143: Now handled via the description / base entity.

custom_components/ramses_cc/sensor.py
```diff
@@ -127,309 +147,305 @@ async def async_setup_platform(
...
-
-
-class RamsesModLevel(RamsesSensor):
-    """Representation of a heat demand sensor."""
+    def available(self) -> bool:
+        """Return True if the entity is available."""
+        return self.state is not None
 
     @property
-    def extra_state_attributes(self) -> dict[str, Any]:
```
trvrnrth, line 229: I believe I'm correct in thinking that these are now all exposed as individual sensors with a global preference for the manner in which they are sourced.

custom_components/ramses_cc/binary_sensor.py
zxdavb: Yes - was a pain, so I renamed them all. 

Regardless, there's a good chance we'll see something similar in the future.

custom_components/ramses_cc/binary_sensor.py
zxdavb: I am not sure of `EntityCategory` - what does it do?

trvrnrth: It controls where a sensor shows up under a device. With the diagnostic being `for an entity exposing some configuration parameter or diagnostics of a device but does not allow changing it, for example, a sensor showing RSSI or MAC address.`. I think it's a bit of a gray area really but in my head at least anything which doesn't reflect the primary purpose/state of the device would be diagnostic. I rather imagine the list as it currently is here isn't quite correct yet.

trvrnrth: Here are the [release notes](https://www.home-assistant.io/blog/2021/11/03/release-202111/#entity-categorization) that will do a better job of explaining than me.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -165,36 +296,26 @@ def available(self) -> bool:
                 seconds=msg.payload["remaining_seconds"] * 3
             )
 
-    @property
-    def extra_state_attributes(self) -> dict[str, Any]:
```
trvrnrth, line 169: Now handled via the description / base entity.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -39,124 +80,214 @@ async def async_setup_platform(
...
+            device_class=BinarySensorDeviceClass.PROBLEM,
+        ),
+        RamsesBinarySensorEntityDescription(
+            key="system",
+            name="System",
+            rf_class=System,
+            entity_class=RamsesSystemBinarySensor,
+            device_class=BinarySensorDeviceClass.PROBLEM,
+            extra_attributes={
+                ATTR_WORKING_SCHEMA: SZ_SCHEMA,
```
trvrnrth, line 105: Just noticed that this changed to `working_schema` recently so have updated accordingly.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -39,124 +80,214 @@ async def async_setup_platform(
...
+    sensor_types: tuple[RamsesBinarySensorEntityDescription, ...] = (
+        RamsesBinarySensorEntityDescription(
+            key="gateway",
+            name="Gateway",
+            rf_class=Gateway,
+            entity_class=RamsesGatewayBinarySensor,
+            device_class=BinarySensorDeviceClass.PROBLEM,
+        ),
+        RamsesBinarySensorEntityDescription(
+            key="system",
```
trvrnrth, line 99: This sensor currently has an entity ID of `..._schema` but as it reflects the system availability `..._system` seemed like a better choice?

zxdavb:
This is a big PR, albeit atomic.

I don't know - don't have time to find out - some of the HA intricacies (e.g. `EntityCategory`), but I can see the the changes are clearly for the better.

What I'm planning to do is treat these changes as a black box.

I'm going to do bug fixes only to the **master** branch. I'll do all my dev/test of **ramses_rf** against this branch.

custom_components/ramses_cc/binary_sensor.py
zxdavb: Yes, I'd noticed this before - was on my (informal) to do list.

I am comfortable with 'minor' breaking changes, such as changing `entity_id`s for these non-core sensors.

trvrnrth:
> This is a big PR, albeit atomic.
> 
> I don't know - don't have time to find out - some of the HA intricacies (e.g. `EntityCategory`), but I can see the the changes are clearly for the better.
> 
> What I'm planning to do is treat these changes as a black box.
> 
> I'm going to do bug fixes only to the **master** branch. I'll do all my dev/test of **ramses_rf** against this branch.

Yeah, unfortunately this was about as small as I could get this set of changes.

The way the platforms are loaded and the devices are passed in via discovery isn't entirely ideal but the approach is geared towards usage with config entries and "new device" events dispatched by the broker with listeners in the platforms. I also didn't want to touch it much in this PR so as not to make it even larger!

Your approach sounds good. Whilst I'm fairly confident in the broad strokes here I'm afraid I can almost guarantee I've missed some subtlety or intricacy somewhere.

zxdavb:
> I can almost guarantee I've missed some subtlety or intricacy somewhere.

The sad fact is, in the absence of a test suite, I have been relying upon people reporting bugs when I push out major updates (and this update will be much more major than most/all so far).

If you white box, and I black box, then we can attack such risks from both ends, which should reduce our bug rate. I'll make the effort to upgrade / downgrade often without `rm -rf .storage`.

I have added the start of tests in ,**custom_components/ramses_cc/tests**... I was unable to see how (or even if) HACS had a supported way of including tests.

I am desperate to make sensor/remote faking a first class citizen - this is the most requested feature. I will be focusing on that.

custom_components/ramses_cc/binary_sensor.py
zxdavb: Yes - schemas are a real challenge.

Lots have people have difficulty with schema, despite my best efforts - a sign it isn't implemented well.

There have been lots of recent changes with warnings in logs, improvements to the wiki, and others like this: renaming the attr to make it clear it is the active schema (and not the cached, config, minimal schema).

zxdavb: > Just noticed that this changed to working_schema recently so have updated accordingly.

There have been changes to **master** that **rafactor** should have kept up with - I did look for one, and found it (removing `oem_code` for all but OTBs), so was happy. Can I leave that with you - I won't make any more big changes without porting across myself.

zxdavb:
> I suspect there may be issues lurking in here as I've not had a chance to do much testing yet, but wanted to open this up for review to check the approach (mainly with respect to the initial discovery in the broker).

Mark ready for review, and I'll have a stab at it.

trvrnrth:
> > I suspect there may be issues lurking in here as I've not had a chance to do much testing yet, but wanted to open this up for review to check the approach (mainly with respect to the initial discovery in the broker).
> 
> Mark ready for review, and I'll have a stab at it.

I've ironed out some issues (with the system and gateway sensors mainly) and everything is now looking good on my system as far as I can see.

PR #120, opened 2023-12-23
Minor refactor of cached packets (and storage) fetch

trvrnrth:
I was looking at this again whilst testing and it seemed like this was perhaps easier to follow.

Also simplifies storage data load as the wrapper function doesn't really do much.

### DIFF
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -92,8 +92,7 @@ class RamsesBroker:
             client_state.get(SZ_SCHEMA, {}) if restore_schema else {},
         )
 
-        @callback
-        def cached_packets() -> dict:
+        def cached_packets() -> dict[str, str]:  # dtm_str, packet_as_str
             if not restore_state:
                 return {}
 
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -77,7 +77,7 @@ class RamsesBroker:
 
         CONFIG_KEYS = (SZ_CONFIG, SZ_PACKET_LOG, SZ_PORT_CONFIG)
 
-        storage = await self._async_load_storage()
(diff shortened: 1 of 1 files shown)


custom_components/ramses_cc/broker.py
```diff
@@ -92,15 +92,23 @@ async def start(self) -> None:
...
         )
 
-        cached_packets = (
-            self._filter_cached_packets(
-                client_state.get(SZ_PACKETS, {}), restore_schema
-            )
-            if restore_state
-            else {}
-        )
+        @callback
```
zxdavb, line 95: Do we need this decorator?

I never 100% understood what it's for, but as its only ever going to be synchronously invoked by our `start()` (and nor, say, `ceate_task()` / `call_soon()`), I don't believe it's needed.

custom_components/ramses_cc/broker.py
```diff
@@ -92,15 +92,23 @@ async def start(self) -> None:
...
+        def cached_packets() -> dict:
```
zxdavb, line 96: Let's try not to use generic types if possible. At least `dict[str, Any]`, but we can have:
```python
def cached_packets() -> dict[str, str]:  # dtm_str, packet_as_str
```

Eventually, I'd be making everything strict-typed, and I've discovered it's a real pain to retro-fit... But I pick up so many bugs doing it.

zxdavb (changes requested):

custom_components/ramses_cc/broker.py
trvrnrth: Yeah, good shout.

custom_components/ramses_cc/broker.py
trvrnrth: No, fair point. Hass will never be calling this directly so there should be no need.

PR #121, opened 2023-12-23
Add config flow

trvrnrth:
Here's an example controller device page (ignore the unavailable entities, that's just a side-effect of my test data):
Screenshot 2023-12-23 at 15 18 16


zxdavb:
Thanks. This is going to be amazing! 

I have never had any time to get my head around config flow, so I'm not sure I can provide any value, other than testing, which will be when I get back home in a few days.


Adding tests would also be a good thing, but I am unsure how HACS supports such.

trvrnrth:
> testing, which will be when I get back home in a few days

There's no real rush on that at the moment as whilst the config flow here works, it's a long way off what it should be! I was planning on doing proper UI for the "standard" features (port selection, clearing cached schema/packets, performing discovery/eavesdropping) with the ability to specify the schema/known lists as yaml. And probably also free-form yaml to allow access to advanced ramses_rf features as well. I need to have a proper think about it.

> Adding tests would also be a good thing, but I am unsure how HACS supports such.

I wouldn't imagine hacs would have any role in it to be honest. From what I've seen other components run their pytest suites standalone/with github actions. It's not an area I've investigated properly but it should be possible to crib from the core components and other third party components to set up a suitable framework. I agree that it would be good to have something in place. It will likely make sense to mock out ramses_rf unless it already has any suitable helpers.

zxdavb:
> I was planning on doing proper UI for the "standard" features (port selection, clearing cached schema/packets, performing discovery/eavesdropping) with the ability to specify the schema/known lists as yaml. And probably also free-form yaml to allow access to advanced ramses_rf features as well. I need to have a proper think about it.

I have had a few thoughts about this...

zxdavb:
Port Selection: 

**ramses_rf** _needs_ to know if its utilizing an evofw3, or a HGI80 (other non-evofw3 gateways may exist, but have never been seen). If it has to, it assumes **evofw3**, which is a safe-enough bet.
 - https://github.com/zxdavb/ramses_rf/blob/1b8fa5a10380e2799d52b8cfc03be14869bc8457/src/ramses_tx/transport.py#L116

Before that, it can sort to using the text in a `by-id` port name, so supplying:
 - `/dev/serial/by-id/usb-SparkFun_evofw3_atmega32u4-if00` instead of
 - `/dev/ttyUSB2`

... is preferable. It also means users can swap the device into another port without issue.

a) The challenge is that the underlying library used, **pyserial*, does not enumerate the `by-id` ports, and so neither does **ramses_rf**. 

b) The opportunity is to auto-detect the port and so offer it as the only 'suggested' port, to the user during config flow.

Nonetheless, it has always been my intention to add a) and maybe b) to **ramses_rf**, so I'll do that now & let you know.  It will be the same API for both.

See: https://github.com/pyserial/pyserial/pull/709/files

zxdavb:
**Clearing cached schema/packets:**

The challenge here is to have a schema where the user can disable restoring one/both of these for the _next restart of HA only_.

Presently, users a) turn off cache, then b) restart HA, then c) *have to re-enable the cache*. It would be great if c) happened automatically.

That reminds me - I have meant to confirm that the old cache is wiped _immediately_ when a user does a), then b), rather than wait for the next cycle of saving the latest cache.

If the user does a) & b) then c) and restarts HA before the next save of the store, they restore the old cache when not expecting to.

zxdavb:
**Performing discovery/eavesdropping:**

Eavesdropping should be an advanced feature - say, only via YAML.

The other **ramses_cc** advanced feature can be UI-enabled (send packets, message evenints).

zxdavb:
> It will likely make sense to mock out ramses_rf unless it already has any suitable helpers.

**ramses_rf** ha s*very good* mocking - basically a virtual RF network with virtual serial ports / RAMSES_II devices on it - **ramses_rf** can't tell the difference.

zxdavb:
I have now modified the `comports` to include /dev/serial/by-id` symlinks:
```python
from ramses_tx.transport import comports
```

zxdavb:
FYI, this is a similar UI to what we might end up with?

image

trvrnrth:
> FYI, this is a similar UI to what we might end up with?

Something along those lines should be possible. Co-incidentally I took delivery of a Zigbee radio yesterday so I'll be paying attention to how that config flow works when I get round to setting it up!

I've been mulling over how to put together a sane menu system for the various options but haven't had a chance to actually prototype anything yet.

zxdavb:
> Zigbee radio yesterday 

I just got a HA SkyScanner (the serial dongle, above) - for matter - can't get it to appear in the **Settings > System > Devices** UI!

:(

trvrnrth:
> I have now modified the `comports` to include /dev/serial/by-id` symlinks:
> 
> ```python
> from ramses_tx.transport import comports
>    ...
> ports = comports(include_links=True)
> ```

I haven't actually used this at present and instead normalise to a list by-id with `get_serial_by_id` following the approach taken in a few of the core components.

zxdavb:
> and instead normalise to a list by-id with get_serial_by_id following the approach taken in a few of the core components.

Well, I didn't know that!

custom_components/ramses_cc/broker.py
```diff
@@ -170,18 +189,96 @@ async def async_save_client_state(self, *args, **kwargs) -> None:
...
+                task.cancel()
+                tasks.append(task)
+
+        return all(await asyncio.gather(*tasks))
+
+    def _update_device(self, device: RamsesRFEntity) -> None:
+        if hasattr(device, "name") and device.name:
+            name = device.name
+        elif isinstance(device, Evohome):
+            name = "Controller"
```
trvrnrth, line 233: This is the simplest solution to maintain the current default name/ID of the main climate entity as `Controller` and `climate.controller`.

It does mean that anyone with multiple evohome systems will have two devices called `Controller` with entity IDs `climate.controller` and `climate.controller_2` but that's not really any different to the current arrangement. The device ID is available as the serial ID to disambiguate and the presence of unique IDs allows for custom renaming of the device name and entity name/ID as desired. I also imagine that this pretty rare!

Without this special case the device would be named after the slug and ID (so `evohome 01:123456`) which isn't really as friendly and would also require overrides at the entity level to maintain a name/ID of `Controller` and `climate.controller` so I chose not to go that route.

custom_components/ramses_cc/translations/en.json
```diff
@@ -0,0 +1,183 @@
...
+            "schema": {
+                "title": "System schema and known devices",
+                "description": "Refer to the configuration section in the [wiki](https://github.com/zxdavb/ramses_cc/wiki/) for further details and examples of system schema configuration.",
+                "data": {
+                    "schema": "System schema(s)",
+                    "known_list": "Known device IDs",
+                    "enforce_known_list": "Accept packets from known devices IDs only"
+                },
+                "data_description": {
+                    "schema": "A mapping of system device IDs to their respective schemas. This should be kept minimal and only contain devices that are not automatically discovered. For example:\n```\n01:111111: # Temperature control system (e.g. evohome)\n  system:\n    appliance_control: 10:123446\n  zones:\n    \"07\": '{sensor: 01:111111}'\n\norphans_hvac: [30:111111, 32:333333, 32:555555, 32:666666]\n```",
```
trvrnrth, line 60: I'm unsure about adding in the examples here. They are convenient but do make the dialog even taller and require more scrolling. Perhaps the link to the wiki in the main dialog description is sufficient.

custom_components/ramses_cc/translations/en.json
```diff
@@ -0,0 +1,183 @@
...
+                    "rotate_bytes": "Maximum size of each packet log",
+                    "rotate_backups": "Number of packet logs to keep"
+                },
+                "data_description": {
+                    "file_name": "Path to a packet log file. Retained backups will use this filename with a suffix."
+                }
+            }
+        }
+    },
+    "options": {
```
trvrnrth, line 89: Annoyingly there is no way to inherit the translations from the config section so this duplicates the keys for the shared dialogs.

custom_components/ramses_cc/water_heater.py
```diff
@@ -149,11 +144,6 @@ def is_away_mode_on(self) -> bool | None:
         except TypeError:
             return None  # unable to determine
 
-    @property
-    def name(self) -> str | None:
-        """Return the name of the zone."""
-        return self._device.name
```
trvrnrth, line 155: This now inherits from the HA device.

custom_components/ramses_cc/climate.py
```diff
@@ -222,11 +214,6 @@ def hvac_mode(self) -> str | None:
             return HVACMode.AUTO  # users can't adjust setpoints in away mode
         return HVACMode.HEAT
 
-    @property
-    def name(self) -> str:
-        """Return the name of the Controller."""
-        return "Controller"
```
trvrnrth, line 228: This now inherits from the HA device.

custom_components/ramses_cc/climate.py
```diff
@@ -371,11 +358,6 @@ def min_temp(self) -> float | None:
         except TypeError:  # 'NoneType' object is not subscriptable
             return None
 
-    @property
-    def name(self) -> str | None:
-        """Return the name of the zone."""
-        return self._device.name
```
trvrnrth, line 377: This now inherits from the HA device.

trvrnrth:
Opening for review as this should now be fully functional.

Will also require a branding PR to [async_setup_platform](https://github.com/home-assistant/brands) in order to add an icon for the integration. I don't know if you have anything suitable?

custom_components/ramses_cc/config_flow.py
```diff
@@ -0,0 +1,591 @@
...
+)
+from .schemas import SCH_GLOBAL_TRAITS_DICT
+
+_LOGGER = logging.getLogger(__name__)
+
+CONF_MANUAL_PATH = "Enter Manually"
+
+
+def get_usb_ports() -> dict[str, str]:
+    """Return a dict of USB ports and their friendly names."""
```
trvrnrth, line 58: As an enhancement it would be good to sort these in order of likelihood of being the correct device. The UI will later default to the first item in the list on initial entry creation.

custom_components/ramses_cc/broker.py
zxdavb: > I also imagine that this pretty rare!

Not from the number of people who PM me to tell me it isn't!

> Without this special case the device would be ... so I chose not to go that route.

There are will breaking changes anyway... So just bite the bullet?

Actually, the best solution for them/us, I think, would be the same as here: [evohome wiki](https://github.com/zxdavb/evohome-async/wiki/Hass:-Multiple-Evohome-Locations): multiple integrations (i.e. add in **ramses_c1**, **ramses_c2**, etc.).

But they'll need multiple dongles in the short term, although I can see multiple ways of getting two **ramses_cc**s talking to one dongle.

custom_components/ramses_cc/translations/en.json
zxdavb: Point to the wiki. Keep the translations thin.

Then the two are decoupled, and maintenance will be 100% easier.

zxdavb:
> Opening for review as this should now be fully functional.
> 
> Will also require a branding PR to https://github.com/home-assistant/brands in order to add an icon for the integration. I don't know if you have anything suitable?

Is that still the case for when this integration is via HACS? I just don't know how that works - I never had to do it for the 3.5 integrations I wrote (it was before branding was a thing).

Or were you planning to get this integration into HA ?

I would have thought best place for it would have been official HACS integration.

Tell me your thoughts on that - I warn you that - IMO - it might be very difficult to get into HA.

custom_components/ramses_cc/climate.py
zxdavb: So, primary entities:
 - every physical device (except FANs/switches?)
 - the TCS (which is also a physical device)
 - each zone
 - HW
 - FANs, and switches
 
 Is that right?

zxdavb:
I submitted an issue on my repo - but applies to this branch.

Also, observations:
 - heat demand of a zone should be a diagnostic sensor?
 - the TCS is called "Controller", but the DHW 'zone' doesn't have a name - it's heat demand should also be a diagnostic sensor?

zxdavb:
It looks fantastic !!

custom_components/ramses_cc/broker.py
trvrnrth: If we're happy with some change in behavior then the decision boils down to what the preferred name for the system device (and climate entity) is. Do we want it to be "Controller" or "evohome 01:123456"? I can see pros/cons to each.

trvrnrth:
> Is that still the case for when this integration is via HACS? I just don't know how that works - I never had to do it for the 3.5 integrations I wrote (it was before branding was a thing).

The custom integrations (regardless of how they are sourced) live in https://github.com/home-assistant/brands/tree/master/custom_integrations

> Or were you planning to get this integration into HA ?
> 
> I would have thought best place for it would have been official HACS integration.
> 
> Tell me your thoughts on that - I warn you that - IMO - it might be very difficult to get into HA.

I doubt it would be worth the effort to get it into core so wasn't something I was anticipating.

custom_components/ramses_cc/climate.py
trvrnrth: My wording could perhaps have been a bit clearer here. What I was meaning to highlight is that this change causes these entities to just use the name of the device. Taking zones as an example there is a HA device for each zone called say "Living room" and then the climate entity for that zone is also just called "Living room" rather than "Living room Climate".

trvrnrth: For this system climate entity the name will be whatever we decide to name the device [here](https://github.com/zxdavb/ramses_cc/pull/121#discussion_r1451712455).

custom_components/ramses_cc/config_flow.py
zxdavb: > As an enhancement it would be good to sort these in order of likelihood of being the correct device. The UI will later default to the first item in the list on initial entry creation.

I can't think of any way, except via its USB descriptor fields.

Or you could ping it, and see if it behaves appropriately - how could you do that, if another library is using it.

trvrnrth:
> * heat demand of a zone should be a diagnostic sensor?
> 
> * the TCS is called "Controller", but the DHW 'zone' doesn't have a name - it's heat demand should also be a diagnostic sensor?
> 
> 
> They such great sensors, though - maybe best to ensure they're exposed ? ... dunno.

I believe you left a list of thoughts on those elsewhere. I have yet to take a proper look at that but I think it will make sense to do in a separate PR. The changes here make the entity category setup more evident but of don't actually change them.

zxdavb:
> In order to add an icon for the integration. I don't know if you have anything suitable?

I don't know of anything suitable. It would have to be a generic icon.

RAMSES II applies to CH/DHW *and* HVAC, and is wireless. So pick anything you like, with that in mind.

custom_components/ramses_cc/config_flow.py
trvrnrth: Yeah, I was just thinking of sorting based on the existence of, say, "evofw3" in the description.

zxdavb: There's a new dongle coming... might be very popular... will do:
 - ramses_rf <---> MQTT (over WiFi) <---> dongle <---> RF

It's by-id is: `/dev/serial/by-id/usb-Espressif_USB_JTAG_serial_debug_unit_84:FC:E6:64:45:8C-if00`

custom_components/ramses_cc/broker.py
zxdavb: I'd say "controller 01:123456".

The unique ID must be be unique - everything else is forgivable.

My fear is, that I didn't set it up right - now is the time to sort that out.

Of note: this doesn't look right to me:

```python
class RamsesEntity(Entity):
    def __init__(self, ...) -> None:
        self._attr_unique_id = device.id
```

A whole bunch of entities will have the same unique-id. I think HA then does some woo to pick up the pieces... ?using the entity-id?

For example, in the entity registry:
```
        "entity_id": "binary_sensor.01_145038_01_window_open",
        "unique_id": "01:145038_01-window_open",
```
... for the controller, and one of the zones:
```json
        "entity_id": "climate.controller",
        "unique_id": "01:145038",
...
        "entity_id": "climate.ramses_cc_01_145038_00",
        "unique_id": "01:145038_00",
```

trvrnrth: The sensors handle that [here](https://github.com/zxdavb/ramses_cc/blob/1ff957a7c7565e7b52ee2d8fa722eca3744243b6/custom_components/ramses_cc/sensor.py#L157) and [here](https://github.com/zxdavb/ramses_cc/blob/1ff957a7c7565e7b52ee2d8fa722eca3744243b6/custom_components/ramses_cc/binary_sensor.py#L119). The RamsesEntity default is used for climate, water heater and remote entities which works out as only one entity from any one of those platform domains will be possible per device at present.

"Controller 01:123456" seems reasonable. That's probably clearer than using the slug.

zxdavb: > That's probably clearer than using the slug.

The slugs are a little messy.

They are used for device _class_ (e.g. BDR, OTB, TRV (and sometimes device _type_: TRV vs TR2)) and also the CH/DHW _roles_ (TCS, ELE, RAD).

I would like to switch away from using slugs for roles... For example, is DHW a CS92A or the HW part of the CH/DHW system.
 ... and people always confuse the TCS (CH/DHW controller) with the CTL (CH/DHW controller!).

zxdavb:
BTW, do you have any appetite for adding config flow to the official **evohome** integration?

```python
CONFIG_SCHEMA = vol.Schema(
    {
        DOMAIN: vol.Schema(
            {
                vol.Required(CONF_USERNAME): cv.string,
                vol.Required(CONF_PASSWORD): cv.string,
                vol.Optional(CONF_LOCATION_IDX, default=0): cv.positive_int,
                vol.Optional(
                    CONF_SCAN_INTERVAL, default=SCAN_INTERVAL_DEFAULT
                ): vol.All(cv.time_period, vol.Range(min=SCAN_INTERVAL_MINIMUM)),
            }
        )
    },
    extra=vol.ALLOW_EXTRA,
)
```

It has been on my to-do list for a very long time.

custom_components/ramses_cc/broker.py
trvrnrth: I guess it would be nice to provide a friendlier name for all device types really. I used the slug for a prefix as it seemed like the best bet at the moment but it seemed worth considering the controller immediately as it's a bit more key.

custom_components/ramses_cc/config_flow.py
trvrnrth: That does sound interesting. I think I'll leave this alone for the time being. It seems like the evofw3 is likely to be the only device that's easy to detect reliably and the extra complexity the sorting adds is perhaps not really worth it just to get it to show up first in the list.

trvrnrth:
> BTW, do you have any appetite for adding config flow to the official **evohome** integration?

I'd probably describe it as limited but necessarily non-existent.

custom_components/ramses_cc/broker.py
trvrnrth: Adjusted to be "Controller 01:123456" for `Evohome` instances. Would it make sense to extend this to all `System` instances?

zxdavb:
I haven't had a chance to play with this as much as I'd have liked - will concentrate on getting faking working on **0.31.x** for a short while.

Would you be OK to switch to python 3.12.x - there are some typing features with StrEnum I'd like to include.

custom_components/ramses_cc/broker.py
zxdavb: > Adjusted to be "Controller 01:123456" for Evohome instances. Would it make sense to extend this to all System instances?

Yes, although I am not sure how many non-evohome systems there are.

zxdavb: I want to change the entity_id of DHW to `water_heater.01_123456_hw` to match that of the climate entites, say: `climate.01_123456_01` - what is the best way to do that for config flow?

The issue with DHW is that - unlike zones - it has a name when instantiated.

So, I want to set the name of the evohome TCS/Zone/DHW regardless of name.

trvrnrth: Updated accordingly.

trvrnrth: > I want to change the entity_id of DHW to `water_heater.01_123456_hw` to match that of the climate entites, say: `climate.01_123456_01` - what is the best way to do that for config flow?
> 
> The issue with DHW is that - unlike zones - it has a name when instantiated.
> 
> So, I want to set the name of the evohome TCS/Zone/DHW regardless of name.

The best thing would be to explicitly provide a preferred `entity_id` in the same way as is [done for the sensors](https://github.com/zxdavb/ramses_cc/blob/94ff40e2d061950e22855ed14c0ec7f75bab82fc/custom_components/ramses_cc/sensor.py#L160). At present as you say the entity ID is generated from the entity name (inherited from the device in this case).

I assume the idea is that the entity ID will remain fixed at a known stable value regardless of zone renames etc.

zxdavb: If the unique_id doesn't change, then the entity_id won't change, even if it (i.e. the 'offer' from the integration) is changed in the integration - you have to remove the entity from the registry.

zxdavb: OK, Climate, Remote & WaterHeater all have:
```python
self.entity_id = f"{PLATFORM}.{device.id}"
```

... and the sensors remain:
```python
self._attr_unique_id = f"{device.id}-{entity_description.key}"
```

trvrnrth: Yep. I wasn't being very clear there. What I meant was that it would perhaps be wise to provide an explicit entity_id 'offer' so that it is not closely coupled to the device/entity name provided. ie. you could change the generated friendly name of the water heater or controller devices in a future version but still retain the same entity id and thus not affect expectations for automations/documentation etc. for newly created entries.

trvrnrth: I did also kind of wonder if an entity_id of `climate.01_123456` might make more sense for the controller to fit with the general approach (it's currently implicitly generated from the provided name as `climate.controller_01_123456`). Or, alternatively `climate.01_123456_controller`.

trvrnrth: > OK, Climate, Remote & WaterHeater all have:
> 
> ```python
> self.entity_id = f"{PLATFORM}.{device.id}"
> ```
> 
> ... and the sensors remain:
> 
> ```python
> self.entity_id = ENTITY_ID_FORMAT.format(f"{device.id}_{entity_description.key}")
> ```

That seems fairly sane to me. Although, for climate for example you can do:
```
from homeassistant.components.climate import ENTITY_ID_FORMAT
...
self.entity_id = ENTITY_ID_FORMAT.format(device.id)
```
Which gets you the same thing but seemed neater to me.

trvrnrth: I was making the change at the same time and also corrected the domain for remote and sensor entities which I'd botched. It's interesting that that gets corrected by HA anyway but much better if it's right to start with!

I've merged your changes in to resolve the conflicts.

trvrnrth:
> Would you be OK to switch to python 3.12.x - there are some typing features with StrEnum I'd like to include (or is that mypy 1.8.x?)

Oddly the core dev container hasn't been updated yet which I was waiting for to save myself some hassle. In the meantime though I've not noticed any typing/linting issues. Anyway, I'm sure I can sort out an upgrade when required so go ahead!

zxdavb:
@trvrnrth 

I am happy to pull this into my **config_flow** branch, however there are conflicts.

I was attempting to resolve the conflicts myself, but think I should leave it to you.

After that, I'll see if it starts without issue, and we can post it to HACS as a pre-release.

trvrnrth:
> @trvrnrth
> 
> I am happy to pull this into my **config_flow** branch, however there are conflicts.
> 
> I was attempting to resolve the conflicts myself, but think I should leave it to you.
> 
> After that, I'll see if it starts without issue, and we can post it to HACS as a pre-release.

I don't have time for any extensive testing this evening but I've merged in and confirmed that at a glance things look OK on my system.

zxdavb:
Merged

zxdavb:
@trvrnrth I will shortly post this as a pre-release, **0.40.0**.

I would like to acknowledge your work - would you be OK if I ping you in the release notes, and/or on the HA forum?

trvrnrth:
> @trvrnrth I will shortly post this as a pre-release, **0.40.0**.

Awesome.

> I would like to acknowledge your work - would you be OK if I ping you in the release notes, and/or on the HA forum?

Yes, a mention is fine by me.

PR #131, opened 2024-01-10
Fix DHW zone discovery

trvrnrth:
This should resolve issues with DhwZone discovery. Unfortunately I do not have a system that I can test this against.

### DIFF
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -17,7 +17,7 @@ from ramses_rf.schemas import (
     SZ_RESTORE_STATE,
     SZ_SCHEMA,
 )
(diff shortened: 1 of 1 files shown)


custom_components/ramses_cc/broker.py
```diff
@@ -186,7 +186,12 @@ def new_entities(
         new_zones = new_entities(
             self._zones,
             set().union(
-                *[s.zones for s in self.client.systems if isinstance(s, MultiZone)]
+                *[s.zones for s in self.client.systems if isinstance(s, MultiZone)],
+                [
```
zxdavb, line 190: There should be an `s.child` - but it wouldn't be too reliable - so let's stick with this - thanks!

PR #132, opened 2024-01-10
Always cast binary sensor is_on to bool

trvrnrth:
This is only really necessary for the active_fault attr at present but there is no harm having it in place for all current and future sensors.

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -125,7 +125,8 @@ class RamsesBinarySensor(RamsesEntity, BinarySensorEntity):
     @property
     def is_on(self) -> bool | None:
         """Return the state of the binary sensor."""
-        return getattr(self._device, self.entity_description.attr)
+        result = getattr(self._device, self.entity_description.attr)
+        return None if result is None else bool(result)
 
     @property
     def icon(self) -> str:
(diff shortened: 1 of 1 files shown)


custom_components/ramses_cc/binary_sensor.py
```diff
@@ -125,7 +125,8 @@ def available(self) -> bool:
     @property
     def is_on(self) -> bool | None:
         """Return the state of the binary sensor."""
-        return getattr(self._device, self.entity_description.attr)
+        result = getattr(self._device, self.entity_description.attr)
+        return None if result is None else bool(result)
```
zxdavb, line 129: I don't think we need to do this - I checked, and all of these attrs appear to be `bool`s.

custom_components/ramses_cc/binary_sensor.py
```diff
@@ -159,12 +160,6 @@ def available(self) -> bool:
         msg = self._device._msgs.get("0418")
         return msg and dt.now() - msg.dtm < timedelta(seconds=1200)
 
-    @property
-    def is_on(self) -> bool | None:
-        """Return the state of the binary sensor."""
-        result = getattr(self._device, self.entity_description.attr)
-        return None if result is None else bool(result)
-
```
zxdavb, line 167: I'll leave this in for now.

zxdavb:
Can re-visit this in future.

custom_components/ramses_cc/binary_sensor.py
trvrnrth: This was just so that the override in `RamsesLogbookBinarySensor` could be removed but now that I think about it I don't think that override is quite what we want.

trvrnrth:
Superseded by https://github.com/zxdavb/ramses_cc/pull/133

PR #135, opened 2024-01-13
Move all service schemas to schemas.py

trvrnrth:
I'm inferring from recent changes that this is the preferred location/naming convention. I was just noticing that it was a bit odd having them split across multiple locations when merging into `config-flow`.

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -13,7 +13,6 @@ from ramses_rf.entity_base import Entity as RamsesRFEntity
 from ramses_rf.system.heat import Evohome
 from ramses_rf.system.zones import Zone
 from ramses_tx.const import SZ_MODE, SZ_SETPOINT, SZ_SYSTEM_MODE
-import voluptuous as vol  # type: ignore[import-untyped]
 
 from homeassistant.components.climate import (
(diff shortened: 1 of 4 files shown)


zxdavb:
I know you moved them out to the respective platforms - was happy with that, but I then wanted a schema used in **sensors** to be available to  **climate** / **water_heater**, so moved them back.

There was another change you made that I reversed - but the data struct was sometimes a set, and sometimes a list (but always iterable), and the **mypy** work just meant I had to pick a side, was easier to drop your sets - no criticism intended.

zxdavb:
If you're keen - can install **voluptuous** 0.14.1, run mypy, then roll back to 0.13.1 (HA shoudl do this when you start it).

The newer version of **voluptuous** support typing, so that may be helpful.

trvrnrth:
> There was another change you made that I reversed - but the data struct was sometimes a set, and sometimes a list (but always iterable), and the **mypy** work just meant I had to pick a side, was easier to drop your sets - no criticism intended.

I imagine you're probably talking about the device/entity discovery. To be honest your refactor with lists is probably easier to understand anyway and it's unlikely that any performance differences matter that much with the number of devices that will be in any given system.

PR #136, opened 2024-01-13
Fix typo in HVAC humidity sensor names

trvrnrth:
As a side note these should really all be in translations.


zxdavb:
> As a side note these should really all be in translations.

Yep.

PR #141, opened 2024-01-18
Consistent approach to entity ID generation across all platforms

trvrnrth:
Uses ENTITY_ID_FORMAT for all platforms.

Also fixes incorrect prefix for remote and sensor. Interestingly this got corrected by HA but it's better if it's correct to start with!


zxdavb:
I think I hit a cognitive block...

They were supposed to be: `f"{DOMAIN}.{device.id}"`

Nice spot.

> Interestingly this got corrected by HA

Because about 3y ago everyone got confused.

trvrnrth:
You had the recent ones correct (using the DOMAIN from the platform as PLATFORM) but those aliases quickly get confusing which is nicely side-stepped by using the format constant instead.

I can imagine plenty of past confusion. Using `DOMAIN` to mean various subtly different things seems like a recipe for it!

PR #142, opened 2024-01-18
Fix type hints for controller temp min/max limits

trvrnrth:
This seems like the best thing to do to maintain the current behaviour.


trvrnrth:
Yeah. It feels like a bit of a danger area of undefined behaviour really.

PR #150, opened 2024-02-03
Neaten entity descriptions

trvrnrth:
Explicitly set `ramses_rf_attr` so we no longer need to default these.

Follow on from https://github.com/zxdavb/ramses_cc/pull/148#issuecomment-1921031732

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -209,21 +209,9 @@ class RamsesBinarySensorEntityDescription(
     icon_off: str | None = None
 
     # integration-specific attributes
-    ramses_cc_class: type[RamsesBinarySensor] = None  # type: ignore[assignment]
-    ramses_rf_attr: str = None  # type: ignore[assignment]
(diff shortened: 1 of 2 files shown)


trvrnrth:
They keys should probably be either hard-coded into the description table (or from ramses CC constants) to de-couple them from ramses_rf. And the names/icons should come from translations. But things for another day!

zxdavb:
I want to make the **refactor** branch stable. I think this is best left out.

Once I've pushed that (tomorrow?), then we change make the **config_flow** branch our new **master** (I only have the brain capacity for two branches!).

zxdavb:
For some reason, my debugger has broken - wont stop at breakpoints - dunno why - has been like that for over a week now - I'm tearing my hair out!

trvrnrth:
> I want to make the **refactor** branch stable. I think this is best left out.
> 
> Once I've pushed that (tomorrow?), then we change make the **config_flow** branch our new **master** (I only have the brain capacity for two branches!).

Sounds good. This PR is not at all important so can easily wait a while.

zxdavb (approved):

PR #160, opened 2024-02-08
Add description to cache clear config options step

trvrnrth:
This helps explain the new immediately clear behaviour as well as why you might want to be doing it in the first place.

Screenshot 2024-02-08 at 08 17 40

### DIFF
--- a/custom_components/ramses_cc/translations/en.json
+++ b/custom_components/ramses_cc/translations/en.json
@@ -173,6 +173,7 @@
             },
             "clear_cache": {
                 "title": "Clear cache",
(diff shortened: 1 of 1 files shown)


zxdavb:
It will need to be:
```python
if pkt[41:45] not in ["0004", "0005", "000C"]
```

Reason is: a person will add/remove a zone & be confused it has the old name.

You _could_ argue it should be anything with a `zone_idx`, but if we immediately discover, it shouldn't be too much of a problem.

Now that I have async `send_cmd()` calls, with QoS - I'd like to switch from polling to event-driven requests for much of this data.

zxdavb (approved):

PR #161, opened 2024-02-08
Use sets for discovered entities

trvrnrth:
This was mostly to satisfy my own curiosity off the back of previous discussion. I won't be offended if you choose to stick to using lists here.

### DIFF
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -84,10 +84,10 @@ class RamsesBroker:
         self._device_info: dict[str, DeviceInfo] = {}
 
         # Discovered client objects...
-        self._devices: list[Device] = []
-        self._systems: list[System] = []
-        self._zones: list[Zone] = []
-        self._dhws: list[Zone] = []
+        self._devices: set[Device] = set()
+        self._systems: set[System] = set()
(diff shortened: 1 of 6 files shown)


custom_components/ramses_cc/broker.py
```diff
@@ -276,37 +276,38 @@ async def async_add_entities(
...
         def find_new_entities(
-            known: list[RamsesRFEntity], current: list[RamsesRFEntity]
-        ) -> tuple[list[RamsesRFEntity], list[RamsesRFEntity]]:
-            new = [x for x in current if x not in known]
-            return known + new, new
+            known: set[RamsesRFEntity], current: set[RamsesRFEntity]
+        ) -> set[RamsesRFEntity]:
+            new = current - known
+            known |= new
+            return new
```
zxdavb, line 283: This makes me nervous. Why did we have the tuple before - why was it needed then, and is no longer needed now?

zxdavb:
> I won't be offended if you choose to stick to using lists here.

I will stick with lists

zxdavb:
FWIW, I also tried it with tuples instead of lists, and it looked really busy 
- I though minimal gain for lots of added complexity to the codebase

PR #165, opened 2024-02-11
DHW should benefit from `update_device()` too

zxdavb:
@trvrnrth I wonder if this is a bug - can you just confirm for me?

```python
for device in self._devices + self._systems + self._zones:
    self._update_device(device)
```
... should be:
```python
for device in self._devices + self._systems + self._zones + self._dhws:
    self._update_device(device)
```

### DIFF
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -295,7 +295,7 @@ class RamsesBroker:
         )
         self._devices, new_devices = find_new_entities(self._devices, gwy.devices)
 
-        for device in self._devices + self._systems + self._zones:
+        for device in self._devices + self._systems + self._zones + self._dhws:
             self._update_device(device)
 
         new_entities = new_devices + new_systems + new_zones + new_dhws


zxdavb:
And OMG, that parent/child code is a real mess!

I was seriously thinking of removing it altogether. 

Part of the issue is that device (e.g. a BDR91A) can have multiple parents - all of equal priority.

Another is misconfigured systems - a person who moves a TRV from one zone to another without destroying/recreating the original zone will find their TRV a child of two zones...

So, what can we get away with here if I deprecate `._parent`:
```python
if isinstance(device, Zone) and device.tcs:
    via_device = (DOMAIN, device.tcs.id)
elif isinstance(device, Child) and device._parent:  # I think remove?
    via_device = (DOMAIN, device._parent.id)
else:
    via_device = None
```

trvrnrth (approved):

trvrnrth:
Personally I find the device topology for zone thermostats and TRVs really quite useful so it would be a shame to lose that functionality. Perhaps something more appropriate could be done to cater for that use case?

zxdavb:
Something will be kept - It just needs to be made much simple (have a look, if you want a giggle).

The other thing that has to go is `zone_idx`, `ufh_idx`, `dhw_idx` - it should just all be `idx`.
 - at the moment, the index type is purely from the command code 
 - but command codes are shared by CH/DHW and HVAC - some are even used by evohome controllers and UFH controllers (both CH/DHW)

PR #194, opened 2024-06-17
Add impersonation support to send_packet service

Kars-de-Jong:
This change adds support for impersonation using the send_packet service.

I use it to send fan_demand packets from my faked CO2 sensor, because ramses_rf doesn't support that yet.

### DIFF
--- a/custom_components/ramses_cc/manifest.json
+++ b/custom_components/ramses_cc/manifest.json
@@ -8,10 +8,10 @@
     "loggers": ["ramses_rf"],
     "requirements": [
       "pyserial-asyncio-fast>=0.14",
-      "ramses-rf==0.42.0"
+      "ramses-rf==0.42.1"
     ],
     "config_flow": true,
     "single_config_entry": true,
     "type": "hub",
-    "version": "0.42.0"
+    "version": "0.42.1"
   }


tests/test_services.py
```diff
@@ -98,6 +98,7 @@
         "32:097710": {"class": "CO2"},
         "32:139773": {"class": "HUM"},
         "40:123456": {"class": "REM", "faked": True},
+        "37:123456": {"class": "FAN"}
```
zxdavb, line 101: In order of device_id, please.
```python
        "32:139773": {"class": "HUM"},
        "37:123456": {"class": "FAN"},
        "40:123456": {"class": "REM", "faked": True},
```

custom_components/ramses_cc/translations/en.json
```diff
@@ -152,7 +152,7 @@
             "advanced_features": {
                 "title": "Advanced features",
                 "data": {
-                    "send_packet": "Enable send_packet service for broadcasting bespoke packets",
+                    "send_packet": "Enable send_packet service for sending bespoke packets",
```
zxdavb, line 155:
```
"Enable send_packet service for casting bespoke packets"
```

custom_components/ramses_cc/services.yaml
```diff
@@ -65,6 +65,14 @@ send_packet:
       example: 01:123456
       required: true
 
+    from_id:
+      description: >-
+        The source device ID (a RAMSES ID, not an entity_id).
+        This can be used to send a packet from a faked device.
+        If not specified, the device ID of the gateway is used.
```
zxdavb, line 72:
```text
Optional: if not specified, the device ID of the gateway is used.
```

zxdavb (changes requested):
Very good - some minor changes please, and we can merge.

Kars-de-Jong:
@zxdavb I processed your comments, could you have another look?

zxdavb:
@Kars-de-Jong I'm curious - why couldn't you solve this use-case with a remote?

Kars-de-Jong:
That's a good question... I just assumed the HRU needs to see both co2_level and fan_demand packets from the same device.

PR #237, opened 2025-04-03
Config flow NL

silverailscolo:
Translation of UI strings, add example in _en port field as well

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -20,5 +20,5 @@
 
 # used for development (typing)
     mypy >= 1.15.0
-    types-paho-mqtt >= 1.6.0
+    types-paho-mqtt >= 1.6.0                     # no longer required when we use paho-mqtt 2.x
     voluptuous >= 0.15.2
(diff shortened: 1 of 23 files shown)


requirements_dev.txt
```diff
@@ -20,5 +20,5 @@
 
 # used for development (typing)
     mypy >= 1.15.0
-    types-paho-mqtt >= 1.6.0
+    types-paho-mqtt >= 1.6.0                     # no longer required when we use paho-mqtt 2.x
```
zxdavb, line 23: why did you remove this comment? (a 3rd PR, unrelated to the other 2?)

requirements_test.txt
```diff
@@ -5,7 +5,7 @@
 # may need to use this version of ramses_rf...
 # -e ~/client/ramses_rf
 
-# for using the the library with HA
+# for using the library with HA
```
zxdavb, line 8: You have three PRs here:
 - one to add translations 
 - another to fix a typo (the the)
 - removing a comment

Generally, try and keep PRs distinct & atomic.

zxdavb (approved):
Happy to merge - but really should be at least two PRs - one for translations, another for housekeeping.

requirements_dev.txt
silverailscolo: I added the comment, in fact. There is a PR in ramses_rf to bump paho_mqtt to >= 2.1.0 and that should allow type hints without this stub. For now, just a marker here.

PR #238, opened 2025-04-09
Housekeeping tidbits

silverailscolo:
Split off from PR #237, as requested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,3 +186,9 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# IntelliJ IDE stuff
+/.idea/
+
+# sphinx folder
+/docs/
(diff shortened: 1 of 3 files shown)


zxdavb (approved):

zxdavb:
As an aside - always useful to base upon the latest version of:
 - https://github.com/github/gitignore/blob/main/Python.gitignore
 
... where it says:
``` 
# PyCharm
#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
#  and can be added to the global gitignore or merged into this file.  For a more nuclear
#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
#.idea/
```

... so maybe you want to append that onto the end of
 - our ignores, then
 - https://github.com/github/gitignore/blob/main/Python.gitignore
 - https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore

PR #240, opened 2025-04-09
Add note on paho typing

silverailscolo:


requirements_dev.txt
```diff
@@ -20,5 +20,5 @@
 
 # used for development (typing)
     mypy >= 1.15.0
-    types-paho-mqtt >= 1.6.0
+    types-paho-mqtt >= 1.6.0                     # no longer required when we use paho-mqtt 2.x in ramses_rf 0.50.2
```
zxdavb, line 23: if it is no longer required, why not simply delete the line altogether?

zxdavb (changes requested):
see comment

requirements_dev.txt
silverailscolo: As this is in `ramses_cc`, removal would have to coincide with next release of `ramses_rf`.

zxdavb: the convention here is that release x is the tagged release - other commits with teh same number a YMMV.

PR #244, opened 2025-04-22
Services i18n

silverailscolo:
- Moved default `en` strings from `services.yaml` to `/translations/en.json`
- Added Dutch translations for all services
- Kept some `name:` values as comments for code readability
- Moved worded `unit_of_measurement` to the description string (e.g. `minutes`), because i18n them is not supported. Leaves room on mobile screen too


zxdavb (approved):
Great work!

PR #251, opened 2025-05-11
Tweak `services`

silverailscolo:
Replace Action `device_id` field drop-down selectors by `target`s, allowing use of HA system picker by room/floor/device.
Note that Descriptions for entity_ids (now targets) were lost and should be deleted in translations.


zxdavb (changes requested):
@silverailscolo would you like to resolve the conflicts, and we can merge

zxdavb (approved):
thanks

PR #261, opened 2025-07-19
Fix config flow cache clearing for all Home Assistant versions

wimpie70:
I found some errors when trying to clear the cache. This should fix them.
Installation method Home Assistant Container
Core 2025.7.0b7
Frontend 20250701.0

File "/config/custom_components/ramses_cc/config_flow.py", line 543, in async_step_clear_cache
    store: Store = self.hass.helpers.storage.Store(STORAGE_VERSION, STORAGE_KEY)
                   ^^^^^^^^^^^^^^^^^
AttributeError: 'HomeAssistant' object has no attribute 'helpers'

  File "/config/custom_components/ramses_cc/config_flow.py", line 551, in async_step_clear_cache
    store = storage.Store(self.hass, STORAGE_VERSION, STORAGE_KEY)
            ^^^^^^^
UnboundLocalError: cannot access local variable 'storage' where it is not associated with a value

PR #262, opened 2025-07-20
Fix/config flow cache clean local

wimpie70:
removed the ha version checks


wimpie70:
sorry, created a new PR instead of a commit to #261. This can be closed.

PR #277, opened 2025-08-22
For issue 276

silverailscolo:
In service .set_temperature, normalise the entered data before passing it on to async_set_zone_mode.
Assume until is more important than duration, or throw an error?

Added in PR: make timed service tests stable. Testing on this PR might still trip on that.


silverailscolo:
Thanks for your reports. I made some tweaks to better match the schema and my table. It's a bit opaque what exactly is passed around, and I tried again but I do not get the full _set_temperature_ pane.
Here's a new `climate.py` zip (also in the branch):
[climate.py.zip](https://github.com/user-attachments/files/21944207/climate.py.zip)

Issue #283, opened 2025-08-29
Value error `sensor/__init__.py`


silverailscolo:
Split from issue #282:

@Kars-de-Jong noticed a lot of exceptions in his logs:
```
Logger: homeassistant.helpers.dispatcher
Bron: util/logging.py:167
Eerst voorgekomen: 09:17:47 (55 gebeurtenissen)
Laatst gelogd: 10:01:50

[...]

PR #284, opened 2025-08-31
Tweak `bypass_mode` sensor

silverailscolo:
Remove sensor unit % to fix issue #283

### DIFF
--- a/custom_components/ramses_cc/sensor.py
+++ b/custom_components/ramses_cc/sensor.py
@@ -34,14 +34,13 @@ from ramses_rf.const import (
     SZ_AIR_QUALITY,
     SZ_AIR_QUALITY_BASIS,
     SZ_BYPASS_MODE,
-    # debug: why marked as not found in IDE, is in RFv0.51.3 (depends on ramses_rf>=0.51.3)
     SZ_CO2_LEVEL,
     SZ_EXHAUST_FAN_SPEED,
     SZ_EXHAUST_FLOW,
     SZ_EXHAUST_TEMP,
     SZ_FAN_INFO,
     SZ_FAN_MODE,
-    SZ_FAN_RATE,  # same issue
+    SZ_FAN_RATE,
     SZ_FILTER_REMAINING,
     SZ_INDOOR_HUMIDITY,
     SZ_INDOOR_TEMP,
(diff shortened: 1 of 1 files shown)


custom_components/ramses_cc/sensor.py
```diff
@@ -409,7 +408,6 @@ class RamsesSensorEntityDescription(RamsesEntityDescription, SensorEntityDescrip
         key=SZ_BYPASS_MODE,
         ramses_rf_attr=SZ_BYPASS_MODE,
         name="Bypass mode",
-        native_unit_of_measurement=PERCENTAGE,
```
Kars-de-Jong, line 412: This is not enough, I still get an exception:
```
2025-09-01 08:46:50.556 ERROR (MainThread) [homeassistant.helpers.dispatcher] Exception in async_write_ha_state when dispatching 'ramses_cc_update': ()
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 663, in state
    numerical_value = int(value)
ValueError: invalid literal for int() with base 10: 'auto'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 666, in state
    numerical_value = float(value)
ValueError: could not convert string to float: 'auto'
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1026, in async_write_ha_state
    self._async_write_ha_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1151, in _async_write_ha_state
    self.__async_calculate_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1087, in __async_calculate_state
    available = self.available  # only call self.available once per update cycle
                ^^^^^^^^^^^^^^
  File "/config/custom_components/ramses_cc/sensor.py", line 150, in available
    ) or self.state is not None  # TODO: but what if None _is_ its state?
         ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 670, in state
    raise ValueError(
    ...<5 lines>...
    ) from err
ValueError: Sensor sensor.32_123456_bypass_mode has device class 'None', state class 'measurement' unit 'None' and suggested precision 'None' thus indicating it has a numeric value; however, it has the non-numeric value: 'auto' (<class 'str'>)
```
You need to set `state_class` to `None` as well.

silverailscolo: Added that, thanks for the hint

silverailscolo:
See PR #285

Issue #287, opened 2025-09-05
Config: serial port drop-down select causes 'Unknown error occurred'


IMMRMKW:
**Describe the bug**
The option to simply select one of the port options in Configuration - `Choose serial port` does not continue (unknown error occurred). If it is entered manually, port config continues without error.

**To Reproduce**
Steps to reproduce the behaviour (for example):
1. In my case: Deinstall Ramses RF
2. Restart Home Assistant
3. Go to `devices & services`
4. Click `+ Add integration`
5. Search for `Ramses RF` and start configuring
6. In `Choose serial port`, select your device. In my case: `USB JTAG/serial debug unit - /dev/serial/by-id/usb-Espressif_USB_JTAG_serial_debug_unit_E4:B3:23:A8:34:F0-if00, s/n: E4:B3:23:A8:34:F0 - Espressif - 303A:1001` and click `next`.
7. In `Serial port config` click `next`, but here, the `Unknown error occurred` is thrown, and further configuring is not possible.
8. If, instead, I choose `enter manually` and copy+paste the device path from step 6, I can make it work.

**Expected behaviour**
I can configure my device simply by selecting the device path in `Choose serial port`.

**Information about setup:**

 - RF gateway: my own Ramses ESP32-C6, with [this firmware](https://github.com/IMMRMKW/ramses_esp)
 - Ramses RF: 0.51.4
 - Serial connection
 - System log:
```
2025-09-05 10:26:29.045 ERROR (MainThread) [aiohttp.server] Error handling request from 192.168.1.149
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 510, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/security_filter.py", line 92, in security_filter_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #275, opened 2025-08-20
NanoCul can't install (serial port)


fysmd:
**Describe the bug**
Cannot setup ramses, after following the installation guide and restarting HA,

**To Reproduce**
1. Add Integration, search for RAMSES RF, select it
2. at Choose Serial Port, I am offered a number of options and I chose:
  NANO CUL - NANO CUL - /dev/serial/by-id/usb-SHK_NANO_CUL_868-if00-port0, s/n: 868 - SHK - 0403:6001
3. Serial Port config dialog shows, noting under Advanced:  "not required for typical use case, make no edits, click Next
4.   UNKNOWN ERROR OCCURRED displayed in Red in same Serial Port config dialog

**Expected behavior**
I would expect the detected serial adapter to work and setup workflow continue past the Serial Port COnfig dialog

**Please include the following information:**
- Ramses RF v 0.51.3
- Installation method. Home Assistant OS
- Core. 2025.8.2
- Supervisor. 2025.08.1
- Operating System. 16.1
- Frontend. 20250811.0

 HA Log contains:
```
2025-08-20 11:02:45.160 ERROR (MainThread) [aiohttp.server] Error handling request from 192.168.0.1
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 510, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/security_filter.py", line 92, in security_filter_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/forwarded.py", line 83, in forwarded_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/http/request_context.py", line 26, in request_context_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #288, opened 2025-09-05
Fix config serial port error

IMMRMKW:
Fix for #287 #275.

- Fix get_options() to preserve options set during config flow
- Add debug logging for port name handling
- Add error handling for missing port names
- Prevents KeyError when port_name is not found
- Prevents Gateway constructor TypeError when port_name is None

Fixes issue where selecting a serial port from the dropdown would result in port_name being lost between config flow steps, causing Gateway initialization to fail.


silverailscolo:
Thanks @IMMRMKW

PR #289, opened 2025-09-06
Add Feature: a ramses_cc.add_command action to add commands outside the config flow

IMMRMKW:
There are Ramses commands that allow a remote to set the ventilation unit to any percentage between 44% and 89.5%, with a defined resolution. For the Orcon ventilation unit, I identified the relevant command to be `31E0`.

While it's straightforward to calculate the required command for a specific percentage using a Python script and transmit it via the remote, the current implementation of ramses_cc does not support this functionality directly.

Although I believe the ideal solution would be to send arbitrary command strings to the remote in real time, I haven't implemented this yet—mainly because I haven't found a way to do this yet. Instead, I devised a workaround by introducing a new action in ramses_cc called `add_command`. Here's how it can be used:

```yaml
service: ramses.add_command
target:
  entity_id: remote.29_162275
data:
  command: setpoint  
  packet_string: " I --- 29:162275 32:146231 --:------ 31E0 008 000046000100AA00"
```

Once registered, the command can be triggered as follows:

```yaml
action: ramses_cc.send_command
target:
  entity_id: remote.29_162275
data:
  command: setpoint
  num_repeats: 3     # optional (default 3)
  delay_secs: 0.05   # optional (default 0.05)
```

I made the following changes in the code for this:

- Added SVC_ADD_COMMAND and SCH_ADD_COMMAND schemas in schemas.py
- Implemented async_add_command method in remote.py for direct command injection
- Removed unused imports and applied consistent code formatting

Note: All ramses_cc actions are also exposed under the remote namespace (e.g., `remote.learn_command`). However, the new `add_command` action is not yet available under that alias.

### DIFF
--- a/custom_components/ramses_cc/remote.py
+++ b/custom_components/ramses_cc/remote.py
@@ -128,9 +128,21 @@ class RamsesRemote(RamsesEntity, RemoteEntity):
         data:
           command: boost
           timeout: 3
+          packet_string: "RQ --- 29:162275 30:123456 --:------ 22F1 003 000030"  # optional
         target:
           entity_id: remote.device_id
         """
+
+        # Extract packet_string from kwargs if provided
+        packet_string = kwargs.pop("packet_string", None)
+
+        _LOGGER.debug(
+            "async_learn_command called with: command=%s, timeout=%s, packet_string=%s, kwargs=%s",
+            command,
+            timeout,
+            packet_string,
+            kwargs,
+        )
 
         # HACK to make ramses_cc call work as per HA service call
         command = [command] if isinstance(command, str) else list(command)
@@ -141,6 +153,11 @@ class RamsesRemote(RamsesEntity, RemoteEntity):
 
         if command[0] in self._commands:
             await self.async_delete_command(command)
+
+        # If packet_string is provided, add it directly without listening
+        if packet_string:
+            self._commands[command[0]] = packet_string
+            return
 
         @callback
         def event_filter(event: Event) -> bool:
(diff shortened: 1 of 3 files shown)


silverailscolo:
Thanks. To understand the approach:
The new action is named 'add_command', so it adds it to.. "the system"?

Should a user run that action (can be automated) after every reboot of HA? And where can the user see what has been added, where is it stored?

> Note: All ramses_cc actions are also exposed under the remote namespace (e.g., remote.learn_command).

That only applies for some, because those override built-in hvac actions (annotated in the _cc service.py code).

IMMRMKW:
I currently copied the approach used in `learn_command`, except this action has a field where the user - or an automation - can freely add a string, which is supposed to be a command. Currently, it is not possible to see which commands have been added, nor are they saved in a way that they are still available after a reboot. (Is this also the case for `learn_command`?)

My intentions were for new commands to be available for my fake remote, but indeed, the action is now `ramses_cc.add_command`, so I guess they are available within the integration. The idea is indeed that commands can be added in automated fashion, so that the user does not have to generate big lists of commands himself. The ideal solution would be to just generate commands programmatically, such that these do not have to be saved somewhere, and can be generated real time, but this is a first step towards something like that.

custom_components/ramses_cc/remote.py
```diff
@@ -142,6 +154,11 @@ async def async_learn_command(
         if command[0] in self._commands:
             await self.async_delete_command(command)
 
+        # If packet_string is provided, add it directly without listening
+        if packet_string:
```
silverailscolo, line 158: Not a problem, but a risk, think about adding a warning or extra confirmation step, making users aware they can upset their hardware

tests/tests_old/test_services.py
```diff
@@ -81,7 +81,7 @@
 
 NUM_DEVS_BEFORE = 3  # HGI, faked THM, faked REM
 NUM_DEVS_AFTER = 15  # proxy for success of cast_packets_to_rf()
-NUM_SVCS_AFTER = 10  # proxy for success
```
silverailscolo, line 84: This is to be expected, since some tests compare to a snapshot to check remotes, and some checks count something and compare before/after

tests/tests_old/test_services.py
```diff
@@ -81,7 +81,7 @@
...
+NUM_SVCS_AFTER = 11  # proxy for success (includes add_command service)
```
silverailscolo, line 84: includes a lot more, so please take this (note) out in a follow-up PR. OK for now

silverailscolo (approved):
Interesting new feature

IMMRMKW:
The changes in `learn_command` should not have been pushed, this was my original way of implementing the feature but I foresaw that it was a bit unhandy so I implemented it in a new function later. I cleaned up the code.

PR #324, opened 2025-10-17
Add 'Use SQLite MessageIndex' config switch

silverailscolo:
Adds a switch in config_flow > Known Devices step to (de)activate the experimental SQLite MessageIndex in `ramses_rf` lib.
Depends on `ramses_rf` release 0.52.1 for Schema.
Bumps ramses_cc version to 0.52.1 (skip 0.52.0).

### DIFF
--- a/custom_components/ramses_cc/manifest.json
+++ b/custom_components/ramses_cc/manifest.json
@@ -12,7 +12,7 @@
     "requirements": [
       "aiousbwatcher>=1.1.1",
       "pyserial-asyncio-fast>=0.14",
-      "ramses_rf==0.52.1"
+      "ramses-rf==0.52.1"
     ],
     "single_config_entry": true,
     "version": "0.52.1"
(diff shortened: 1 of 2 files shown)


silverailscolo:
Testing fails as expected until ramses_rf 0.52.0 is on Pypi
`ERROR: No matching distribution found for ramses_rf==0.52.0`

silverailscolo:
Tested loading 0.52.0 `ramses_rf` lib (only) by @peternash:

> At first glance there are some problems with Ramses RF 0.52.0 although there are no errors in the logs:
>
> All the heating zones are showing unavailable
> The DHW control is showing unavailable
> All the Window open sensors are unavailable although the heat demand, battery and OTB sensors appear to be OK
> 5 zone names have reverted to the RAD 01:216136_00 form - the same 5 zones as before
> The Evohome controller device is showing OK and I can change modes and turn it on/off.
>
> All the services appear in Developer Tools and I can, for example, use ramses_cc.set_zone_mode to control a climate device even though it shows as unavailable in the UI.
> 
> Tried clearing the cache but no improvement.

silverailscolo:
Must fix 0.52.0 (Zones) first

PR #333, opened 2025-10-27
`broker`: add a delay when req. all params

wimpie70:
I found the system freeze on a dev system, probably caused by too many messages flooding the system and causing time-outs.
Added a delay of .5sec between RQ in request_all_params and removed a duplicate RQ from added_to_hass.

Didn't notice this happening before. Could be related to my other project that i'm building on top of _cc.


wimpie70:
I suspect the freeze came from my IDE. I was watching the logfile live. That can become quite big.
But this PR is an improvement anyways.

PR #347, opened 2025-11-10
Refactor services w/targets

silverailscolo:
- remove unused target translations
- mark target in FAN params services (partial)  refactor in `broker.py` (see separate PR for those)

### DIFF
--- a/custom_components/ramses_cc/translations/en.json
+++ b/custom_components/ramses_cc/translations/en.json
@@ -220,34 +220,49 @@
         },
 
         "get_fan_param": {
+            "name": "Get Fan Parameter",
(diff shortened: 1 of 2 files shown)


custom_components/ramses_cc/broker.py
```diff
@@ -924,12 +924,15 @@ def _get_device_and_from_id(
         data: dict[str, Any] = call.data if hasattr(call, "data") else call
```
silverailscolo: Would need to also receive the target (address) in this method

wimpie70:
@silverailscolo 
I try to have a look soon. Bit busy getting the other project in a production state, and my normal work.

silverailscolo:
Coverage test failing is for a different reason (outside of this PR)

PR #350, opened 2025-11-14
Fan Param Actions refactor

silverailscolo:
- Use action targets instead of `device_id` field to follow HA (domain hvac + remote) > works
- Adjust SVC schemas > done

TODO:
- @wimpie70 confirm refactor with hardware and front-end

FanParamsActionPaneNew

### DIFF
--- a/custom_components/ramses_cc/number.py
+++ b/custom_components/ramses_cc/number.py
@@ -785,9 +785,14 @@ class RamsesNumberParam(RamsesNumberBase):
             if self._is_boost_mode_param():
                 display_value = round(float(value), 1)
                 self.set_pending(display_value)
-                await self.async_set_fan_param(
-                    param_id=self._normalized_param_id,
-                    value=float(value),
+                await self.hass.services.async_call(
+                    DOMAIN,
+                    "set_fan_param",
+                    {
+                        "device_id": self._device.id,
+                        "param_id": self._normalized_param_id,
+                        "value": float(value),
+                    },
                 )
                 return
 
@@ -806,9 +811,14 @@ class RamsesNumberParam(RamsesNumberBase):
             # Set pending state with the display value
             self.set_pending(display_value)
 
-            await self.async_set_fan_param(
-                param_id=self._normalized_param_id,
-                value=float(value),
+            await self.hass.services.async_call(
+                DOMAIN,
+                "set_fan_param",
+                {
+                    "device_id": self._device.id,
+                    "param_id": self._normalized_param_id,
+                    "value": float(value),
+                },
             )
         except Exception as err:
             _LOGGER.error(


silverailscolo:
This is mostly done, see my remarks above.
Fine to keep your methods in broker, accessible to both climate and remote.
Biggest change: entity service calls provide separate params, not a ServiceCall/dict object, so I  add the device_id and fold that into a dict before passing the call to broker. (simplified that too)

Didn't have to touch `ramses_rf`, but there was a device_id in the form "remote.12:345678" that somehow popped up in the _cc broker get_fan_param. Split it to pass. Use address?

I don't have a fan that wants to cooperate at startup.
Please have a good look and also check how you want to access this part from your .js add on.

custom_components/ramses_cc/climate.py
```diff
@@ -606,6 +606,58 @@ def preset_mode(self) -> str | None:
...
+    async def async_get_fan_param(self, param_id: str = "", from_id: str = "") -> None:
+        """Handle 'get_fan_param' service call.
+
+        :param param_id: The parameter ID of the entity to find
+        :type param_id: str
+        :param from_id: Source device ID (defaults to controller)
+        :type from_id: str
+        """
+        call: dict[str, Any] = {
+            "device_id": self.entity_id,
```
silverailscolo: should be: `self.device_id`

wimpie70:
What I found so far, (running out of time now, but i will continue and create proper commits):
There might be automations already depending on the servicecalls, Not sure if they will still work with the new format. Didn't test that yet. 

broker:
from_id for 2411 RQ/I messages can be a HGI, but when setting (W) a value, with HGI as a source, the FAN won't respond. It must come from device that was bound by the FAN. (defined in config_flow or not). HGI fallback was removed in #346.
I noticed some comments/user feedback still mentions HGI as an option. I'll try to find them all.

get/set_fan_param: fan_id, good catch, that's obsolete code from way back and is obsolete.

wimpie70:
I think i have it working now, but was called afk. Will look at it again after diner

wimpie70:
Why target a REM, Not the FAN ?
When i started on the 2411 params i 1st thought the REM should hold the logic for it, since real devices like a display do that too. but for ha thats not handy at all. The FAN is the device holding the real values. Keep it in 1 place, we can have multiple rem's.
However, with a co2 rem or other measuring rems this could be different for their entities since these can and will depend on their specific locations.

But it could be done. I would itterate over the devices to find a hvacVentilator and check that for the bound REM, if it has any. But maybe only on the fly, not build and store that dict and keep it updated. Depends on how often we expect it to be used i guess.


⁣BlueMail voor Android downloaden ​

Op 18 nov 2025, 22:57, om 22:57, Egbert Broerse ***@***.***> schreef:

silverailscolo:
@wimpie70 Can you verify this works with your plugin scripts? Got this working by making sure:
- the name of climate/remote UI services is not the same as the method on `broker.py`, else it would pickup the request and stumble on the call data.
- split get and set_param for remotes, looking up the "parent" bound device
- updated the translations to match
- not publishing extra services in init, or they fill the Action dropdown

Hope to hear

PR #352, opened 2025-11-14
Cov wf

silverailscolo:
disable Add cov table as comment


silverailscolo:
@wimpie70 These extra checks help to keep test coverage at least at the same level. Failed to automatically insert the report as a PR comment, copied an example from the test output.

When you run
`coverage -m pytest`
locally, followed by
`coverage html`
you get the most useful report to view in a browser.

PR #371, opened 2025-12-16
Pin `aiodns`

silverailscolo:
fix dependency via `pytest_homeassistant_custom_component`, copied from HA core
as found by @PWhite-Eng


silverailscolo:
Ignoring Coverage failure as GitHub runner overrides our dep

PR #395, opened 2026-01-07
fix(schemas): Allow `bound: None` in FAN_TRAITS with strict validation

PWhite-Eng:
**Problem Description** Currently, `ramses_cc` enforces a strict regex (`_SCH_DEVICE_ID`) on the `bound` trait for Fan devices. While this ensures valid device IDs, it inadvertently rejects `None` (or `null` in YAML). This makes it impossible for a user to "unbind" a device via configuration once it has been bound, as the configuration validation fails before the integration can load (and thus before `async_reload` can clear the mapping).

**Proposed Solution** I have updated `custom_components/ramses_cc/schemas.py` to use `vol.Any`:

Python

```
vol.Optional(SZ_BOUND_TO): vol.Any(None, _SCH_DEVICE_ID),

```

This change serves two purposes:
1. **Strictness (Addressing Concerns):** It retains the strict `_SCH_DEVICE_ID` regex for any provided string, preventing "garbage" data or invalid formats from entering the system.
2. **Functionality:** It explicitly allows `None`, enabling the unbind workflow.

**Testing & Safety Verification** To address concerns that allowing `None` might cause "unwanted behavior" or crashes downstream, I have implemented three layers of tests.

**1. Configuration Validation (`test_init_config.py`)**
- **Scenario:** Attempt to load a config with `bound: null` for a Fan.
- **Result (Before Fix):** FAILED. Raises `voluptuous.error.MultipleInvalid: not a valid value... got None`. This proves the current code blocks unbinding.
- **Result (After Fix):** PASSED.

**2. Integration Startup Safety (`test_init_data.py`)**
- **Scenario:** Initialize the `ramses_cc` integration (loading `async_setup_component`) with a `bound: null` device in the `known_list`.
- **Concern Addressed:** Verifies that the initialization logic (broker setup, gateway creation) does not crash when parsing a `None` bound value.
- **Result:** PASSED. The integration loads successfully, and the broker is instantiated without errors.

**3. Runtime Logic Safety (`test_broker_fan.py`)**
- **Scenario:** Attempt to call `set_fan_param` on a Fan device that has no bound remote (simulating the "unbound" state during operation).
- **Concern Addressed:** Verifies that runtime code does not crash (e.g., `AttributeError: 'NoneType' object has no attribute 'id'`) when the expected remote is missing.
- **Result:** PASSED. The code handles the missing binding gracefully without crashing the HA instance.

**Conclusion** This change provides the functionality required to unbind devices while satisfying the requirement for strict data validation. The added tests confirm that `None` propagates safely through configuration, startup, and runtime execution.

### DIFF
--- a/tests/tests_new/test_broker_fan.py
+++ b/tests/tests_new/test_broker_fan.py
@@ -8,6 +8,7 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 from homeassistant.core import HomeAssistant
+from homeassistant.exceptions import HomeAssistantError
 
 from custom_components.ramses_cc.broker import RamsesBroker
 from custom_components.ramses_cc.const import DOMAIN
@@ -330,16 +331,12 @@ async def test_broker_set_fan_param_no_binding(
     # This forces the broker to look for the bound remote
     call_data = {"device_id": FAN_ID, "param_id": PARAM_ID_HEX, "value": 21.5}
 
-    # 3. Expectation: It should NOT crash.
-    # It should likely log a warning or raise a specific friendly error,
-    # but NOT an AttributeError (which would be a crash).
-
-    # Depending on current implementation, it might just log a warning and return
-    # or raise a ValueError. Check broker.py logic.
-    # If the code handles it safely, this test passes.
-
-    # Example assertion (adjust based on actual broker.py behavior):
-    await mock_broker.async_set_fan_param(call_data)
+    # 3. Expectation: It SHOULD raise HomeAssistantError
+    # We use pytest.raises to catch it and verify the message (optional match)
+    with pytest.raises(
+        HomeAssistantError, match="Cannot set parameter: No valid source device"
+    ):
+        await mock_broker.async_set_fan_param(call_data)
 
     # Verify NO command was sent (because there is no source ID)
     mock_gateway.async_send_cmd.assert_not_called()


PWhite-Eng:
Hi @silverailscolo,

I’ve opened this PR as a draft to get your thoughts on a behavior I encountered while updating coverage tests.

I noticed that `ramses_cc` currently re-defines the `bound` trait for FAN devices using a strict validator (`_SCH_DEVICE_ID`). This overrides the upstream `ramses_tx` definition and effectively prevents setting the value to `null` (which is required for unbinding a device).

This PR removes that local override from `FAN_TRAITS` so that it falls back to the library's default, which correctly handles `None`.

Could you let me know if this strict validation was intentional or accidental? If it was intended to be this way, I am happy to close/delete this PR.

Thanks

Phil

wimpie70:
As i said before: Setting bound to None does nothing. Downstream is strict about this. Actually, allowing it to be set to null/none would probably cause unwanted behaviour.

PWhite-Eng:
Hi @wimpie70 and @silverailscolo,

I have updated the **PR Title and Description** to fully reflect the scope of this update.

Based on our previous discussion, I agreed that simply removing the validation was too loose. I have now implemented `vol.Any(None, _SCH_DEVICE_ID)`. This ensures that if a user *does* provide a value, it must strictly match the Device ID regex (preventing invalid strings), but it explicitly allows `None` so that the unbind action can pass the config check.

**Robustness & Safety:** To address the concerns about downstream stability ("unwanted behavior"), I have added three specific test cases to verify the integration is robust and handles `None` without crashing:
1. **Config Test:** Confirms the current master branch completely blocks `bound: null`, making unbinding impossible.
2. **Startup Test:** verified that the integration loads and starts up safely with `bound: null` (preventing the load failures/crashes I encountered during testing when the schema was too strict).
3. **Runtime Test:** Confirms that `set_fan_param` handles an unbound fan gracefully without raising exceptions.

I have verified locally that all tests pass with this solution. This approach meets the requirement for strict validation while ensuring the system is robust enough to handle the unbind state safely.

If you are happy with this approach, I am ready to mark this as ready for review. If you decide you would prefer not to support unbinding via config at all, please let me know and I will happily close the PR.

Thanks, Phil

PWhite-Eng:
@wimpie70 Thanks for the review and the code snippet.

I have updated `broker.py` to integrate your suggested implementation for `_setup_fan_bound_devices`. This is a much cleaner approach.

Regarding testing:
- **Functional Logic:** The core "unbind" logic (`if not bound_device_id: return`) is effectively covered by the new test case in `test_init_data.py` (initializing with `bound: null`), which passes successfully.
- **Defensive Checks:** I reviewed the additional type check (`isinstance(..., str)`). Since `schemas.py` validates the input types before they reach this point, this check is purely defensive/unreachable in normal operation, so I haven't added a dedicated test case for it.

I've verified that all existing tests pass with these changes.

Phil

wimpie70:
Ah, you are correct, This str check won't be triggered anymore in normal operation and it's only defensive against future refactor or validation bypass like edit the .storage. I don't see the use for that atm. Remove or Keep it ? Hmm, I don't think it is harmful to keep it, but I'm fine when it's removed. Test it ? Only for coverage...

PR #397, opened 2026-01-08
feat(number): proactively request parameter values for new entities

PWhite-Eng:
This PR implements proactive state fetching for `number` entities, specifically targeting HVAC (FAN) configuration parameters using the `2411` command class.

**Origin & Discovery:** This issue was discovered while undertaking a significant refactor of the testing suite for `number.py` aimed at extending overall code coverage. During the expansion of unit tests, it became clear that newly added or restored entities remained in an "Unknown" state because the integration relied on passive broadcasts that are infrequent for static fan parameters.

**The Problem:** Historically, fan parameter entities would often initialize without a value. Because these parameters are not actively polled by the `ramses_rf` integration to conserve RF bandwidth and adhere to strict duty cycle limits, a user could see "Unknown" or "Unavailable" in their dashboard indefinitely or until a manual service call was made.

**The Solution:** This change modifies the `add_devices` callback within `number.py` to proactively trigger a state request for every newly added entity.
1. **Staggered Requests:** It utilizes the existing `_request_parameter_value` method, which interacts with the broker to schedule requests.
2. **Duplicate Protection:** Added a robust check to ensure the system only schedules requests for entities that are not already present in the current platform, preventing redundant RF traffic.
3. **Visual Feedback:** By initiating this request immediately upon entity creation, the UI can transition from "Unknown" to a "Pending" state (using the `mdi:timer-sand` icon) and finally to the actual value as soon as the RF packet is received.

### Impact & Why It Is Needed
- **Contextual Rationale:** This is critical for 2411 parameters because they represent device configurations (e.g., boost overrun time, sensor sensitivity) that do not change often and thus are rarely broadcast spontaneously by the hardware.
- **Affected Devices:** This change **exclusively affects HVAC/Ventilation devices** (e.g., Nuaire, Itho, and other RAMSES-compatible fans). It has **zero impact** on traditional central heating (CH) or domestic hot water (DHW) controllers, as they do not use the `2411` parameter trait.
- **Network Safety:** To prevent a "broadcast storm" when a large schema is loaded (e.g., during a system restart with 20+ parameters), the requests are handled via the broker's sequential queueing logic, ensuring one-at-a-time delivery with appropriate delays.

### DIFF
--- a/custom_components/ramses_cc/number.py
+++ b/custom_components/ramses_cc/number.py
@@ -153,6 +153,11 @@ async def async_setup_entry(
             if entities_to_add:
                 _LOGGER.debug("Adding %d new entities directly", len(entities_to_add))
                 async_add_entities(entities_to_add)
+
+                # Request values for these entities too
+                for entity in entities_to_add:
+                    if hasattr(entity, "_request_parameter_value"):
+                        broker.hass.async_create_task(entity._request_parameter_value())
             return
 
         # Otherwise, process as devices and create entities
--- a/tests/tests_new/test_number_extended.py
+++ b/tests/tests_new/test_number_extended.py
@@ -168,3 +168,115 @@ async def test_get_param_descriptions_precision(mock_fan_device: MagicMock) -> N
         descriptions = get_param_descriptions(mock_fan_device)
         assert descriptions[0].precision == 0.1
         assert descriptions[0].mode == "slider"
+
+
+async def test_proactive_state_request_on_add(
+    hass: HomeAssistant, mock_broker: MagicMock, mock_fan_device: MagicMock
+) -> None:
+    """Test that entities proactively request their state when added to the platform.
+
+    This ensures that newly discovered fan parameters don't stay 'Unknown'.
+    """
+    from custom_components.ramses_cc.number import RamsesNumberParam
+
+    # Create a mock entity with a tracked request method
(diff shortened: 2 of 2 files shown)


PWhite-Eng:
@silverailscolo and @wimpie70

Hi both,

While I was undertaking a significant refactor of the testing suite for `number.py` to extend code coverage, I discovered that HVAC/Ventilation parameter entities often initialize in an "Unknown" state upon discovery or system restart.

Because these specific parameters (e.g., sensor sensitivity, boost overrun) are not frequently broadcast by the hardware and the integration avoid active polling to conserve RF bandwidth, they can remain unpopulated in the UI indefinitely.

I have implemented a change that proactively triggers a `_request_parameter_value` call when these entities are first added to the platform. This ensures the UI transitions to a "Pending" state (using the `mdi:timer-sand` icon) and populates with a real value as soon as the RF response is received, significantly improving the initial user experience.

I wanted to check with you both: **Was the current passive initialization an intentional design choice to minimize RF traffic, or is this proactive fetching a welcome addition to the project?**

I have ensured that:
- **Duplicate Protection:** Requests are only sent for entities not already loaded.
- **Network Safety:** The requests leverage the existing staggered sequence logic in the broker to prevent packet collisions.
- **Affected Scope:** This only impacts 2411-compatible HVAC devices and has no effect on heating (CH/DHW) logic.

I would value your feedback on whether you would like me to proceed with this PR.

Phil

PR #398, opened 2026-01-08
refactor(remote): update event_filter to use event_data dictionary

PWhite-Eng:
This PR refactors the RF command "learning" logic within the `remote` platform. It specifically updates how incoming RF packets are filtered during the learning phase to ensure greater reliability, better performance, and strict adherence to Home Assistant’s event-processing standards.

### The Problem

Previously, the `event_filter` within `async_learn_command` was designed to accept a full Home Assistant `Event` object. However, when using the `async_listen` helper on the event bus, the filter logic is often passed the event's data payload directly as a dictionary rather than the container object.

This discrepancy led to two primary issues:
1. **Logic Fragility**: The filter attempted to access attributes that might not exist in the format provided by the event bus caller, potentially leading to silent failures during command learning.
2. **Testing Complexity**: Mocking the learning process required complex nested objects (Events containing Data) which did not accurately reflect the high-speed packet processing of the `ramses_rf` library.

### The Solution

We have updated the `event_filter` signature to strictly expect `event_data: dict[str, Any]`.
- **Direct Payload Access**: The filter now inspects the `src` (Source Device ID) and `code` (RAMSES Opcode) directly from the data dictionary.
- **Strict Filtering**: Only specific HVAC-related RAMSES codes (`22F1`, `22F3`, `22F7`) are permitted to trigger the "Learn" success, preventing unrelated RF traffic from corrupting the command database.
- **Lifecycle Management**: Improved the task cleanup using `contextlib.suppress(asyncio.CancelledError)` to ensure that when a learning session times out or is cancelled, the system state remains clean.

### Impact
- **Reliability**: HVAC remotes will now more accurately "capture" commands during the learning window without interference from temperature sensors or other system packets.
- **Performance**: Reducing the overhead of object inspection inside the filter allows the integration to handle busier RF environments more gracefully.
- **Standardization**: This brings `remote.py` in line with the patterns used in `broker.py` and other core components of the integration.

### Verification & Testing

The new code was validated through additions to  the testing suite, achieving a high degree of confidence:
1. **Signature Validation**: `test_remote_learn_command_success` was updated to verify the filter correctly processes raw dictionaries.
2. **Logic Exhaustion**: Added `test_remote_learn_filter_logic` to check three distinct scenarios:
  - **Valid Packets**: Confirmed `22F1` from the target device returns `True`.
  - **Invalid Source**: Confirmed packets from other devices return `False`.
  - **Invalid Opcode**: Confirmed unrelated codes (like `30C9` temperature data) are correctly ignored even if they come from the correct remote.

### DIFF
--- a/tests/tests_new/test_remote.py
+++ b/tests/tests_new/test_remote.py
@@ -7,7 +7,7 @@ and fan parameter coordination in remote.py.
 from __future__ import annotations
 
 import asyncio
-import contextlib  # Add this import at the top of the file
+import contextlib
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
@@ -148,27 +148,20 @@ async def test_remote_learn_command_success(
     # Prepare the payload dictionary
     learn_payload = {"src": REM_ID, "code": "22F1", "packet": "learned_pkt_123"}
 
-    # We no longer need to wrap this in a MagicMock() representing an Event
-    # because the filter now takes the dict directly.
-
-    # Patch async_listen on the bus instance specifically for this test
-    # instead of using patch.object which fails on read-only attributes.
+    # Patch async_listen on the bus instance to intercept listener registration
     with patch("homeassistant.core.EventBus.async_listen") as mock_listen:
-        # Start learning task
         task = asyncio.create_task(remote.async_learn_command("test_cmd", timeout=1))
 
-        # Give the task a moment to register the listener
+        # Allow the task to register the listener
         await asyncio.sleep(0.1)
 
-        # Verify and trigger the captured listener/filter
         assert mock_listen.called
-        # async_listen call: (event_type, listener, event_filter)
+        # Retrieve the registered listener and filter from the call args
+        # async_listen signature: (event_type, listener, event_filter)
         _, listener, event_filter = mock_listen.call_args[0]
 
-        # Simulate a bus event matching the criteria
-        # Pass the payload directly to the filter and listener
-        if event_filter(learn_payload):  # Pass dict directly
-            # The listener still expects an Event object with a .data attr
+        # Simulate a bus event: the filter accepts a dict, the listener expects an Event
+        if event_filter(learn_payload):
             mock_event = MagicMock()
             mock_event.data = learn_payload
             listener(mock_event)


wimpie70:
I'll pass to comment on this at this point. I didnt use/ look into the command learning so far. But if needed i can dig into this later...

silverailscolo:
@PWhite-Eng did you try this out in real life? I know your HA setup is incomplete, but this type of functionality should be tested with at least one physical fan + remote without errors.

tests/tests_new/test_remote.py
```diff
@@ -7,6 +7,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib  # Add this import at the top of the file
```
silverailscolo: Who's this comment for? Sounds like an AI hint

PWhite-Eng:
Hi @silverailscolo,

You are absolutely right that this needs to be tested in the real world. Unfortunately, I don't currently have the physical fans or remotes to verify this myself.

The primary reason I raised this PR is that the current code setup does not align with how Home Assistant expects these listeners to be implemented. I actually spent a very long time attempting to mock the existing setup to get tests passing, but I failed to successfully test the code because the implementation didn't behave the way HA standard components do.

Stepping back to understand why I couldn't make it work is what led me to realise the setup was incorrect. As soon as I aligned the code to Home Assistant's expectations (switching to `event_data`), the mocking became straightforward.

My aim here is stability and ensuring we can robustly test the code. Without this change, I don't think we can configure mocks to imitate this behavior without a disproportionate amount of effort.

I felt confident enough to submit this for review rather than as a draft because it is now written exactly as Home Assistant expects it to be. However, I agree the physical test is the final piece of the puzzle. Do you know anyone who might have the kit to test this out for us?

Thanks, Phil

tests/tests_new/test_remote.py
```diff
@@ -144,10 +145,11 @@ async def test_remote_learn_command_success(
     )
     remote.hass = hass
 
-    # Prepare real data dictionary
+    # Prepare the payload dictionary
     learn_payload = {"src": REM_ID, "code": "22F1", "packet": "learned_pkt_123"}
-    mock_event = MagicMock()
-    mock_event.data = learn_payload
+
+    # We no longer need to wrap this in a MagicMock() representing an Event
```
silverailscolo: For final code, just describe actual function (feel free to point to HA expected format)

tests/tests_new/test_remote.py
PWhite-Eng: Sorry, missed that inserting the updated code

PR #403, opened 2026-01-09
Fix startup crashes due to corrupted storage and invalid config

PWhite-Eng:
This PR addresses critical stability issues identified during the risk assessment of `broker.py`. It hardens the `async_setup` method to prevent the integration from crashing during startup when encountering corrupted data in the `.storage` file or invalid configuration schemas.

### Problem
1. **Corrupted Storage:** The previous implementation used a dictionary comprehension to parse packet timestamps from the JSON storage file using `dt.fromisoformat()`. If the storage file contained invalid date strings (e.g., due to corruption or manual editing), this raised an uncaught `ValueError`, causing the entire integration to fail startup. The user would be unable to use the integration until manually deleting the hidden storage file.
2. **Unprotected Fallback:** The fallback logic for client creation (if the merged schema failed) was not wrapped in a `try/except` block. If the base configuration was invalid, it resulted in a raw traceback instead of a handled error.

### Solution
1. **Refactored Packet Loading:** Replaced the dictionary comprehension in `cached_packets` with a standard loop wrapped in a `try/except ValueError` block. Invalid dates are now logged as warnings and skipped, allowing the valid packets to be loaded and the system to start up successfully.
2. **Guarded Client Creation:** Wrapped the fallback `_create_client` call in a `try/except` block. It now catches `ValueError` and `vol.Invalid`, logs a specific error message to the HA log, and raises a clean `ValueError` that clearly indicates the initialization failure.

### Testing
- **New Unit Tests:** Added `tests/tests_new/test_broker_startup.py` covering:
  - `test_setup_with_corrupted_storage_dates`: Injects invalid date strings into the mocked storage load and verifies that `async_setup` completes without error, filtering out the bad entry.
  - `test_setup_fails_gracefully_on_bad_config`: Simulates a `vol.Invalid` exception during client creation and verifies that the broker catches it and raises a descriptive `ValueError`.

### Risks
- **Not Implementing:** High risk of "Check Output" loops for users with minor storage corruption. Users are forced to delete their packet cache (losing historical data) to restore functionality.
- **Implementing:** Low risk. The change is purely defensive.
  - *Mitigation:* The `try/except` block only catches `ValueError` specifically for the date parsing, ensuring that other unexpected errors are still surfaced if necessary (though none are expected in that specific conversion). Logging is added to ensure users are aware that some cache data was discarded.

### DIFF
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -169,24 +169,43 @@ class RamsesBroker:
                 _LOGGER.warning("Failed to initialise with merged schema: %s", err)
 
         if not self.client:
-            self.client = self._create_client(config_schema)
+            try:
+                self.client = self._create_client(config_schema)
+            except (ValueError, vol.Invalid) as err:
+                _LOGGER.error(
+                    "Critical error: Failed to initialise client with config schema: %s",
+                    err,
+                )
+                raise ValueError(f"Failed to initialise RAMSES client: {err}") from err
 
         def cached_packets() -> dict[str, str]:  # dtm_str, packet_as_str
             msg_code_filter = ["313F"]  # ? 1FC9
             _known_list = self.options.get(SZ_KNOWN_LIST, {})
-            return {
-                dtm: pkt
-                for dtm, pkt in client_state.get(SZ_PACKETS, {}).items()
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):
Nice fix for user experience, thanks

PR #410, opened 2026-01-11
Fix critical async migration Issues in `climate.py`

PWhite-Eng:
### The Problem:
- Critical methods in `climate.py` (e.g., `async_set_system_mode`, `async_set_zone_mode`) were defined with `@callback` but wrapped synchronous calls to `ramses_rf` that perform database writes.
- `RamsesZone.async_set_hvac_mode` contained a logic error where `HVACMode.AUTO` resulted in a `pass` operation (or an unawaited call), effectively failing to reset the zone mode.

### Consequences:
- **System Crash:** With the new `ramses_rf` SQLite backend, attempting to write to the database from a synchronous `@callback` (which runs on the Main Thread) without yielding or using proper async patterns would likely cause `sqlite3.ProgrammingError` or block the event loop.
- **Silent Failures:** If the underlying library methods were converted to async but not awaited in `climate.py`, the commands would simply not execute.

### The Fix:
- **Refactored Controller & Zone:** Converted all "write" helper methods (`set_mode`, `reset_mode`, `set_config`, `reset_config`) from synchronous `@callback`s to standard `async def` coroutines.
- **Implemented Await:** Added `await` to all calls to the backend library (`self._device.set_mode`, etc.).
- **Updated Entry Points:** Updated standard HA service hooks (`async_set_hvac_mode`, `async_set_preset_mode`, `async_set_temperature`) to `await` the now-async helper methods.
- **Logic Repair:** Fixed `RamsesZone.async_set_hvac_mode` to correctly `await self.async_reset_zone_mode()` when setting `HVACMode.AUTO`.

### Technical Implementation:
- Removed `@callback` decorator from `async_set_system_mode`, `async_reset_system_mode`, `async_set_zone_mode`, `async_reset_zone_mode`, `async_set_zone_config`, `async_reset_zone_config`.
- Changed definitions to `async def`.
- Inserted `await` keyword before `self._device.*` calls and self-calls to helpers.
- Ensured exception handling (`try...except vol.Invalid`) wraps the `await`ed calls.

### Testing Performed:
- **Unit Tests:** Added `tests/tests_new/test_climate_async_fixes.py` mocking `Evohome` and `Zone` devices.
- **Coverage:**
  - Verified `async_set_hvac_mode` calls `set_mode` asynchronously on the device.
  - Verified `async_set_temperature` calls `set_mode` asynchronously with correct arguments.
  - Verified `HVACMode.AUTO` logic calls `reset_mode`.
  - Verified `ServiceValidationError` is raised when `vol.Invalid` occurs during the awaited call.

- Immediate integration failure/crash upon attempting to change temperature or modes due to threading violations with the new SQLite backend.

### Risks of Implementing:
- **Timing Changes:** Changing synchronous callbacks to async coroutines slightly alters the timing of state updates, though `async_write_ha_state_delayed` helps mitigate this.
- **Race Conditions:** If the UI triggers multiple updates rapidly, proper async locking might be needed in the future, though the current implementation is standard for HA.

### Mitigation Steps:
- Strict adherence to Home Assistant's async patterns.
- Comprehensive unit testing mocking the backend to ensure calls are awaited.

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -264,8 +264,7 @@ class RamsesController(RamsesEntity, ClimateEntity):
             )
 
         try:
-            # Removed @callback - now async and awaited to prevent SQLite threading issues
-            self.async_set_system_mode(target_mode)
+            await self.async_set_system_mode(target_mode)
         except vol.Invalid as err:
             raise ServiceValidationError(
                 translation_domain=DOMAIN,
@@ -288,8 +287,7 @@ class RamsesController(RamsesEntity, ClimateEntity):
             )
 
         try:
-            # Removed @callback - now async and awaited
-            self.async_set_system_mode(target_mode)
+            await self.async_set_system_mode(target_mode)
         except vol.Invalid as err:
             raise ServiceValidationError(
                 translation_domain=DOMAIN,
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
@silverailscolo, I missed some of the @callback's in the last PR for climate.py. I think I have got all of them now.
Phil

silverailscolo (approved):
Big step forward

PR #428, opened 2026-01-15
Refactor broker.py for 100% test coverage and improved error handling

PWhite-Eng:
### Summary
This PR achieves 100% test coverage for `broker.py`, this required refactoring some functions in `broker.py` to be more testable and adding code quality improvements, such as removing dead code and standardizing service call normalization. It adds a comprehensive suite of new edge-case tests to achieve 100% test coverage of `broker.py`

#### Code Refactoring (`broker.py`):
- **`_resolve_device_id`:** Simplified logic and removed dead code paths involving empty list checks. Added an inner helper function `_get_first` to adhere to DRY principles and handle list flattening consistently.
- **`_async_run_fan_param_sequence`:** Removed redundant and defensive type-checking logic for dictionary copying, as the input is guaranteed to be a dictionary.
- **`_normalize_service_call`:** Enhanced logic to robustly handle `ServiceCall` objects, dictionaries, and `target` attributes (including `.as_dict()` support).
- **Cleanup:** Moved imports to `const.py` where appropriate.

#### New Tests
Added the following test files to achieve 100% coverage of `broker.py`
- `test_coordinator.py`
- `test_services.py`
- `test_storage.py`
- `test_fan_handler.py`

### Verification
- **Coverage:** `broker.py` now reports 100% coverage when running the full test suite of the 4 listed files above.
- **Validation:** All new and existing tests pass (verified via `pytest`).

Issue #427, opened 2026-01-15
Siber Evo2 fan + 0.53.1: Service actions missing


silverailscolo:
- Dev: Where in the ramses_cc/HA code are these actions registered/filtered?
- User: Confirm this is a persistent issue, still after reinstall
- Related error:
```
2026-01-15 14:02:00.142 WARNING (MainThread) [custom_components.ramses_cc.remote] Error sending command 'low' to device 37:145236: 
2026-01-15 14:03:00.144 WARNING (MainThread) [custom_components.ramses_cc.remote] Error sending command 'low' to device 37:145236: 
```

silverailscolo:
The fix mentioned above solves the issue in 0.53.1.
The cause of the issue was fixed in 0.53.2 (PR #431).

User can remove "extra" bound traits after updating. Closing this sub-issue.

PR #375, opened 2025-12-18
`fan_param` after PR 350

wimpie70:
## ⚠️ Stacked on PR #350
This PR is based on the branch from PR #350 and is intended to be merged *after* #350,
or cherry-picked by the maintainer.

If you prefer, you can cherry-pick just this commit:
- 351ce3d Fix set_fan_param device mapping and service schema

## What this adds
- `ramses_cc.set_fan_param` works with HA UI device targets (HA device registry IDs) and also accepts an optional explicit RAMSES `device_id` in service data.
- Restores `set_fan_param` as a broker/domain service (pre-#350 behavior).
- Adds [services.yaml](cci:7://file:///home/willem/dev/ramses_cc/custom_components/ramses_cc/services.yaml:0:0-0:0) schema + translations (EN/NL) for the updated service.
- set_fan_param: supports HA UI device targeting (and optional explicit RAMSES device_id field).
- Broker fan-param handlers now avoid `dict(call.data)` conversion so `ServiceCall.target` is preserved and HA UI device targets can be resolved by [_normalize_service_call](cci:1://file:///home/willem/dev/ramses_cc/custom_components/ramses_cc/broker.py:1079:4-1098:19).
- This is a small internal refactor (no API change) but required for `set_fan_param` to work from the HA Services UI using device targets.
- Other fan services (get_fan_param, update_fan_params, _clim_, _rem_): other than above...intentionally unchanged in this PR.
- Helpers: added as a potential follow-up refactor to reuse mapping logic across more services, but not used yet.

PR #292, opened 2025-09-10
feat: Add support for 2411 fan parameters

wimpie70:
- Add RamsesNumber entity for handling fan parameters
- Implement parameter reading and writing in broker
- Add services: get_fan_param, set_fan_param, get_all_fan_params
- Add tests for fan parameter functionality
- Add 'bound' trait in config flow for FAN (a bound REM/DIS is needed to set a 2411 param)
- Update config flow to support fan parameter configuration

needs a bit more testing and some cleanup
requires ramses_rf counterpart

PR #429, opened 2026-01-16
Fix/get fan param#427

### LINKED ISSUE #424: Siber: 0.53.1 breaks Fan Actions, Sensors
**Describe the bug**
Ramses RF update 0.53.1 broke Siber FAN sensors (unavailable) and shows only 3 service actions.
See this [HA User Forum post](https://community.home-assistant.io/t/honeywell-ch-dhw-via-rf-evohome-sundial-hometronics-chronotherm/151584/4884)

**To Reproduce**
Steps to reproduce the behaviour (for example):
1. Open Developer Tools > Actions and filter on 'ramses'
2. Only 3 services show up

Please provide code if you can. For example, if you are having an issue with a service call, provide a copy of the service call YAML/

**Expected behaviour**
All hvac services should show up to select from.

**Please include the following information:**
 - `ramses_cc:` your integration **config** (copy from HA > Integrations > Ramses RF > Diagnose > Gateway > Attributes dropdown)
 - contemporaneous portion of the **home-assistant.log**
 - contemporaneous portion of the **packet.log**

**Screenshots**
Only if applicable, add screenshots to help explain your problem. Do not provide a screenshot as an alternative to providing the above information.

**Additional context**
Add any other context about the problem here, such as the make/model of your hardware dongle, interface (USB or MQTT) etc.

wimpie70:
This should fix #424, #427

I found a few commits related to this
c300438 introduced get_fan_param (#292)

but here it went wrong: 15148c4, a04b917 (#375) schemas.py, services.yaml, __init__.py
I think it got commented out /removed when working on clima

This will bring back get_fan_param, with device selection/ramses_id for destination

My guess is that  the user had problems with the other entities being unavailable cause of an error blocking the rest of the polling/actions not available. But we need confirmation on this and may need to fix more.

We now have a lot of fan_param servicecalls...I guess you needed those clima ones...but do you still need them ?
I think it would be good to open a discussion (issue) on how to(re) organize _cc and slowly move into that direction. I can see that _rf goes first. In _extras I  had a lot of work on how separate the features from the framework, you might wanna have a look on how I did that there. Not how things should be done here, but just info...

@PWhite-Eng  I noticed that #428 may have to be adjusted after this PR

### DIFF
--- a/custom_components/ramses_cc/broker.py
+++ b/custom_components/ramses_cc/broker.py
@@ -1204,7 +1204,7 @@ class RamsesBroker:
 
             # Extract id's
             original_device_id, normalized_device_id, from_id = (
-                self._get_device_and_from_id(data) # from_id or bound rem
+                self._get_device_and_from_id(data)  # from_id or bound rem
             )
             param_id = self._get_param_id(data)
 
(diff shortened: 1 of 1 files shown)


silverailscolo:
Thanks @wimpie70 
Continues in #431

And we also learned a branch name should _not_ contain one of the special chars such as #

PR #431, opened 2026-01-16
Fix issue 427: Get `fan_param`

wimpie70:
Replaces #429 
Removed "#" from the branch name

### DIFF
--- a/tests/tests_new/test_fan_param.py
+++ b/tests/tests_new/test_fan_param.py
@@ -24,7 +24,11 @@ from pytest_homeassistant_custom_component.common import (  # type: ignore[impor
 
 from custom_components.ramses_cc.broker import RamsesBroker
 from custom_components.ramses_cc.const import DOMAIN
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
Hi @wimpie70,

I've taken a closer look at your PR changes in remote.py, climate.py, and services.yaml. I see now that #431 is adding the specific service implementations for fan parameters, while my PR #428 is focused on refactoring the broker's device resolution logic.

Proposal: I will updated my broker.py in this PR #428 to include your HGI Gateway fallback logic and your loop safety improvements.

This makes PR #428 the "infrastructure" update that supports your "feature" update.

If you are happy with this, the best path forward is:

We get my PR #428 merged first (it now contains the broker fixes you need).

You then revert/remove the changes to broker.py from your PR #431 (since they are now in mine).

Your PR will then merge cleanly on top, adding the new services and entity methods in remote.py/climate.py.

This keeps the git history clean (Architecture update -> Feature update) and resolves the merge conflict. Does that sound like a plan?

Phil

Issue #414, opened 2026-01-13
RFC: Architectural Refactor of broker.py into a Modular Coordinator System


PWhite-Eng:
### **1. Problem Statement**

The current `broker.py` has become a "God Object" responsible for too many domains:
- **Gateway Orchestration:** Managing the `ramses_rf.Gateway` lifecycle.
- **Persistence:** Handling `homeassistant.helpers.storage.Store` operations.
- **Business Logic:** Managing complex `2411` Fan parameters and device bindings.
- **Service Handling:** Parsing `ServiceCall` objects for packet injection and binding.

This complexity makes unit testing difficult and deviates from Home Assistant's recommended `DataUpdateCoordinator` patterns.

---

### **2. Proposed Modular Architecture**

I propose extracting the logic into four distinct modules while maintaining a temporary "Facade" in `broker.py` for backward compatibility.
- **`coordinator.py`**: The central hub. It will own the `Gateway` and the "State of the World" (devices, zones, systems).
- **`storage.py`**: A dedicated `RamsesStore` class to isolate all JSON packet/schema I/O.
- **`fan_handler.py`**: Encapsulates all Fan-specific logic, parameter sequences, and bound-remote lookups.
- **`services.py`**: Isolated handlers for registering and processing Home Assistant services.

---

#### **Phase 1: Safety Net (Testing)**
1. Consolidate existing tests in `tests/tests_new/test_broker_fan.py` and `test_broker_services.py`, along with additional testing to ensure full coverage of the current `broker.py`. These tests will be split into the following 4 files, aligning with the future modules:
    - `test_coordinator.py`
    - `test_storage.py`
    - `test_fan_handler.py`
    - `test_services.py`
2. Establish baseline "Virtual RF" mocks using `tests/virtual_rf/virtual_rf.py`.

#### **Phase 2: Iterative Extraction & The "Facade" Pattern**
To avoid breaking `climate.py`, `water_heater.py`, and `remote.py`, we will follow a "Wrapper" transition:
1. **Extract Storage:** Create `storage.py`. Update `RamsesBroker` to delegate storage calls to this new class.
2. **Extract Fan Logic:** Create `fan_handler.py`. Move `_setup_fan_bound_devices` and parameter updates there.
3. **Extract Services:** Create `services.py`. Move the handlers for registering and processing Home Assistant services into this new module
[...]

PR #440, opened 2026-01-18
Refactor: Adopt hybrid `DataUpdateCoordinator` for instant updates & reliability

PWhite-Eng:
### The Problem
The current integration architecture relies on a custom update mechanism that may not fully align with modern Home Assistant architectural standards. Relying solely on a custom loop or pure polling can lead to:
1.  **Latency:** State changes in the physical world (e.g., a thermostat turning on) are not reflected in the UI until the next poll interval.
2.  **Maintenance Overhead:** Custom state management logic is harder for other developers to review and maintain compared to the standard `DataUpdateCoordinator`.
3.  **Consistency Issues:** Pure push-based systems can sometimes drift if a packet is missed, without a periodic mechanism to reconcile state or discover new entities.

### Consequences
If this architecture is not modernized:
* **User Experience:** The UI will feel "laggy" or unresponsive to physical changes.
* **Technical Debt:** The codebase diverges further from Home Assistant best practices, increasing the burden on maintainers and making future core updates (like the recent 2025.x entity service changes) harder to adopt.
* **Stability:** Without the robust error handling provided by the `DataUpdateCoordinator`, transport issues could lead to silent failures rather than properly marked "unavailable" entities.

### The Fix
I have refactored the core logic to implement a **"Hybrid" DataUpdateCoordinator pattern**:
1.  **Periodic Reliability:** Implemented `RamsesCoordinator` (subclassing `DataUpdateCoordinator`) which runs a lightweight refresh every 60 seconds. This ensures entity availability and handles new device discovery.
2.  **Instant Responsiveness:** Leveraged the `ramses_rf` packet stream to trigger immediate state updates.
3.  **Standardization:** Migrated all entities (`Climate`, `Sensor`, `BinarySensor`, etc.) to inherit from `CoordinatorEntity`, ensuring they handle availability and lifecycle events correctly.

### Technical Implementation
* **Coordinator (`coordinator.py`):** The `RamsesCoordinator` manages the connection to the `ramses_rf` client. Its `_async_update_data` method is responsible for periodic discovery of new entities, acting as a safety net.
* **Signal Dispatch (`__init__.py`):** The `async_process_msg` callback now intercepts valid incoming RF packets and immediately fires a specific dispatcher signal: `async_dispatcher_send(hass, f"{SIGNAL_UPDATE}_{device_id}")`.
* **Entity Updates (`entity.py`):** The base `RamsesEntity` listens for this device-specific signal. When received, it calls `self.async_write_ha_state()`.
    * **Optimization:** This calls `async_write_ha_state()` directly (reading the latest state from the in-memory `ramses_rf` object) rather than triggering a full coordinator refresh. This prevents "double polling" and ensures updates are instant and low-overhead.
* **Error Handling:** The coordinator setup includes `try/except` blocks for `TransportError` to handle serial port issues gracefully, raising `ConfigEntryNotReady` so HA can retry setup automatically.

### Testing Performed
I have performed extensive testing using the `pytest` suite, verifying all 418 test cases pass.

### Risks of NOT Implementing
* The integration risks becoming obsolete or incompatible with future HA versions that enforce stricter threading or execution models.
* Users will continue to experience delays between physical device changes and UI updates.

### Risks of Implementing
* **Regression in Discovery:** Changing how devices are discovered could potentially miss edge-case devices if they do not broadcast during the initial window.
* **Event Flooding:** In extremely busy RF environments, the 1:1 packet-to-update mapping could theoretically cause high bus traffic.

### Mitigation Steps
* **Throttle Prevention:** The architecture uses `async_write_ha_state()` which is lightweight. The `ramses_rf` library inherently filters duplicate packets, reducing noise.
* **Discovery Safety Net:** The `DataUpdateCoordinator` runs every 60 seconds to catch any devices that were missed during startup or initial packet bursts.
* **Error Guards:** Added explicit `TransportSourceInvalid` and `TransportError` handling during `async_setup_entry` to prevent the integration from crashing the HA startup loop if the hardware is missing.

This PR implements Phase 3, Step 4 (Updated) of the refactoring plan Issue #414.

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -33,8 +33,7 @@ if ENABLE_DEV_HOOK and os.path.isdir(DEV_LIB_PATH):  # pragma: no cover
     )
 # ------------------------
 
-from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Final
+from typing import TYPE_CHECKING, Any
 
 import voluptuous as vol  # type: ignore[import-untyped, unused-ignore]
 from homeassistant import config_entries
@@ -44,17 +43,13 @@ from homeassistant.components.remote import DOMAIN as REMOTE_ENTITY_DOMAIN
 from homeassistant.components.sensor import DOMAIN as SENSOR_ENTITY_DOMAIN
 from homeassistant.components.water_heater import DOMAIN as WATERHEATER_ENTITY_DOMAIN
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import ATTR_ID, Platform
 from homeassistant.core import HomeAssistant, ServiceCall, callback
 from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers import config_validation as cv, service
-from homeassistant.helpers.device_registry import DeviceInfo
-from homeassistant.helpers.dispatcher import async_dispatcher_connect
-from homeassistant.helpers.entity import Entity, EntityDescription
+from homeassistant.helpers.dispatcher import async_dispatcher_send
 from homeassistant.helpers.service import verify_domain_control
 from homeassistant.helpers.typing import ConfigType
 
-from ramses_rf.entity_base import Entity as RamsesRFEntity
 from ramses_tx import exceptions as exc
 
 from .const import (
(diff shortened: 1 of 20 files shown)


silverailscolo:
PR #439 has been merged.

Issue #436, opened 2026-01-17
Improved Error Messaging in broker.py


PWhite-Eng:
**Describe the bug** A misleading warning message is generated in `broker.py` when a destination `device_id` is missing. Instead of alerting the user that the `device_id` is null or missing, the system warns: *"No valid source device available... need explicit from_id or bound..."*. This implies a configuration or binding issue with the source/HGI gateway rather than identifying the missing destination parameter.

**To Reproduce** Steps to reproduce the behaviour:
1. Initiate a service call via `ramses_cc` where the target `device_id` is missing or fails to resolve.

2. Observe the system logs for `broker.py`.

3. See the warning regarding "valid source device" despite the issue being the destination.

**Expected behaviour** The warning should clearly state that the destination `device_id` is missing or invalid before attempting to evaluate source/binding requirements.

**Please include the following information:**
- **`ramses_cc:` config**: *[Placeholder for @wimpie70]*
- **home-assistant.log**: *[Placeholder for @wimpie70]*
- **packet.log**: *[Placeholder for @wimpie70]*

**Screenshots** *Not applicable.*

**Additional context**
- This was identified during the review of PR #428.

- The logic stems from recent refactoring intended to achieve 100% test coverage and improved error handling.

- Improving this will clarify troubleshooting for users without bound remotes who rely on HGI fallback logic.

wimpie70:
It's not a biggy...
In get_fan_param action/servicecall, when entering a false (non-existing) from_id, it gives:
`
Failed to perform the action ramses_cc.get_fan_param. Failed to get fan parameter: Command excluded by device_id filter: RQ --- 18:149487 32:153289 --:------ 2411 003 000031`
When entering a false (non-existing) device_id, it gives a similar warning.

It's ok, only it could be more explainatory (is that a word ??) on what field is wrong.
I think before we had separate warnings, that's why I noticed it.

PR #448, opened 2026-01-20
Fix: Improved error messaging for service calls

PWhite-Eng:
### The Problem:
Currently, when a service call (like `set_fan_param`) is made with a missing or invalid destination `device_id`, the system provides a misleading error message: *"No valid source device available... need explicit from_id..."*. This occurs because the validation logic groups both destination and source checks into a single boolean evaluation, defaulting to a "missing source" warning even when the destination is the root cause (Ref: Issue #436).

### Consequences:
Users are misled into debugging their gateway configuration or binding settings when the actual error is a typo or missing parameter in their service call. This increases troubleshooting time and confusion.

### The Fix:
I have decoupled the validation logic in `services.py` for both `async_get_fan_param` and `async_set_fan_param`. The system now performs a two-step validation:
1. **Check Destination:** Verifies `device_id` exists. If not, raises/logs a specific "Destination missing" error.
2. **Check Source:** Only if the destination is valid does it check for a valid `from_id`.

### Technical Implementation:
- Refactored `RamsesServiceHandler.async_get_fan_param` and `async_set_fan_param`.
- Separated the `if not all([...])` check into distinct `if not original_device_id:` and `if not from_id:` blocks.
- Updated `hass.loop.create_task` calls to `self.hass.async_create_task` to ensure tasks are properly tracked and cleaned up by Home Assistant.
- Standardized exception handling to ensure `ValueError` (internal) is caught and logged, while `HomeAssistantError` is raised to the UI.

### Testing Performed:
- Added specific test cases in `tests/tests_new/test_services.py`:
  - `test_set_fan_param_raises_error_missing_destination`: Verifies the new specific error message.
  - `test_get_fan_param_logs_error_missing_destination`: Verifies the logging behavior for getters.
- Ran full coverage suite: Confirmed **100% coverage** across all files in `custom_components/ramses_cc`.
- Verified that existing "missing source" tests still pass with the updated logic.

The integration remains confusing for users troubleshooting service calls, leading to repeated support questions regarding valid bindings when the issue is actually a simple syntax error.

### Risks of Implementing:
Potential minor regressions if edge-case service calls relied on the "swallow everything" behavior of the previous `all()` check, though unit tests suggest this path is covered.

### Mitigation Steps:
- Comprehensive unit testing (100% coverage) ensures all validation paths are exercised.
- Retained the fallback generic exception handlers to prevent system crashes during unexpected failures.

### DIFF
--- a/custom_components/ramses_cc/services.py
+++ b/custom_components/ramses_cc/services.py
@@ -154,6 +154,13 @@ class RamsesServiceHandler:
             original_device_id, normalized_device_id, from_id = (
                 self._get_device_and_from_id(data)
             )
+
+            # 1. Validate Destination specifically
+            if not original_device_id:
+                msg = f"Cannot get parameter: Destination 'device_id' is missing or invalid in call: {data}"
+                _LOGGER.warning(msg)
+                raise ValueError(msg)
+
             param_id = self._get_param_id(data)
 
             # If no from_id or a bound device was found then try gateway HGI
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
Hi @silverailscolo,

This PR addresses Issue #436 regarding misleading error messages during service calls.

I've split the validation logic in services.py to strictly separate "missing destination" errors from "missing source" errors. While I was in there, I updated the task creation to use hass.async_create_task (per HA best practices) and added new tests to cover these scenarios.

Phil

PR #453, opened 2026-01-20
Fix: Enforce Timezone Awareness to Prevent Comparison Crashes

PWhite-Eng:
### The Problem:
The integration was using naive datetime objects (via standard library `datetime.now()`) in critical logic paths within `climate.py`, `water_heater.py`, and `binary_sensor.py`. Additionally, the upstream library `ramses_rf` can occasionally return naive timestamps. When these naive objects are compared or subtracted from Home Assistant's timezone-aware objects, Python raises a `TypeError: can't subtract offset-naive and offset-aware datetimes`.

### Consequences:
- **System Crashes:** Any logic involving scheduling (e.g., "Boost for 1 hour") or timeouts would crash the integration immediately upon execution.
- **Logic Errors:** Time calculations could be off by the UTC offset amount if naive system time was compared against an aware UTC timestamp without crashing.

### The Fix:
We have migrated the entire integration to enforce **Strict Timezone Awareness**. All internal time generation now uses Home Assistant's `dt_util` helper, and all external data entering the integration is sanitized to ensure it carries timezone information.

### Technical Implementation:
1. **Global Replacement:** Replaced all instances of `datetime.now()` with `homeassistant.util.dt.now()`.
2. **Defensive Boundary:** Added a new helper function `fields_to_aware()` in `helpers.py`. This is now used in `climate.py` and `water_heater.py` to intercept `until` timestamps from the `ramses_rf` library. If a naive timestamp is received, it is automatically converted to the local timezone before the integration attempts to use it.

### Testing Performed:
- **Unit Tests:** Updated `tests/tests_new/test_water_heater.py` and `test_climate.py` to assert against timezone-aware datetimes.
- **Regression Testing:** Ran the full suite (431 tests) with `pytest`, achieving a 100% pass rate.
- **Static Analysis:** `mypy` checks pass with no errors.
- **Coverage:** Confirmed 100% code coverage for the `custom_components/ramses_cc` directory.

Leaving the code as-is guarantees `TypeError` crashes for any user attempting to use Boost, set schedules, or use features that rely on relative time calculations.

### Risks of Implementing:
- **Type Hint Confusion:** There is a minor risk that a specific edge case in the typing (mixing `datetime` class vs `dt` module alias) was missed, though `mypy` suggests this is clean.
- **Upstream Changes:** If `ramses_rf` changes its timestamp format significantly, the `fields_to_aware` helper might need adjustment.

### Mitigation Steps:
- The `fields_to_aware` helper includes a specific check: if the input is already aware, it returns it immediately. If it is naive, it assumes local time. This "catch-all" approach ensures compatibility regardless of whether the library returns UTC or naive local time.

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
-from datetime import datetime as dt, timedelta
+from datetime import timedelta
 from types import UnionType
 from typing import Any
 
@@ -19,6 +19,7 @@ from homeassistant.const import EntityCategory
 from homeassistant.core import HomeAssistant, callback
 from homeassistant.helpers import entity_platform
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
+from homeassistant.util import dt as dt_util
 
 from ramses_rf.device.base import BatteryState, HgiGateway
 from ramses_rf.device.heat import (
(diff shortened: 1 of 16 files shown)


silverailscolo:
Thanks @PWhite-Eng this is a big improvement and not trivial, I know. Making tests pass has cost me nights in the recent past. I cheated on a couple, as you might have found out.

PS Please review #451 as a second pair of eyes when it suits you.

Issue #460, opened 2026-01-22
evohome + 0.53.3: set mode etc. delay increased to 25 sec.


silverailscolo:
Reported by @peternash in HA User Forum:

> I’ve done some testing with Ramses RF integration 0.53.3 on my Evohome system. I’ve noticed a significant change in responsiveness when updating Zone and DHW status in the HA UI. I’ve done this comparison twice now and I’m seeing this consistently.
>
> Note that when making changes to Zones, the setpoint temperature updates almost immediately. The delay is in the **climate** entity updating its status - i.e. _Off_ to _Heat_ or _Temporary_ to _Permanent_.

| | 0.53.2	| 0.53.3 Pre release |
|-|---------|---------------|
| Approx delay between running a script that changes Zone or DHW modes and the HA UI reflecting the change | Within 3 seconds | 20 - 30 seconds |
| Approx delay when changing an override temperature or mode for a Zone in the UI and that change showing in the UI | Within 3 seconds	| 20 - 30 seconds |
| Approx delay when changing DHW mode in the UI and that change showing in the UI | Within 3 seconds	| 20 - 40 seconds |

silverailscolo:
@peternash Is the response time normal again running Ramses RF 0.53.5?

silverailscolo:
@PWhite-Eng I suggest we focus on fixing these `evohome` issues, before any other PRs

silverailscolo:
Fix is in 0.53.6 so closing this.

PR #469, opened 2026-02-02
fix: restore validation error details in config flow

PWhite-Eng:
### The Problem:

A regression was introduced in PR #464 (Fix hassfest validation). As pointed out by @trvrnrth in the review comments, the `async_step_schema` method was overwriting the `description_placeholders` dictionary when injecting the `wiki_url`.

### Consequences:

If a user enters invalid data during the Schema/Traits step (e.g., malformed JSON or an invalid device ID), the integration currently fails to display the specific error message returned by the validator. The `{error_detail}` placeholder is lost, likely resulting in a broken UI string or a generic error, making it difficult for users to understand what they did wrong.

### The Fix:

I refactored `async_step_schema` in `config_flow.py` to preserve the existing `description_placeholders` dictionary.

### Technical Implementation:

Instead of passing a new dictionary literal to `async_show_form`, I now assign the wiki URL to the existing dictionary:

Python

```
# Before (Bug)
return self.async_show_form(
    ...,
    description_placeholders={"wiki_url": ...} # Overwrites errors
)

# After (Fix)
description_placeholders["wiki_url"] = ...
return self.async_show_form(
    ...,
    description_placeholders=description_placeholders # Preserves errors
)

```

### Testing Performed:

The complete pytest suite was ran with all tests passing

Users will encounter poor UX during configuration failures. They will be unable to see specific reasons for rejection (e.g., "Invalid JSON"), leading to frustration and increased support tickets.

### Risks of Implementing:

None. This is a standard dictionary update operation correcting a logical oversight.

### Mitigation Steps:

The change is minimal and strictly limited to the dictionary handling within the schema step.


PWhite-Eng:
@silverailscolo, I saw your comment on issue #460, and I agree the focus will be on fixing the evohome issues. I just wanted to fix this mistake before I started work on looking into the evohome issues.

silverailscolo:
In 2 workflows this test fails:
`ERROR tests/tests_new/test_init.py::test_entities[minimal] - AssertionError: assert (False or False or '_run_safe_shutdown_loop' in 'Thread-278 (_monitor)')`

Issue #465, opened 2026-02-01
Integration fails to initialise correctly under HA 2026.2.0b1


peternash:
> Fixed: **install Ramses RF 0.54.0** or later with HA Core 2026.2.0

**Describe the bug**
After upgrading to the pre-release HA version 2026.2.0b1 the integration fails to create/update climate entities and logs errors repeatedly. Observed with both integration version 0.53.2 and the latest pre-release 0.53.4

**To Reproduce**
Steps to reproduce the behaviour (for example):
1. Update HA to 2026.2.0b1 from the beta channel
2. Climate entities all report `Unavailable` status and multiple errors recorded in the log (see below)

**Please include the following information:**
Log extract:
```
2026-02-01 08:21:37.596 ERROR (MainThread) [homeassistant.components.binary_sensor] Error adding entity binary_sensor.10:064873_ch_active for domain binary_sensor with platform ramses_cc
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 679, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 826, in _async_add_entity
    raise HomeAssistantError(f"Invalid entity ID: {entity.entity_id}")
homeassistant.exceptions.HomeAssistantError: Invalid entity ID: binary_sensor.10:064873_ch_active
2026-02-01 08:21:37.604 ERROR (MainThread) [homeassistant.components.binary_sensor] Error adding entity binary_sensor.10:064873_ch_enabled for domain binary_sensor with platform ramses_cc
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 679, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 826, in _async_add_entity
    raise HomeAssistantError(f"Invalid entity ID: {entity.entity_id}")
homeassistant.exceptions.HomeAssistantError: Invalid entity ID: binary_sensor.10:064873_ch_enabled
2026-02-01 08:21:37.606 ERROR (MainThread) [homeassistant.components.binary_sensor] Error adding entity binary_sensor.10:064873_cooling_active for domain binary_sensor with platform ramses_cc
Traceback (most recent call last):
[...]

PR #471, opened 2026-02-02
Fix startup crash in HA 2026.2.0b1 by enforcing valid Entity ID generation

PWhite-Eng:
### The Problem:

As reported in Issue #465, the integration fails to initialize under Home Assistant 2026.2.0b1 (and likely future versions). The core issue is that the integration was manually assigning `self.entity_id` using raw device IDs that contain colons (e.g., `binary_sensor.10:064873_ch_active`). Stricter validation rules in newer HA versions reject these IDs, causing the platform setup to crash.

### Consequences:

If not fixed, the integration will completely fail to load for any user upgrading to Home Assistant 2026.2.0b1 or later. Users will lose control of their heating system via Home Assistant, and the logs will be flooded with `HomeAssistantError` exceptions regarding invalid entity IDs.

### The Fix:

I have removed the manual assignment of `self.entity_id` in `binary_sensor.py` and `climate.py`. Instead, we now rely on Home Assistant's built-in `EntityRegistry` to automatically generate valid, slugified entity IDs (e.g., transforming `10:064873` into `10_064873`) based on the entity's `unique_id` and name.

### Technical Implementation:
1. **Code Changes:** Removed `self.entity_id = ...` lines in the `__init__` methods of `RamsesBinarySensor` and `RamsesController`.
2. **Test Updates:** Refactored `tests/tests_old/test_evo_control.py` and `tests/tests_old/test_services.py`. These tests previously hardcoded expected Entity IDs with colons (e.g., `climate.01:145038`). They now use a helper function to dynamically look up the correct, slugified `entity_id` from the `unique_id` at runtime, ensuring tests pass regardless of HA's specific slugification logic.
3. **Snapshots:** Updated `tests/tests_new/test_init.py` snapshots to match the new, valid Entity ID format.

### Testing Performed:
- Ran the full `pytest` suite locally.
- Verified that all 438 tests passed, including:
  - `tests/tests_new/test_init.py`: Confirmed that entities are correctly registered with valid IDs.
  - `tests/tests_old/test_services.py`: Confirmed that service calls still work correctly by targeting the new entity IDs.
  - `tests/tests_old/test_evo_control.py`: Confirmed that external interfaces relying on entity attributes remain functional.

The integration becomes unusable for anyone on the beta channel or future stable releases of Home Assistant, leading to a total loss of functionality.

### Risks of Implementing:
- **Breaking Change for YAML/Scripts:** The `entity_id` for existing devices *might* change if users have not renamed them in the UI. For example, `climate.01:123456` might become `climate.ctl_01_123456`. However, because the `unique_id` is preserved, HA should migrate history and settings seamlessly for most users. Scripts hardcoding the old "invalid" IDs (with colons) will need updating.

### Mitigation Steps:
- The `unique_id` was strictly preserved, which is the primary mechanism HA uses to link entities to the database. This ensures that historical data and Lovelace dashboards (which often track by unique ID internally) generally survive the migration.
- The extensive test suite was updated and passed to ensure no logic regressions occurred alongside the ID change.


peternash:
@PWhite-Eng - I noticed that there are similar code patterns in `remote.py` and `number.py` which don't appear to affect my Evohome setup but presumably need the same treatment.

PR #483, opened 2026-02-05
Feature: Add MQTT Transport Layer & Comprehensive Test Suite Refactor

PWhite-Eng:
This PR introduces a new **MQTT Transport Layer**, allowing `ramses_cc` to communicate via Home Assistant's native MQTT service. This decouples the integration from local hardware. 
As part of the works the test suite (ONLY the test in `tests/tests_new`) has been refactored, to align test files with files on a 1-to-1 relationship.

### Technical Implementation:
- **MQTT Bridge:** Implemented `RamsesMqttBridge` and `MqttGateway` classes to act as an intermediary, forwarding packets between the `ramses_rf` library and the HA MQTT bus.
- **Coordinator Updates:** Updated `RamsesCoordinator` to support initialization with an MQTT-based gateway factory when configured.
- **Test Suite Overhaul:**
  - Restructured `tests/test_new` to mirror the `custom_components/` directory structure (1:1 file mapping).

### Testing Performed:
- **Automated Testing:**
  - Ran the full `pytest` suite with **449 tests passed** (100% pass rate).
  - Achieved **100% code coverage** across the new test files.
  - Verified strict type compliance using `mypy` with no issues found in 17 source files.
- **Physical Testing:**
  - Validated the installation on a **Raspberry Pi**.
  - Confirmed successful operation using a **USB dongle** (legacy local transport).
  - Confirmed successful operation using the **HA MQTT service** (new remote transport).

### Risks of Implementing:
- **Transport Regression:** There is a minor risk that changes to the coordinator's initialization logic could impact existing USB-based setups if the factory selection logic is flawed.
- **mqtt Dependency:** Users misconfiguring MQTT in Home Assistant could experience startup failures for the integration if the fallback logic isn't robust.

### Mitigation Steps:
- **Dual-Path Testing:** I have physically verified that both USB and MQTT transport methods function correctly on actual hardware (Raspberry Pi).
- **Comprehensive Mocks:** The new test suite explicitly tests the coordinator's ability to handle transport initialization failures (e.g., `TransportError`, `ConfigEntryNotReady`) ensuring the system fails gracefully rather than crashing.
- **Coverage:** 100% test coverage ensures that all new branching logic for transport selection has been exercised.

### DIFF
--- a/tests/tests_new/test_config_flow.py
+++ b/tests/tests_new/test_config_flow.py
@@ -29,6 +29,7 @@ from custom_components.ramses_cc.config_flow import (
 from custom_components.ramses_cc.const import (
     CONF_MESSAGE_EVENTS,
     CONF_MQTT_HGI_ID,
+    CONF_MQTT_TOPIC,
     CONF_MQTT_USE_HA,
     CONF_RAMSES_RF,
     CONF_SCHEMA,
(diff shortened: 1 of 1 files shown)


PWhite-Eng:
Thanks for the feedback @silverailscolo

**Guard Logic Update** I did have a guard in place in the original submission, but upon reviewing your comment, I realised why it wasn't effective (and why it might have looked like it was missing entirely).
- **The Issue:** My original code checked `hass.config_entries.async_entries("mqtt")`. This only confirmed that an MQTT configuration *existed*, but it returned `True` even if the integration was not actually running (e.g., if it was in a failed state or the broker was disconnected).
- **The Fix:** I have updated the `config_flow.py` to strictly check for `ConfigEntryState.LOADED`.
  - The option now explicitly distinguishes between "Not Found" and "Not Ready" (configured but not loaded).
  - It strictly prevents selection unless the MQTT integration is fully active and operational.

**Testing & Verification**
- **Unit Tests:** I updated `tests/tests_new/test_config_flow.py` to mock the various config entry states (`LOADED` vs `SETUP_ERROR`). We have maintained **100% code coverage** on these new checks.
- **Manual Testing:** I verified the changes on two environments:
  1. **Main HA Server:** The guard correctly detected my active MQTT integration and allowed setup.
  2. **Test Bench:** The guard correctly handled the state. Interestingly, my fresh test bench installation came pre-configured with a running MQTT service, and the updated guard correctly identified it as valid immediately.

Ready for re-review

silverailscolo (approved):
Runs nice on my HA kit too, thanks

PR #488, opened 2026-02-08
fix: MQTT Transport Interoperability & Handshake Logic

PWhite-Eng:
### The Problem:

The current MQTT bridge implementation fails on standard `ramses_esp` firmware builds.
1. **Type Mismatch:** The firmware returns command results as **integers** (e.g., `{"return": 0}`), but the bridge expects **strings**, causing a crash during type operations.
2. **Missing Handshake:** `ramses_rf` requires a specific response string (starting with `# evofw3`) to the `!V` command to verify the gateway. Standard firmware often returns only an exit code, causing `ramses_rf` to abort initialization.
3. **TX Buffering:** Outgoing packets sent to the `/tx` topic were missing the `\r\n` terminator inside the JSON payload, causing some firmware versions to buffer the packet indefinitely instead of transmitting it.

### Consequences:
- **System Failure:** The integration successfully receives packets (RX) but fails to initialize the gateway or send commands (TX), rendering it effectively read-only.
- **Crash Loop:** Any attempt to send a command results in an unhandled `TypeError` in the event loop when the acknowledgement is received.

### The Fix:

Updated `mqtt_bridge.py` to be robust against different firmware response types and framing requirements. It now "polyfills" missing features from the firmware to satisfy the strict requirements of the `ramses_rf` library.

### Technical Implementation:
- **TX Framing:** In `mqtt_packet_sender`, the JSON payload generation was updated to explicitly append `\r\n` to the frame string (e.g., `{"msg": "FRAME\r\n"}`).
- **Type Handling:** In `_handle_cmd_message`, added logic to check if `data["return"]` is an `int`.
- **Handshake Synthesis:** If the bridge detects a successful integer response (`0`) to a `!V` (version) command, it synthesizes a fake response frame (`# evofw3 0.1.0`) and injects it into the transport. This tricks `ramses_rf` into accepting the gateway connection.
- **Fallback:** For other integer returns, they are safely cast to strings before processing.

### Testing Performed:
- **Unit Tests:** Updated `tests/tests_new/test_mqtt_bridge.py` to match the new framing requirements.
- **New Test Case:** Added `test_bridge_handle_cmd_result_int` to specifically test the integer handling path. It verifies that a `0` return from `!V` generates the correct `evofw3` signature, and that other integer returns are converted to strings.
- **Coverage:** Verified **100% code coverage** for `mqtt_bridge.py` via pytest-cov.

The MQTT transport feature will remain broken for the majority of users running standard `ramses_esp` hardware.

### Risks of Implementing:

**Low Risk.** The primary risk is masking actual firmware faults by synthesizing the handshake. If the firmware is actually broken/disconnected but returns `0`, `ramses_rf` might think it's connected when it isn't.

### Mitigation Steps:
- The handshake synthesis is **conditional**: it only triggers if the firmware explicitly reports success (`return: 0`).
- The code retains support for the original string-based responses, ensuring backward compatibility with custom firmware builds that already work.

### DIFF
--- a/custom_components/ramses_cc/mqtt_bridge.py
+++ b/custom_components/ramses_cc/mqtt_bridge.py
@@ -70,7 +70,7 @@ class RamsesMqttBridge:
             else:
                 # Wrap in JSON for the /tx topic as per ramses_esp expectation
                 try:
-                    json_payload = json.dumps({"msg": frame})
+                    json_payload = json.dumps({"msg": frame + "\r\n"})
                     _LOGGER.debug("MqttTransport: TX (frame) -> %s", json_payload)
                     self.publish_tx(json_payload)
                 except TypeError as err:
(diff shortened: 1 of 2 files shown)


silverailscolo:
I think we should merge this, as it fixes master. But that leaves unresolved the Port > HA MQTT issue.

PR #499, opened 2026-02-19
Refactor imports to match ramses_rf, remote actions feedback in UI

silverailscolo:
- Bump version, dependency
- Imports:
   - store send_cmd consts in ramses_rf.const, direct imports to remote
   - sensor imports via _rf
   - fix refactored ramses_rf DeviceIdT path

- Throw HomeAssistantErrors, not WARN, in remote actions, showing as feedback in UI
- Check for remote.send_command use and warn in log that entry is uncontrolled (it might fail assert in ramses_tx.protocol)
- Adjust remote.tests

Verified in local HA kit.

### DIFF
--- a/custom_components/ramses_cc/remote.py
+++ b/custom_components/ramses_cc/remote.py
@@ -120,7 +120,7 @@ class RamsesRemote(RamsesEntity, RemoteEntity):
         # HACK to make ramses_cc call work as per HA service call
         command = [command] if isinstance(command, str) else list(command)
         # if len(command) != 1:
-        #     raise TypeError("must be exactly one command to delete")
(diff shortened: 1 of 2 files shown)


custom_components/ramses_cc/remote.py
```diff
@@ -26,6 +26,7 @@
 from ramses_tx import DeviceIdT
```
PWhite-Eng: @silverailscolo, I think this one should also be changed to:
from ramses_tx.typing import DeviceIdT

silverailscolo: `DeviceIdT` happens to be imported into ramses_tx init, but it seems more consistent to use same route as used in fan_handler

Issue #532, opened 2026-03-14
0.55.5: RamsesGatewayBinarySensor unavailable


silverailscolo:
Since updating via HACS TO 0.55.5, the gateway binary sensor for my ramses_esp is unavailable after several restarts:

This is the `RamsesGatewayBinarySensor(RamsesBinarySensor)`.

Error in log after rebooting bridge:

```
Logger: ramses_tx.transport.mqtt
Bron: /usr/local/lib/python3.14/site-packages/ramses_tx/transport/mqtt.py:330
Eerst voorgekomen: 13:10:41 (1 gebeurtenis)
Laatst gelogd: 13:10:41

MqttTransport(QosProtocol(IsInIdle, len(queue)=0)): the ESP device is offline (via LWT): RAMSES/GATEWAY/18:130140
```

System log after restart:

[ramses_cc_0555_startuplog.txt](https://github.com/user-attachments/files/25994501/ramses_cc_0555_startuplog.txt)

See no errors in log, but the RamsesGatewayBinarySensor is not created:
```
[32m2026-03-14 11:03:59.362 INFO (MainThread) [homeassistant.components.binary_sensor] Setting up ramses_cc.binary_sensor[0m
[36m2026-03-14 11:03:59.363 DEBUG (MainThread) [custom_components.ramses_cc.coordinator] Registering platform binary_sensor[0m
[36m2026-03-14 11:03:59.363 DEBUG (MainThread) [custom_components.ramses_cc.coordinator] Connecting signal for platform binary_sensor: ramses_cc_new_devices_binary_sensor[0m
[36m2026-03-14 11:03:59.366 DEBUG (MainThread) [custom_components.ramses_cc.coordinator] Platform setup completed for binary_sensor[0m
[32m2026-03-14 11:03:59.367 INFO (MainThread) [custom_components.ramses_cc.binary_sensor] Found 29:123150: battery_low[0m
[32m2026-03-14 11:03:59.368 INFO (MainThread) [custom_components.ramses_cc.binary_sensor] Found 29:099029: bypass_position[0m
[32m2026-03-14 11:03:59.368 INFO (MainThread) [custom_components.ramses_cc.binary_sensor] Found 37:153226: bypass_position[0m
[32m2026-03-14 11:03:59.369 INFO (MainThread) [custom_components.ramses_cc.binary_sensor] Found 32:022222: bypass_position[0m
[...]

PR #535, opened 2026-03-14
fix bridge status discovery gap

wimpie70:
Title: Fix timing issue in resolve_async_attr and restore gateway status discovery

The code follows [HA Architecture Rules](https://developers.home-assistant.io/docs/architecture_components?_highlight=integrations)
I have reviewed and tested these code changes myself, successfully ran pre-commit hooks and am available to address review comments.
AI Coding Assistance disclosure: used for drafting/implementation assistance
PR Summary

This PR fixes a timing issue in resolve_async_attr() that could trigger async_write_ha_state() before Home Assistant had assigned an entity_id to the entity.

Problem
After the recent async-resolution compatibility changes, some entities could resolve background async attributes during setup and immediately write state back to HA. If this happened before the entity had been fully added, HA raised errors like:

No entity id specified for entity Battery
No entity id specified for entity Bypass position
At the same time, the gateway/bridge status binary sensor was no longer appearing reliably after startup.

Root cause
There were two related issues:

resolve_async_attr() updated HA state too early during entity setup
coordinator discovery relied on device_registry.devices, but the active HGI/gateway device was not always materialized there early enough to create the gateway status sensor
Changes
1. Guard early state writes in resolve_async_attr()
Only call entity.async_write_ha_state() when the entity already has an entity_id.

This prevents setup-time async resolution from causing HA entity lifecycle errors.

2. Ensure active HGI is materialized during discovery
In RamsesCoordinator._discover_new_entities():

read the active HGI ID from transport extra info or engine fallback
if it is a valid RAMSES device ID and is not yet present in the device registry, explicitly create/materialize it
This restores discovery of the gateway/HGI device so the Gateway status binary sensor can be created again.

Result
no more setup-time No entity id specified ... errors from async attribute resolution
gateway/bridge status sensor is discoverable again
binary sensor setup remains compatible with the recent ramses_rf changes
Testing
Tested locally with:

targeted binary sensor / discovery tests
full _cc local CI
Results:

21 passed for focused binary sensor/coordinator/init tests
462 passed
3 snapshots passed
make local-ci passed
Related issue
This PR fixes issue #532 - bridge issue

Status
PR text drafted
Ready to paste into the PR form

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import asyncio
 import inspect
 import logging
+import re
 from collections.abc import Callable, Coroutine
 from contextlib import suppress
 from copy import deepcopy
@@ -34,7 +35,7 @@ from ramses_rf.entity_base import Entity as RamsesRFEntity
 from ramses_rf.gateway import Gateway, GatewayConfig
 from ramses_rf.system import Evohome, System, Zone
 from ramses_rf.topology import Child
-from ramses_tx.const import Code
+from ramses_tx.const import SZ_ACTIVE_HGI, Code
 from ramses_tx.schemas import extract_serial_port
 
 from .const import (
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):
Works great on my dongle, thanks

PR #555, opened 2026-03-23
Fix 'duplicate id' errors on numbers (2411_params)

wimpie70:
This PR rolls back the slug prefix naming that was also rolled back for other entities.

Without this PR we can see duplicate id Errors on FAN devices.
It also includes a migration to rename existing entities to the format without slug prefix.

c26a835 removed manual entity_id handling for other platforms like binary_sensor.py and climate.py, letting HA generate IDs.
c27d6e2 rolled back binary sensor changes.
In number.py, the fan_ object-id prefix came from ed5d894:
suggested_object_id = f"{slug_prefix}{device_id}_param_{param_id.lower()}"

### DIFF
--- a/custom_components/ramses_cc/number.py
+++ b/custom_components/ramses_cc/number.py
@@ -114,7 +114,7 @@ def _migrate_legacy_param_entity_ids(hass: HomeAssistant) -> None:
     if entries is None:
         entries = getattr(registry, "_entities", {})
 
-    values = entries.values() if hasattr(entries, "values") else ()
+    values = list(entries.values()) if hasattr(entries, "values") else []
     for entity in values:
         if getattr(entity, "domain", None) != "number":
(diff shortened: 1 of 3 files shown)


silverailscolo (dismissed):

wimpie70:
As far as i can tell from here (phone) all entities except numbers got the slug prefix on ids.
So i made a PR to do this for numbers as well.
Then I noticed the slug was gone, so I made this PR  to follow again with numbers. Including a downgrade fallback.

If you want the slug prefix (can be handy) then don't merge this, but we should still have fallbacks to both naming conventions since there is quite a bit up- and downgrading going on.
First I wanted to solve this with a migration helper, to rename id's...maybe we should have that anyways.

silverailscolo (approved):

Issue #537, opened 2026-03-15
0.55.5: BDR91, faked temperature sensors freeze


silverailscolo:
Reported in HA user forum by Wilco Slegers and @peternash :

> Having issues with ~0.55.3 and~ 0.55.5:
> When I install it my faked temperature sensors are stuck on the last reported temperature and become unavailable after a reboot, as do some BDR91 readings.
> 
> Had to revert back to 0.55.3 to get it working again.

PR #562, opened 2026-03-29
Refactor Gateway and System binary sensors to use async state model (#537)

PWhite-Eng:
### The Problem:

Currently, the integration relies on synchronous attribute polling to determine Gateway and System status. This bypasses the Home Assistant async event loop and evaluates missing data as False. As a result, the integration falsely reports an "All Clear" (off) state the millisecond it boots up, before any actual telemetry is received from the devices. This addresses Issue #537 (Add status to gateway device).

### Consequences:

If this isn't fixed, the Home Assistant UI will flash incorrect/stale network states during initialization. Additionally, keeping the old synchronous calls will break compatibility with the newly refactored ramses_rf underlying library, leading to silent state evaluation failures or UI freezing.

### The Fix:

We refactored RamsesGatewayBinarySensor, RamsesSystemBinarySensor, and RamsesLogbookBinarySensor to strictly rely on the integration's resolve_async_attr helper and evaluate availability based on packet timestamps.

### Technical Implementation:
- **State Resolution:** Updated BINARY_SENSOR_DESCRIPTIONS to target the newly refactored is_active property from ramses_rf.
- **Availability Logic:** Modified .available properties to query self._device.state_store._msgs_ directly, calculating time deltas between Home Assistant's dt_util.now() and the dtm of Code._0418 or Code._1F09 telemetry.
- **Initial States:** Modified is_on to return None when async states are unresolved. This allows Home Assistant to correctly display unknown on boot rather than incorrectly defaulting to off (No problem detected).
- **Test Modernization:** Updated test_binary_sensor.py to mock state_store packet dictionaries instead of overriding the legacy boolean flags.
- **Snapshot Alignment:** Patched the strictly-overridden .available properties in test_init.py and regenerated the snapshots to capture the safer startup logic and recent HA core feature sets.

### Testing Performed:
- Executed the full pytest suite for ramses_cc (472 passed, 0 failed, 8 skipped).
- Verified strict type compliance using mypy (0 errors across 18 source files).
- Validated that extra_state_attributes gracefully hide themselves when the entity correctly shifts to unavailable or unknown.

Failing to adopt this change leaves the integration tightly coupled to deprecated internal variables of the ramses_rf engine. It will cause direct crashes or silent data desyncs when paired with the newest versions of the ramses_rf backend.

### Risks of Implementing:

Users monitoring the exact startup sequence might notice sensors initialize as unknown before transitioning to on/off. Poorly written user automations that trigger strictly on "State changed from X to off" without ignoring unknown might experience minor hiccups during reboot events.

### Mitigation Steps:

Implemented robust dictionary fallbacks (e.g., getattr(gwy, "_exclude", {})) to ensure edge-case configurations don't throw KeyError or AttributeError exceptions during the critical startup phase. Used strict Mypy typing to guarantee runtime safety.

### AI Assistance Disclosure:

---

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -41,7 +41,7 @@ from ramses_rf.entity_base import Entity as RamsesRFEntity
 from ramses_rf.gateway import Gateway
 from ramses_rf.schemas import SZ_BLOCK_LIST, SZ_CONFIG, SZ_KNOWN_LIST, SZ_SCHEMA
 from ramses_rf.system.heat import Logbook, System
-from ramses_tx.const import SZ_BYPASS_POSITION, SZ_IS_EVOFW3, Code
+from ramses_tx.const import SZ_BYPASS_POSITION, SZ_IS_EVOFW3
 
 from .const import (
     ATTR_ACTIVE_FAULTS,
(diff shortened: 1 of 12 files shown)


silverailscolo (approved):
Wait for review comments in ramses_rf PR addressed to release hold > Done, released

silverailscolo:
@PWhite-Eng Thanks, this will be merged as is.
For the gateway binary_sensor I prefer to override `available()` so that any user always has the system schema under details. That is part of PR #559 coming next.

PR #570, opened 2026-03-31
Fix: Resolve 8-hour climate timeouts and delegate binary sensor availability

PWhite-Eng:
### The Problem:

As reported by users (Peter Nash), climate devices are dropping to an Unknown state after exactly 8 hours, despite telemetry still updating history graphs. Additionally, the Evohome controller fault sensor and OpenTherm bridge sensors are incorrectly evaluating as Unavailable. This is caused by a "split-brain" state: climate entities rely on system_mode packets that expire in 8 hours, and certain binary sensors retained hardcoded 20-minute expiry checks that override the protocol library's heartbeat logic. This also addresses test modernization requests from PR #559.

### Consequences:

If this isn't fixed, perfectly healthy climate entities will continually drop offline in Home Assistant, breaking automations and corrupting state history. Fault sensors will falsely report as unavailable, masking genuine system errors and causing significant user confusion.

### The Fix:

We implemented a fallback in the climate entities to evaluate active heat demand if the system mode data expires, keeping the entity alive. We also stripped the legacy timeout overrides from the binary sensors, delegating health checks entirely to the updated ramses_rf library. Tests were modernized to strictly mock the async boundaries.

### Technical Implementation:
- **Climate Fallbacks (climate.py):** Modified hvac_action and preset_mode in RamsesController and RamsesZone. If resolve_async_attr for system_mode returns None, the logic now safely falls back to checking heat_demand instead of prematurely defaulting to an Unknown state.
- **Availability Delegation (binary_sensor.py):** Deleted the localized .available property overrides in RamsesLogbookBinarySensor and RamsesSystemBinarySensor. They now cleanly inherit the base RamsesEntity.available logic.
- **Test Modernization (test_binary_sensor.py & test_climate.py):** Replaced dictionary duck-typing with @patch("custom_components.ramses_cc.binary_sensor.resolve_async_attr") to strictly mock the Home Assistant async integration boundary. Injected heat_demand = MagicMock(return_value=None) to satisfy the new climate fallback assertions.

### Testing Performed:
- Ran the full pytest suite (100% pass rate achieved after updating test mocks).
- Executed pytest --snapshot-update to align historical packet logs with the newly delegated availability states.
- Verified strict Mypy compliance (0 errors across all modified files).

Users will continue to experience "fragile" entities dropping offline daily. The integration will remain architecturally entangled with legacy sync cycles rather than fully utilizing the protocol library's dynamic heartbeat engine.

### Risks of Implementing:

There is a minor risk of entities appearing "stuck" in a heating state if the underlying library fails to clear the heat_demand attribute when a device genuinely loses power right after broadcasting a demand packet.

### Mitigation Steps:

Relied strictly on resolve_async_attr to ensure we are only reading data validated by the ramses_rf async event loop. Used strict None checks in the fallback logic to prevent AttributeErrors or false positives during the critical startup and evaluation phases.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
-from datetime import timedelta as td
 from types import UnionType
 from typing import Any
 
@@ -18,7 +17,6 @@ from homeassistant.const import EntityCategory
 from homeassistant.core import HomeAssistant, callback
 from homeassistant.helpers import entity_platform
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
-from homeassistant.util import dt as dt_util
 
 from ramses_rf.device.base import BatteryState, HgiGateway
 from ramses_rf.device.heat import (
@@ -41,7 +39,7 @@ from ramses_rf.entity_base import Entity as RamsesRFEntity
 from ramses_rf.gateway import Gateway
 from ramses_rf.schemas import SZ_BLOCK_LIST, SZ_CONFIG, SZ_KNOWN_LIST, SZ_SCHEMA
 from ramses_rf.system.heat import Logbook, System
-from ramses_tx.const import SZ_BYPASS_POSITION, SZ_IS_EVOFW3, Code
+from ramses_tx.const import SZ_BYPASS_POSITION, SZ_IS_EVOFW3
 
 from .const import (
     ATTR_ACTIVE_FAULTS,
@@ -168,24 +166,6 @@ class RamsesLogbookBinarySensor(RamsesBinarySensor):
     _device: Logbook
 
     @property
-    def available(self) -> bool:
-        """Return True if the device has been seen recently.
-
-        :return: Availability status based on recent messages.
-        :rtype: bool
-        """
-        if hasattr(self._device, "state_store"):
-            msg = self._device.state_store._msgs_.get(Code._0418)
-        else:
-            msg = getattr(self._device, "_msgs", {}).get(Code._0418)
-
-        if not msg:
-            return False
-
-        msg_time = dt_util.as_utc(msg.dtm)
-        return bool(dt_util.now() - msg_time < td(seconds=1200))
-
-    @property
     def is_on(self) -> bool | None:
         """Return the state of the binary sensor.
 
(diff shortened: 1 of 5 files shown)


peternash:
@PWhite-Eng and @silverailscolo - running with "master" including [PR #579](https://github.com/ramses-rf/ramses_cc/pull/579) the DHW device has remained "known" for over 8 hours so that looks like a fix.  Still got some OpenTherm sensors showing `Unknown`.  The Evhome controller fault entity `binary_sensor.01_216136_active_fault` is also still `Unknown`

Also, I've spotted one new issue with 0.55.8/master:

After a restart, the `system_mode` attribute of the Evohome controller entity is `null`.  It doesn't populate with the correct value until the system mode has been changed (or re-set to the current mode).  It then tracks mode correctly until the next HA restart.  This one is a little troublesome since some of my automations check the current mode before they take action.

Looking at the state of the entity `climate.01_216136` this is what I see after a restart under 0.55.8:

```
hvac_modes:
  - heat
  - "off"
  - auto
min_temp: null
max_temp: null
preset_modes:
  - none
  - away
  - custom
  - home
  - eco
current_temperature: 17.9
hvac_action: heating
preset_mode: none
id: "01:216136"
heat_demand: 1
heat_demands:
  FC: 1
relay_demands:
  FC: 1
  FA: 0
  F9: 1
system_mode: null
tpi_params:
  cycle_rate: 6
  min_on_time: 2
  min_off_time: 0
  proportional_band_width: null
icon: mdi:thermostat
friendly_name: Evohome controller
supported_features: 400
```

Note that `system_mode` is `null`.  Here is what I see under earlier versions or _after_ a controller mode change under 0.55.8:

```
hvac_modes:
  - heat
  - "off"
  - auto
min_temp: null
max_temp: null
preset_modes:
  - none
  - away
  - custom
  - home
  - eco
current_temperature: 17.9
hvac_action: heating
preset_mode: none
id: "01:216136"
heat_demand: 1
heat_demands:
  FC: 1
relay_demands:
  FC: 1
  FA: 0
  F9: 1
system_mode:
  system_mode: auto
tpi_params:
  cycle_rate: 6
  min_on_time: 2
  min_off_time: 0
  proportional_band_width: null
icon: mdi:thermostat
friendly_name: Evohome controller
supported_features: 400

```

Happy to provide logs. Do you want me to track these under a new issue?

Issue #572, opened 2026-03-31
Sensor Heartbeat too short?


silverailscolo:
Since 0.55.7 and latest PRs, my HVAC `31DA` sensor is updating only 1x per 30 mins.
- MQTT ramses_esp gateway.
- MessageIndex turned on.

I see `31DA` messages in the log every ~6 minutes, and the graph until yesterday showed a more detailed line:

### 0.55.7:
![image](https://github.com/user-attachments/assets/8a1439fc-430a-4876-ba1e-e70014c21f87)

### 0.55.6:
![image](https://github.com/user-attachments/assets/80fa8137-4ec8-421d-aaea-647c42b87f24)

After customisable gateway timeout, set to 10 minutes, still an issue with SQLite MessageIbdex on. Better with SQL off.

PR #574, opened 2026-04-01
Feat: Expose configurable gateway timeout in UI (#572)

PWhite-Eng:
### The Problem:
Resolves Issue #572 where entities belonging to less chatty devices (like HVAC 31DA sensors broadcasting every ~6 minutes) suffer from state-dropping in Home Assistant. This is caused by the backend library's hardcoded 5-minute gateway timeout, which frequently flags the integration as inactive during expected quiet periods on sparse networks.

### Consequences:
Without this fix, users with battery-powered or sparse RF/MQTT environments will experience erratic entity availability. The Home Assistant history recorder drops intermediate state changes when the entity flaps to `unavailable`, resulting in flatlined 30-minute graphs instead of high-resolution tracking.

### The Fix:
This PR exposes a "Gateway timeout" parameter in the Home Assistant configuration flow. It allows users to manually define their own keep-alive threshold to match the polling or heartbeat frequency of their specific hardware setup. 

*Note: This is a complimentary PR to `ramses-rf/ramses_rf#548` (feat: Add configurable gateway timeout and increase default to 10 minutes (#572)). If this frontend PR is not merged, the backend will still safely default to its new 10-minute hardcoded fallback.*

### Technical Implementation:
* **`custom_components/ramses_cc/const.py`:** Defined `CONF_GATEWAY_TIMEOUT`.
* **`custom_components/ramses_cc/config_flow.py`:** Added an integer `NumberSelector` (1 to 60 minutes) to `async_step_config` with a default of 10.
* **`custom_components/ramses_cc/coordinator.py`:** Updated `_create_client()` to extract the timeout from the HA options registry and route it via `route_special_arg` directly into the backend `GatewayConfig` DTO.
* **`translations/`**: Added UI localization keys and descriptions for both `en.json` and `nl.json`.

### Testing Performed:
* **`tests/tests_new/test_config_flow.py`:** Updated flow simulations to inject and assert the persistence of the new `CONF_GATEWAY_TIMEOUT` parameter.
* **`tests/tests_new/test_coordinator.py`:** Updated the `FakeGatewayConfig` dummy dataclasses to accept the new parameter, preventing `TypeError` during simulated client creation. Verified `coordinator.py` correctly extracts and routes the option.
* Verified 100% pass rate across the `pytest` suite and zero errors on `mypy --strict`.

If the complimentary backend PR (#548) is merged without this frontend PR, users will simply be locked to a new 10-minute hardcoded default. While this fixes Issue #572, it removes the user's autonomy to configure failure detection times for highly customized or extremely quiet environments.

### Risks of Implementing:
Adds an additional field to the main configuration UI, which slightly increases complexity for novice users.

### Mitigation Steps:
Placed the setting in the primary "Gateway configuration" menu alongside `scan_interval` with a robust description and a safe, out-of-the-box default of 10 minutes. This ensures non-technical users can safely ignore it while power users get the control they need.

### AI Assistance Disclosure:


silverailscolo:
Good as is, thanks. Will release tonight

silverailscolo (approved):

PR #589, opened 2026-04-06
Fix: Poll device state on entity initialization to prevent Unknown states

PWhite-Eng:
### The Problem:

Following a Home Assistant restart or a system cache clear, entities (`system_mode` on the Controller, `mode` on Zones, and `active_fault` on the Logbook) frequently initialize as `null` or `Unknown`. This happens because the integration passively waits for broadcast packets from the controller rather than actively asking for its current state. (References Issue #583, points 1 and 2).

### Consequences:

If not fixed, this breaks HA startup automations that rely on `system_mode` or fault states. It also creates a confusing user experience where the dashboard shows `Unknown` entities for an extended period until a state change is natively triggered or broadcasted.

### The Fix:

Added active polling during the entity initialization lifecycle. If an entity's required state is missing upon creation, it will immediately request the data from the gateway.

### Technical Implementation:
- Overrode `async_added_to_hass` in `RamsesController`, `RamsesZone`, and `RamsesLogbookBinarySensor`.
- `RamsesController`: Evaluates `system_mode`. If `None`, issues `RQ {id} 2E04 FF`.
- `RamsesZone`: Evaluates `mode`. If `None`, issues `RQ {tcs.id} 2349 {zone.idx}`.
- `RamsesLogbookBinarySensor`: Evaluates `active_faults`. If `None`, calls the native `get_faultlog()` method, falling back to a raw `RQ {tcs.id} 0418 00` command if necessary.
- Wrapped all transmission logic in `try/except Exception` blocks so that if the gateway is overwhelmed during startup, the entity still successfully initializes (falling back to its passive listening behavior).

### Testing Performed:

Added targeted `async_added_to_hass` unit tests in `test_climate.py` and `test_binary_sensor.py`. The tests mock the gateway and verify that:
1. `RQ` commands are correctly formatted and dispatched when the state is `None`.
2. Polling is completely skipped if the cache already populated the state.
3. Network exceptions raised during polling do not crash the entity setup.
Mypy strict typing checks and the full pytest suite pass successfully.

Users who clear their cache or restart Home Assistant will continue to experience degraded stability, broken automations, and missing UI states.

### Risks of Implementing:

There will be a slight, localized burst of RF traffic (RQ packets) broadcast onto the network immediately as Home Assistant boots up and initializes the integration.

### Mitigation Steps:

Polling is strictly conditional; it only fires if the resolved attribute is currently `None` (meaning the cache failed or is empty). Catch-all exception handlers guarantee that any RF packet collisions or timeouts during this startup burst will not crash the entity loading process.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -39,6 +39,7 @@ from ramses_rf.entity_base import Entity as RamsesRFEntity
 from ramses_rf.gateway import Gateway
 from ramses_rf.schemas import SZ_BLOCK_LIST, SZ_CONFIG, SZ_KNOWN_LIST, SZ_SCHEMA
 from ramses_rf.system.heat import Logbook, System
+from ramses_tx.command import Command
 from ramses_tx.const import SZ_BYPASS_POSITION, SZ_IS_EVOFW3
 
 from .const import (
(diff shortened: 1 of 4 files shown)


silverailscolo (approved):
Unrelated to the _rf big bang PR, I assume. If I create _rf 0.56.1 from current master and let _cc point to that as a dependency, this could already work?

PWhite-Eng:
I was just looking at this regarding the CI test failures: 

The tests are currently failing because the GitHub Actions runner is pulling the `ramses_rf 0.56.0` release from PyPI, which does not yet include the `SZ_BUFFER_CAPACITY` constant introduced in the `feat/530-phase-1-schema-wiring` branch, which has been merged into `master`.

I have fully tested this PR on my local machine linked against `ramses_rf` `master`, and I can confirm that Mypy is clean and 100% of the pytest suite passes successfully. 

Do you want to merge this in before you do a version bump to 0.56.1? and then hopefully it will all pass?

PR #600, opened 2026-04-08
Fix #592: Add ConfigEntry migration to resolve 0.56.1 setup failures

### LINKED ISSUE #592: Loading 0.56.1 fails: resubmit Config > Packet-log to fix
As described at the top of the 0.56.1 Release Notes:

To continue loading the integration, open HA > Integrations > Ramses RF > Config, open Packet-Log tab, click Submit. No changes needed on the tab. Just a one time action. Sorry for the inconveniance.

PWhite-Eng:
### The Problem:

Users upgrading to version 0.56.1 experience a "Not loaded" state during setup. This occurs because legacy configuration options stored in `.storage/core.config_entries` (such as deprecated database flags and old `packet_log` keys) fail validation against the newly refactored, stricter `ramses_rf` voluptuous schema. (Resolves Issue #592).

### Consequences:

If not fixed, the integration will silently fail to initialize for existing users upon upgrading. Users are currently forced to manually open the Integrations UI, navigate to the config flow, and click "Submit" to force Home Assistant to rebuild and sanitize the stored configuration dictionary.

### The Fix:

Implemented Home Assistant's standard `async_migrate_entry` lifecycle hook to automatically sanitize the stored configuration data on startup, stripping out deprecated keys before `async_setup_entry` runs.

### Technical Implementation:
- Bumped the `ConfigFlow` `VERSION` from `1` to `2` in `config_flow.py`.
- Added `async_migrate_entry` in `__init__.py` to intercept v1 configurations.
- The migration creates a deep copy of the immutable `MappingProxyType`, uses `isinstance(..., dict)` checks to ensure safe unpacking, and pops legacy keys (`use_database`, `database_file`, `file_name`) using `.pop(key, None)`.
- The entry is then updated and bumped to version 2 via `hass.config_entries.async_update_entry`.

### Testing Performed:

Added two new tests to `tests/tests_new/test_init.py`:
1. `test_async_migrate_entry_v1_to_v2`: Verifies that legacy keys are correctly stripped and the version is bumped to 2.
2. `test_async_migrate_entry_v2_no_change`: Verifies that a version 2 entry is ignored and not mutated (idempotency).
- Confirmed safe handling of `NoneType` values in the `packet_log` dictionary to resolve earlier `TypeError` exceptions.
- Mypy passes with no errors. Full pytest suite passes with 100% success rate.

Existing users will continue to experience broken setups upon upgrading, leading to frustration, a broken smart home experience, and increased support tickets regarding Issue #592.

### Risks of Implementing:

If the migration logic mutates the dictionary incorrectly or encounters unexpected data types (e.g., throwing a `TypeError` during unpacking), it could permanently prevent a user's config entry from loading, requiring them to delete and reinstall the integration.

### Mitigation Steps:

Used safe `.pop(key, None)` methods and strict `isinstance(val, dict)` checks before attempting dictionary unpacking. Full unit test coverage was added specifically to cover the migration lifecycle and ensure no regressions occur.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -181,6 +181,49 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     return True
 
 
+async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
+    """Migrate legacy configuration options to the current version.
+
+    This handles the transition away from user-selectable database storage
+    and removes deprecated packet log keys (e.g., `file_name`) that cause
+    strict schema validation to fail during setup.
+
+    :param hass: The Home Assistant instance.
+    :param entry: The ConfigEntry to migrate.
+    :return: True if the migration succeeded.
+    """
+    _LOGGER.debug("Migrating ramses_cc config entry from version %s", entry.version)
+
+    if entry.version == 1:
+        # Create a deep copy of the immutable MappingProxyType to mutate it
+        new_options = {**entry.options}
+
+        # 1. Clean up packet_log dictionary
+        if isinstance(new_options.get("packet_log"), dict):
+            packet_log = {**new_options["packet_log"]}
+            # Remove deprecated key mentioned in issue #592
+            packet_log.pop("file_name", None)
+            new_options["packet_log"] = packet_log
+
+        # 2. Clean up ramses_rf dictionary (legacy database storage flags)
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
@silverailscolo, My friend did not have the HA log to see exactly which keys in configflow are the issue, but hopefully if there are any more it is easy enough to add to `def async_migrate_entry`
I do not know how this will handle downgrading once this is merged into master?
Phil

silverailscolo (approved):

PR #604, opened 2026-04-09
Relax config schema

silverailscolo:
- to allow extra keys (during migration from V1 to V2)
- log as deprecated using HA built in cv.deprecated("key") << not yet active
- add MINOR_VERSION (explicitly show current default value = 1)
- edit comments
- safe get(attr) replacing [attr]

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -182,7 +182,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
 
 async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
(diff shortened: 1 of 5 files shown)


silverailscolo:
@PWhite-Eng is this not contradicting your recent schema changes? I don't think it is harmful to receive a legacy key during the deprecation period.

PWhite-Eng:
Hi @silverailscolo,

You are correct that this relaxes the stricter validation I recently introduced, but I completely agree with your reasoning.

Temporarily allowing legacy keys during a deprecation period is definitely the right approach to ensure a smooth upgrade path for users, rather than abruptly breaking their setups. As long as we are throwing a warning (e.g., via `cv.deprecated` or a logger warning) so users know to they have a out of date config, I am fully in favour of this change.

Thanks for handling this
Phil

PR #610, opened 2026-04-11
refactor: replace brittle async checks and flatten config flow nesting

PWhite-Eng:
### The Problem:
1. **Issue 3 (Brittle Asynchronous Resolution):** In `helpers.py`, the `resolve_async_attr` function relied on string matching class names (`type(val).__name__ == "coroutine"`) to detect asynchronous objects.
2. **Issue 4 (Arrow Anti-Pattern):** In `config_flow.py`, the `async_step_zigbee_device` method suffered from deep nesting (multiple `if`, `for`, `try` blocks wrapped inside each other), making the logic difficult to trace and maintain.

### Consequences:
- **Brittle Async:** String matching is highly fragile. If Home Assistant or Python were to change how futures or tasks are named under the hood in future versions, the integration's property resolution would break silently.
- **Deep Nesting:** The Arrow Anti-Pattern severely increases cognitive load for future maintainers, increasing the likelihood of introducing bugs during UI flow modifications.

### The Fix:
- Refactored `helpers.py` to use Python's standard library `inspect` module for type detection.
- Refactored `config_flow.py` by extracting the IEEE identification logic into a standalone helper function and using early returns to flatten the main execution path.

### Technical Implementation:
- **`helpers.py`:** Replaced the string evaluation with `inspect.isawaitable(val)` and `isinstance(val, asyncio.Future)`.
- **`config_flow.py`:** Created a private `_extract_ieee_from_device(device_entry)` helper method. Implemented guard clauses (early returns) in `async_step_zigbee_device` to prevent logic from drifting rightward across the screen. Re-added a missing `initial_setup` parameter to the `BaseRamsesFlow` initialization to satisfy strict Mypy typing.

### Testing Performed:
- Verified strict `mypy` compliance.
- Ran the full `pytest` suite to confirm 100% pass rate. The existing config flow tests and property resolution tests continue to pass, proving the structural changes did not alter the public contract.

Leaving the code as is leaves behind accumulating technical debt. The brittle async checks specifically pose an upgrade risk for future Python or Home Assistant Core releases.

### Risks of Implementing:

Very low. The changes are structurally isolated and rely on highly stable standard library functions rather than custom string evaluations.

### Mitigation Steps:

Verified the internal UI state machine routing against the existing test suite to ensure the guard clauses did not inadvertently short-circuit necessary ConfigFlow steps.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/config_flow.py
+++ b/custom_components/ramses_cc/config_flow.py
@@ -110,6 +110,15 @@ async def async_get_usb_ports(hass: HomeAssistant) -> dict[str, str]:
     return await hass.async_add_executor_job(get_usb_ports)
 
 
+def _extract_ieee_from_device(device_entry: dr.DeviceEntry) -> str | None:
+    """Extract the IEEE address from a device registry entry."""
+    for _domain, ident in device_entry.identifiers:
+        ident_str = str(ident)
+        if re.fullmatch(r"[0-9A-Fa-f:]{8,}", ident_str):
+            return ident_str
+    return None
+
+
 class BaseRamsesFlow(FlowHandler):
     """Mixin for common Ramses flow steps and forms."""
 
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo, PR No. 2 of 3 for ramses_cc. Please note this PR is NOT stacked
Thanks
Phil

silverailscolo (approved):

PR #611, opened 2026-04-11
chore: standardize diagnostic logging to preserve stack traces in broad exception traps

PWhite-Eng:
### The Problem:

**Issue 5 (Catching Exceptions in Polling/Message Loops):** In `climate.py` and `mqtt_bridge.py`, there are necessary broad `except Exception as err:` blocks designed to prevent the Home Assistant event loop or MQTT listener from crashing when encountering unexpected data or database locks. However, many of the corresponding `_LOGGER.error` or `_LOGGER.warning` calls did not include the stack trace.

### Consequences:

When an unexpected exception occurs inside one of these polling or message loops, the integration survives, but the error is logged opaquely (e.g., `Failed to decode JSON: list index out of range`). Without the stack trace, it is nearly impossible for developers to track down the exact line of code or the root cause of the failure.

### The Fix:

Added `exc_info=True` to all diagnostic logger calls nested inside broad `Exception` traps across the climate platform and the MQTT bridge.

### Technical Implementation:
- **`climate.py`:** Updated the generic exception handlers in properties like `current_temperature` to explicitly pass `exc_info=True` to `_LOGGER.warning` and `_LOGGER.error`.
- **`mqtt_bridge.py`:** Standardized the logging output in `_handle_rx_message` and `_handle_cmd_message` so that `UnicodeEncodeError`, `JSONDecodeError`, and generic `Exception` traps preserve their stack traces. Converted instances of `_LOGGER.exception` to `_LOGGER.error(..., exc_info=True)` to maintain strict uniformity and adhere to line-length constraints.

### Testing Performed:
- Verified strict `mypy` compliance.
- Ran the full `pytest` suite to confirm a 100% pass rate, ensuring no logging mock assertions or logical execution paths were inadvertently broken.

Future bugs caused by upstream library changes or firmware updates will fail silently or opaquely, severely increasing the time required for triage and debugging.

### Risks of Implementing:

Negligible. In the event of a catastrophic failure loop, the Home Assistant log file could grow slightly faster due to the inclusion of full tracebacks.

### Mitigation Steps:

Only applied `exc_info=True` to explicitly caught, exceptional error states rather than standard operational info/debug logs.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -186,7 +186,10 @@ class RamsesController(RamsesEntity, ClimateEntity):
                 await self._device._gwy.async_send_cmd(cmd)
             except Exception as err:
                 _LOGGER.debug(
-                    "Failed to poll system_mode for %s: %s", self.entity_id, err
+                    "Failed to poll system_mode for %s: %s",
+                    self.entity_id,
+                    err,
+                    exc_info=True,
                 )
 
     @property
@@ -491,7 +494,12 @@ class RamsesZone(RamsesEntity, ClimateEntity):
                 )
                 await self._device._gwy.async_send_cmd(cmd)
             except Exception as err:
-                _LOGGER.debug("Failed to poll mode for %s: %s", self.entity_id, err)
+                _LOGGER.debug(
+                    "Failed to poll mode for %s: %s",
+                    self.entity_id,
+                    err,
+                    exc_info=True,
+                )
 
     @property
     def current_temperature(self) -> float | None:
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo, No. 3 of 3 PR's I have for ramses_cc, I will now move over to look at the issues in ramses_rf. 
This PR is NOT stacked, but it appears that this morning you are too quick for me for this to make a difference!

silverailscolo (approved):
Great to get to the source of problems.

PR #617, opened 2026-04-12
Fix/get fan param guard

wimpie70:
Fix: guard get_fan_param calls against DeviceHvac without the method

## Problem
On integration startup, when a device is in the known_list with class: FAN but hasn't yet been fingerprinted into the FAN subclass, ramses_cc creates fan parameter number entities for it. The `_request_parameter_value` method immediately calls `self._device.get_fan_param(param_id)` on what is still a generic `DeviceHvac` object — which does not have that method — causing:

```
AttributeError: 'DeviceHvac' object has no attribute 'get_fan_param'
Error doing job: Task exception was never retrieved
```
This fired multiple times per startup (once per parameter entity) and also blocked with 20 s × 3-retry timeouts on RQ 2411 before the device was properly resolved.

## Fix
Added hasattr(self._device, "get_fan_param") guards at both call sites in _request_parameter_value:

First call (store read) — returns early with a debug log if the method is absent
Second call (RQ dispatch) — skipped silently if the method is absent
No functional change for fully-fingerprinted FAN devices.

## Tests
Added to tests/tests_new/test_number.py:

test_request_parameter_value_no_get_fan_param — verifies the method returns early without raising or scheduling a pending task when the device is a generic DeviceHvac (no get_fan_param)
test_request_parameter_value_no_get_fan_param_second_call — verifies that for a proper FAN device both call sites are reached (call_count == 2)
mock_hvac_device fixture — a MagicMock with only ["id", "_SLUG", "supports_2411"] in its spec, simulating pre-fingerprint DeviceHvac

- [x] AI Coding Assistance disclosure: not used

### DIFF
--- a/custom_components/ramses_cc/number.py
+++ b/custom_components/ramses_cc/number.py
@@ -733,6 +733,13 @@ class RamsesNumberParam(RamsesNumberBase):
             return
 
         # This just checks the store, doesn't send RQ
+        if not hasattr(self._device, "get_fan_param"):
+            _LOGGER.debug(
+                "Device %s (%s) has no get_fan_param, skipping",
+                self._device.id,
+                type(self._device).__name__,
+            )
+            return
         value = self._device.get_fan_param(param_id)
 
         _LOGGER.debug(
@@ -760,7 +767,8 @@ class RamsesNumberParam(RamsesNumberBase):
 
         self.set_pending()
 
-        self._device.get_fan_param(param_id)
+        if hasattr(self._device, "get_fan_param"):
+            self._device.get_fan_param(param_id)
 
         self.hass.async_create_task(self._clear_pending_after_timeout(30))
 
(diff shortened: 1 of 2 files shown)


wimpie70:
I also had a look on how this works with the supports_2411 check.

So supports_2411 is checked at entity creation time (create_parameter_entities and the device filter at line 241-244). If the device has supports_2411 = True (which a DeviceHvac with class: FAN in known_list can have), entities are created — but the device object may still be generic DeviceHvac without get_fan_param.

The sequence is:

1. Known list has FAN → ramses_cc creates a DeviceHvac (generic) with supports_2411 = True (set by class)
2. create_parameter_entities sees supports_2411=True → creates entities ✓
3. _request_parameter_value is called immediately → DeviceHvac has no get_fan_param → crash

The guard is correct and necessary. The supports_2411 check does not protect against this — it only gates entity creation, not the method call on the generic object. Our hasattr guard is the right fix in the right place.

The existing test test_request_parameter_value_no_get_fan_param already validates this correctly since mock_hvac_device has supports_2411 = True but no get_fan_param — exactly matching the real failure scenario.

silverailscolo (approved):
Nice, thanks

PR #622, opened 2026-04-13
List serial ports with the `usb.scan_serial_ports` helper

puddly:
Home Assistant Core is moving to using the USB integration directly to list serial ports. This is one of the few custom components that does it completely manually but still uses helpers from the USB integration so I'm opening this PR to align it with how Core will function next release.

CI will fail for this PR until the next Core release/beta is made in a few weeks.

- [x] AI Coding Assistance disclosure: not used

**PR Summary**

### DIFF
--- a/custom_components/ramses_cc/config_flow.py
+++ b/custom_components/ramses_cc/config_flow.py
@@ -26,7 +26,6 @@ from homeassistant.helpers import (
     selector,
 )
 from homeassistant.helpers.storage import Store
-from serial.tools import list_ports  # type: ignore[import-untyped]
 
 from ramses_rf.schemas import (
     SCH_GATEWAY_DICT,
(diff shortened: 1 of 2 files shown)


silverailscolo:
Thanks for your PR.
From your description I learn that we'll have to wait for the next? HA Core release to pass this Test currently failing:
`E   ImportError: cannot import name 'SerialDevice' from 'homeassistant.components.usb' (/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/homeassistant/components/usb/__init__.py) `

puddly:
Indeed. `usb.usb_device_from_port` will be removed in the next Home Assistant Core release in a few weeks. I apologize for the breaking change.

One way to work around this would be to release a version of `ramses_cc` before the next Home Assistant release that implements `get_usb_ports` with feature detection:
```python
if not hasattr(usb, "async_scan_serial_ports"):
    def get_usb_ports() -> dict[str, str]:
        # Old function compatible with earlier HA Core versions
else:
    def get_usb_ports() -> dict[str, str]:
        # New function compatible with HA Core 2026.6.0+
```

If you think that approach is fine, I can add it to this PR.

silverailscolo:
Adding that seems good for  UX, would migrate unnoticed. Including the new version is good as a tag (and a TODO: remove Q3 2026)

puddly:
Done, let me know if this works. I've adjusted unit tests to work with both versions as well.

---

Unrelated to this change, the motivation for HA Core centralizing serial port listing is because we are swapping our serial backend from pyserial to [serialx](https://github.com/puddly/serialx).

I've noticed that this library and the ramses-rf package use pyserial extensively and both have many workarounds for outstanding pyserial bugs. Check out `serialx` and consider migrating internally as well within your library. Serialx has native async support but does not support the construction method used by ramses-rf (passing the synchronous `Serial` object [to a serial transport](https://github.com/ramses-rf/ramses_rf/blob/9453567d0f332afa113a798ea8da4de194cba26a/src/ramses_tx/transport/port.py#L157)) so this part would need a little bit of rewriting to use a proper `asyncio.Protocol` instead of the raw transport. You can see https://puddly.github.io/serialx/usage.html#async-transport for a bit more documentation.

silverailscolo (approved):

silverailscolo:
Tested to operate and config (still using pyserial on Core 2026.4.2) OK on test rig with original HGI80, ramses_esp, evofw3 over USB.

PR #628, opened 2026-04-18
Refactor: Transition to Modern DTO-based Gateway Configuration

PWhite-Eng:
### The Problem:

The `ramses_cc` integration was utilizing a legacy flat `**kwargs` unpacking pattern to initialize the `ramses_rf.Gateway`. This approach broke encapsulation, bypassed strict type checking, and relied on internal backward-compatibility shims to resolve nested Home Assistant configuration dictionaries. Additionally, legacy mock classes in the test suite prevented the use of modern Python introspection, and legacy configuration keys threatened to pollute the runtime coordinator logic.

### Consequences:

Failure to modernize this integration would lead to:
- **Configuration Instability**: Implicit reliance on `**kwargs` and library-level deprecation shims makes the integration prone to silent failures if those shims are removed.
- **Brittle Tests**: Testing against fake mock classes rather than the true DTOs masked underlying typing issues and structural changes in the core library.
- **Architectural Debt**: Forcing the runtime coordinator to translate legacy Home Assistant configuration keys violates separation of concerns.

### The Fix:

Implemented a high-level modernization of the `RamsesCoordinator` to natively construct and pass `GatewayConfig` and `EngineConfig` DTOs. This was coupled with moving all legacy configuration translations to the HA migration lifecycle (`__init__.py`), keeping the runtime code strictly typed and clean. Finally, the test suite was upgraded to validate against the true nested DTO architecture.

### Technical Implementation:
- **DTO Composition**: Refactored `_create_client` in `coordinator.py` to utilize `dataclasses.fields()` for dynamic, type-safe filtering of HA configuration into isolated `EngineConfig` and `GatewayConfig` containers.
- **Configuration Migration**: Leveraged `async_migrate_entry` in `__init__.py` to permanently clean up and translate legacy configuration keys (like `file_name` and `rotate_backups`) on the user's disk upon version bump, protecting the coordinator from malformed data.
- **Test Suite Modernization**: Stripped legacy `FakeGatewayConfig` mock classes from `test_coordinator.py` to allow native `dataclasses.fields` introspection to succeed, and updated all assertions to correctly target nested properties (e.g., `kwargs["config"].engine.known_list`).

### Testing Performed:
- **Test Suite Pass**: Executed the `pytest` suite for `test_coordinator.py`, successfully clearing all `TypeError`s and achieving a 100% pass rate for the initialization logic.
- **Type Safety**: Ensured strict compatibility with the new nested DTO structures, passing Ruff and Mypy standards.

[...]

PR #629, opened 2026-04-18
Fix: Ad-hoc concurrency lock to prevent event loop saturation

PWhite-Eng:
### The Problem:

Home Assistant's event loop can be overwhelmed by concurrent state-change events triggered by rapid-fire RF packet bursts from `ramses_rf` (Issue ramses-rf/ramses_rf#608). When multiple updates for the same device arrive within milliseconds, multiple concurrent `async_write_ha_state` tasks are scheduled, bypassing any structural limits.

### Consequences:

Without a safeguard, users experience severe CPU ramping and event loop saturation as Home Assistant struggles to process overlapping UI state writes. This degrades the performance of the entire Home Assistant instance and forces users to downgrade.

### The Fix:

Implemented a Modbus-integration style `asyncio.Lock()` per entity to ensure `async_write_ha_state` cannot be executed concurrently for the same device. Also introduced a throttled logger to track suppressed duplicate updates.

### Technical Implementation:
- **Concurrency Locks:** Introduced `self._update_lock = asyncio.Lock()` inside `RamsesEntity.__init__`.
- **Dispatcher Wrapper:** Replaced the direct HA signal callback with `_async_update_and_write_state`.
- **Skip & Absorb:** If the lock is already held, subsequent rapid requests return immediately. Because Home Assistant pulls the most recent data from the device memory when `async_write_ha_state` finally executes, it naturally absorbs the intermediate changes.
- **Throttled Logging:** Added a `time.monotonic()` based tracker that safely prints a `WARNING` to the logs (maximum once per minute) summarizing how many redundant updates were caught by the lock.
- **Test Alignment:** Updated the assertion in `test_entity.py` to target the new wrapper function.

### Testing Performed:
- Passed `mypy` (strict mode).
- Passed the full `pytest` suite.
- Specifically verified `test_async_added_to_hass` in `test_entity.py` to ensure the dispatcher properly binds to the new locked wrapper method.

Leaving the code as-is leaves the integration highly vulnerable to catastrophic HA event loop saturation during RF floods, rendering the host system unresponsive.

### Risks of Implementing:

If the lock is currently held and the final packet in a rapid burst is dropped, the HA UI might theoretically miss that final state transition until the next periodic update or RF packet arrives.

### Mitigation Steps:
- Added the throttled warning logger so the maintainers have full visibility into exactly how many updates are being dropped and when.
- Acknowledged that full Priority 5 implementation (strict state diffing and debouncing—waiting until the burst ends before writing) will eventually replace this to completely solve the end-of-burst delay.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -203,6 +203,12 @@ async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             packet_log = {**new_options["packet_log"]}
             # Remove deprecated key mentioned in issue #592
             packet_log.pop("file_name", None)
+            # Translate legacy rotate_backups to modern key
+            if "rotate_backups" in packet_log:
+                packet_log["packet_log_retention_days"] = packet_log.pop(
+                    "rotate_backups"
+                )
+
             new_options["packet_log"] = packet_log
 
         # 2. Clean up ramses_rf dictionary (legacy database storage flags)
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import dataclasses
 import inspect
 import logging
 import re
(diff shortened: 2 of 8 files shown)


silverailscolo (approved):

PWhite-Eng:
@silverailscolo, regarding the failing tests. I have tested it on my local setup which is configured to point to ramses_rf master, where it passes all tests. As this PR/Branch is a stacked PR ontop of #628, which needs a version bump, this is causing this PR to fail.

silverailscolo:
Stacked on top of PR #628

silverailscolo:
@PWhite-Eng Thanks for the fix. Master is passing all checks.

PR #635, opened 2026-04-23
Fix duplicate unique_id errors in number platform

wimpie70:
I still had duplicate errors on params. This should fix it.

Fix duplicate unique_id errors and circular import
Changes
Fix duplicate unique_id errors in number platform
Added coordinator tracking for parameter entities with three sets:
_parameter_entities_pending: entities scheduled for addition
_parameter_entities_loaded: entities successfully added to platform
_parameter_entities_created: entities instantiated in create_parameter_entities

Clear all tracking sets during platform setup to ensure clean state on reload
Check pending set before adding entities to prevent duplicate scheduling
Track created entities at source in create_parameter_entities to skip duplicate instantiations
Log with actual unique_id for easier debugging of registry migrations

Fix circular import with ramses_tx
Removed top-level from ramses_tx import exceptions as exc which triggered circular import at module load
Added lazy import helper _get_ramses_tx_exceptions() that imports ramses_tx.exceptions only when needed
Use the helper in async_setup_entry to catch TransportSourceInvalid and TransportError exceptions
This allows Home Assistant to load the integration without triggering the ramses_tx ↔ ramses_rf circular dependency

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import logging
 import os
 import sys
+from types import ModuleType
 
 # from collections.abc import Callable
 #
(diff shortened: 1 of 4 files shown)


silverailscolo:
Thanks for the PR @wimpie70 
The circular import error in ramses_rf that you patch here using lazy import is not the same as [the one that has just been fixed ](https://github.com/ramses-rf/ramses_rf/pull/628)where it happened in master?

silverailscolo (approved):

PR #654, opened 2026-05-05
Fix MQTT transport persistence regression and heal missing serial_port without config-entry version bump

### LINKED ISSUE #655: Blank Config Schema
Current ramses_cc master in Config the System Schema shows as blank, and the gateway status sensor is always **Problem** (but it contains the schema under Details).

Integration operates normally (except other open bugs).

wimpie70:
Fixes #655 

Note: my first idea was to bump Config version to v3, but reverted this.

## Summary
This PR fixes a regression where RAMSES RF transport settings could be lost from options storage (ending up as `serial_port: {}`), which can cause startup failures with:

```
TypeError: Either a port_name or an input_file must be specified
```

The fix is implemented in a rollback-safe way (no config-entry major version bump).

## Root Cause
In options flow, transport-related options could be saved from incomplete state in some paths.

Existing entries that already had missing/empty serial_port.port_name were not automatically healed.

Unrelated options saves (e.g. schema/logging toggles) could then persist the broken transport block.

### 1) Options flow persistence fix

Explicitly bind the active config_entry when creating the options flow handler.
This ensures `get_options()` always starts from persisted entry options (including serial_port), preventing accidental overwrites with empty data.

Files:

`custom_components/ramses_cc/config_flow.py`

### 2) Rollback-safe setup-time healing (no major version bump)

Added setup-time healing for entries with missing serial_port.port_name.
If MQTT usage is implied (mqtt_use_ha, mqtt_hgi_id, mqtt_topic, or existing HA MQTT entry), the integration now heals options to:
```
serial_port.port_name = "mqtt_ha"
mqtt_use_ha = True (if absent)
```
Healing is persisted via `async_update_entry` during setup.

Files:

`custom_components/ramses_cc/__init__.py`

### 3) Keep config-entry VERSION = 2

Intentionally did not bump major config-entry version.

Preserves downgrade compatibility with builds expecting version 2.

Existing v1->v2 migration remains unchanged for legacy cleanup.

Files:

- `custom_components/ramses_cc/config_flow.py`
- `custom_components/ramses_cc/__init__.py`

### 4) Coordinator fallback

Kept existing coordinator fallback as an additional safety net for already-corrupted entries and unexpected edge cases.

## Tests
Added/updated tests covering:

- options flow schema save preserves existing serial_port
- migration behavior remains v1->v2
- v2 entries are not migrated
- setup-time serial-port healing helper behavior (heal + no-heal cases)

Files:

- `tests/tests_new/test_config_flow.py`
- `tests/tests_new/test_init.py`

## Validation

Targeted tests passed locally:

`pytest tests/tests_new/test_init.py -k 'migrate_entry or healed_serial_port_options'`

`pytest tests/tests_new/test_config_flow.py -k 'options_flow_schema_save_preserves_serial_port'`

## Impact

- Prevents future accidental loss of transport settings in options flow.
- Auto-recovers affected MQTT-based entries during setup.
- Maintains rollback safety by avoiding a new config-entry major version.

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -14,6 +14,7 @@ import logging
 import os
 import sys
 from types import ModuleType
+from typing import Any
 
 # from collections.abc import Callable
 #
@@ -165,6 +166,18 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         _LOGGER.debug("Entry %s is already set up", entry.entry_id)
         return True
 
+    healed_options = _healed_serial_port_options(
+        {**entry.options},
+        mqtt_entries_present=bool(hass.config_entries.async_entries("mqtt")),
+    )
+    if healed_options is not None:
+        hass.config_entries.async_update_entry(entry, options=healed_options)
+        _LOGGER.warning(
+            "Healed missing serial_port for entry %s by defaulting to mqtt_ha. "
+            "Please verify transport settings in the options flow.",
+            entry.entry_id,
+        )
+
     coordinator = RamsesCoordinator(hass, entry)
 
     try:
@@ -208,7 +221,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
 
 async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
-    """Migrate legacy configuration options to the current version 3.
+    """Migrate legacy configuration options to the current version 2.
 
     This handles the transition away from user-selectable database storage
     and removes deprecated packet log keys (e.g., `file_name`) that cause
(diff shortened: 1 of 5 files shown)


wimpie70:
I invested a bit on the question on how to downgrade to an earlier version. It might be obsolete, but it may also help so here's that little conversation:

- i'm looking for a mechanism to create in-between versions that will handle downgrading (add missing entries, other breaking fixes) so users can downgrade after that even further
---
Yes — this is exactly the right way to think about it.

For HA config entries, there is no native reverse migration hook, so the safest pattern is:

Recommended mechanism: “Bridge releases” (no major bump)
Keep ConfigFlow.VERSION stable (e.g. stay on 2).
Add an internal options schema marker, e.g. options["_schema_rev"] = 3.
In async_setup_entry(), run a normalizer/healer that:
adds missing keys,
repairs invalid combos,
keeps unknown/legacy keys,
writes back only backward-safe shapes.
In options flow save paths, do dual-write when needed:
write new key(s) + legacy key(s) for a while.
After 1–2 bridge releases, older versions can still load because entry VERSION never changed and keys remain compatible.
This gives you downgrade resilience across multiple versions.

If you must introduce v3 later
Then downgrades to v2/older are not guaranteed anymore. To mitigate:

Ship a pre-v3 bridge release first (as above), so most entries are already “safe”.
Add a backup of old options before v3 transform (e.g. options["_backup_v2"] or separate storage).
Provide a small documented/manual rollback step (script/service) to restore v2-compatible shape.
But practically, if downgrade support is important, prefer no major bump + runtime normalization.

silverailscolo (approved):
@wimpie70  Nice improvement, thanks. Will test changed files locally tonight > Done, perfect. Fixes issue #655

PR #675, opened 2026-05-12
Bump version, dep, date

silverailscolo:
... to 0.56.8 pre-release


silverailscolo:
@PWhite-Eng _rf 0.56.8 is on PyPi, so you can point to it during development. The Testing check on this PR requires adjustments on import paths from _rf.

PR #676, opened 2026-05-13
chore: Update ramses_rf import paths to support new package structure

PWhite-Eng:
### The Problem: 
The upstream `ramses_rf` library recently underwent a significant structural refactor, which included renaming and moving several core modules. Because of this, `ramses_cc` threw fatal `ModuleNotFoundError` exceptions during initialization and test collection (specifically failing to find `ramses_rf.entity_base` and `ramses_rf.message`).

### Consequences: 
If left unpatched, `ramses_cc` is fundamentally incompatible with the latest version of `ramses_rf`. The integration will silently fail to initialize, crash the setup process, and entirely break functionality for users updating their environments.

### The Fix: 
Updated all internal import paths within the `ramses_cc` codebase to accurately reflect the newly consolidated and renamed modules in `ramses_rf`.

### Technical Implementation:
- Executed a global update mapping `ramses_rf.entity_base` to the new root-level `ramses_rf.entity` module.
- Executed a global update mapping the singular `ramses_rf.message` file to the new plural `ramses_rf.messages` directory package.

### Testing Performed:
- Executed the complete `pytest` suite, transitioning from widespread collection failures to a 100% pass rate.
- Ran `mypy` across the codebase to ensure static typing and interfaces remained intact with the new paths. Zero errors reported.

The `ramses_cc` integration will remain hard-broken against the current `ramses_rf` branch. This will halt ongoing development and ultimately break the integration for end-users when the upstream changes are formally released.

### Risks of Implementing:
The risk is very low, as these are primarily structural path updates rather than logical rewrites. However, there is a minor risk that if `ramses_rf` altered underlying logic alongside the structural changes, some edge cases might behave differently.

### Mitigation Steps:
Fully verified the changes against the existing `pytest` suite to confirm behavioral expectations. Strict type checking via `mypy` was also used to ensure that the contracts between `ramses_cc` and the imported `ramses_rf` classes remain valid.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -21,7 +21,7 @@ from homeassistant.helpers.entity_platform import AddEntitiesCallback
 
 from ramses_rf.device.base import BatteryState, HgiGateway
 from ramses_rf.device.heat import BdrSwitch, OtbGateway, TrvActuator
-from ramses_rf.entity_base import Entity as RamsesRFEntity
+from ramses_rf.entity import Entity as RamsesRFEntity
 from ramses_rf.gateway import Gateway
 from ramses_rf.schemas import SZ_CONFIG, SZ_SCHEMA
 from ramses_rf.system.heat import Logbook, System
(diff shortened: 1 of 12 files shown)


PWhite-Eng:
@silverailscolo, all the tests pass on my local machine, when pointing to the latest master of ramses_rf. I assume the CI tests have failed as they are pointing to the latest release?

silverailscolo (approved):

Issue #698, opened 2026-06-02
MQTT TX payload includes trailing CRLF (`\r\n`) causing malformed HVAC messages


SilvioMessi:
## Describe the bug  
When using `ramses_cc` with Home Assistant MQTT, messages published on the `tx` topic contain a trailing CRLF sequence (`\r\n`) inside the `msg` payload.

This results in malformed packet strings being sent to the [ramses_esp dongle](https://github.com/IndaloTech/ramses_esp), and the messages are not correctly interpreted by my HVAC system ([RDCD25SK](https://www.broferpura.it/en/products/Domestic/rdcd25sk.html)).

## To Reproduce  
Steps to reproduce the behaviour:

1. Configure `ramses_cc` with Home Assistant MQTT
2. Connect a MQTT inspector tool (e.g. MQTT Explorer) to the broker  
3. Monitor the topic:  
   `RAMSES/GATEWAY/18:<gateway_id>/tx`  
4. Observe messages such as:

```json
{"msg": "RQ --- 32:124786 32:156438 --:------ 2411 003 000044\r\n"}
````

## Expected behaviour

Messages published on the `tx` topic should not include transport-level line terminators.
The payload should not contain the CRLF sequence (`\r\n`) inside the `msg` field.

Expected example:

```json
{"msg": "RQ --- 32:124786 32:156438 --:------ 2411 003 000044"}
```

## Root cause

I identified and fixed the root cause here:
[https://github.com/SilvioMessi/ramses_cc/commit/1b17e306365f8b459cb8dea17b0ec8200f257d51](https://github.com/SilvioMessi/ramses_cc/commit/1b17e306365f8b459cb8dea17b0ec8200f257d51)

I also tested the patched version of the integration.

Without this fix, messages sent by the integration are ignored or not correctly processed by the HVAC system. After applying the fix, commands and queries are correctly executed by the HVAC.

I am not fully sure whether this is specific to my setup, but the presence of CRLF in MQTT payloads appears unnecessary.

## Additional context

* Home Assistant OS
  * Core: 2026.5.4
  * Supervisor: 2026.05.1
  * Operating System: 17.3
  * Frontend: 20260429.4
* `ramses_cc`: 0.56.8
* `ramses_esp`: 0.4.9

silverailscolo:
Thanks for reporting, points to `Matt_bridge.py` line 83.
I will strip the added `/r/l` and see what others report.

silverailscolo:
Included in 0.57.0

PR #700, opened 2026-06-04
Strip HA MQTT payload

silverailscolo:
This PR fixes issue #698

**PR Summary**
- Do not add newline as txt to mqtt payload


silverailscolo:
@PWhite-Eng Apart from the test failures that also pop up in master, please confirm this change (from issue 698) is in line with the `mqtt_bridge` intents. Will hold until confirmed, thanks

Issue #702, opened 2026-06-05
Hopper D375: Unsupported packet


wlcrs:
**Device:** Hopper D375 by Hopper Ventilation (https://hopper-ventilation.com/oplossingen/wtw-hopper-375-525)
**Ramses version:** 0.56.9

An unknown message was logged after doing a power cycle of the device.

**Log messages:**

```
2026-06-05 17:57:10.345 INFO (MainThread) [ramses_tx.protocol.base] Recv'd: 070  I --- 32:137527 --:------ 32:137527 042F 009 00003200320032EB30
2026-06-05 17:57:10.348 INFO (MainThread) [ramses_rf.dispatcher] ||  32:137527 |            |  I | unknown_042F     |  00  || {'counter_1': '0x0032', 'counter_3': '0x0032', 'counter_5': '0x0032', 'unknown_7': '0xEB30'}
2026-06-05 17:57:10.939 INFO (MainThread) [ramses_tx.protocol.base] Recv'd: 068  I --- 32:137527 63:262142 --:------ 10E0 038 000001C84F0E0A6AFEFFFFFFFFFF0B0C07E1564D442D30325250533534000000000000000000
2026-06-05 17:57:10.941 INFO (MainThread) [ramses_rf.parsers.system] <ramses_rf.parsers.decoder._MockMessage object at 0x7fed39733150> < Support the development of ramses_rf by reporting this packet, with the make/model of device: <ramses_rf.parsers.decoder._MockAddress object at 0x7fed396ace30> (device type 32 not known to have signature: 0001C84F0E0A6AFEFF)
2026-06-05 17:57:10.942 INFO (MainThread) [ramses_rf.parsers.system] <ramses_rf.parsers.decoder._MockMessage object at 0x7fed39733150> < Support the development of ramses_rf by reporting this packet, with the make/model of device: <ramses_rf.parsers.decoder._MockAddress object at 0x7fed396ace30> (device type 32 not known to have signature: 0001C84F0E0A6AFEFF)
2026-06-05 17:57:10.943 INFO (MainThread) [ramses_rf.parsers.system] <ramses_rf.parsers.decoder._MockMessage object at 0x7fed39731710> < Support the development of ramses_rf by reporting this packet, with the make/model of device: <ramses_rf.parsers.decoder._MockAddress object at 0x7fed396ace30> (device type 32 not known to have signature: 0001C84F0E0A6AFEFF)
[...]

Issue #701, opened 2026-06-04
0.56.9: Failing to load/tests w/known_list


silverailscolo:
Pre-release 0.57.9 will not load devices. See error reported [here](https://github.com/ramses-rf/ramses_cc/pull/693).

The following three _init_  tests fail repeatedly, despite the integration working without errors on HVAC:

- tests/tests_new/test_init.py#L119
`
[...]

PR #704, opened 2026-06-05
Fix `known_list` trait propagation in `coordinator.py`

wlcrs:
This PR fixes (the second part of) issue #702 (and part of #701 ?).

- [x] AI Coding Assistance disclosure: used for 100.0%

**PR Summary**

**Problem**: HVAC entities were not created even with known_list class hints (FAN/REM).
**Root cause**: known_list traits were passed only to EngineConfig, while device trait lookup uses GatewayConfig.known_list.
**Fix**: also pass sanitized known_list into GatewayConfig in [coordinator.py](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/8761a5560c/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
**Result**: explicit class/bound/scheme traits are applied during device instantiation, enabling expected HVAC entity creation.

**Full story**: after investigating the code base with Copilot, I updated my known_lists to:

```yaml
"32:125274":
  class: REM
  scheme: orcon
"32:137527":
  class: FAN
  scheme: orcon
```

however, I still saw no entities being created in my HA. After further investigation by the Copilot, it proposed the fix in this PR, after which the entities appeared.

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -338,6 +338,9 @@ class RamsesCoordinator(DataUpdateCoordinator):
             for device_id, traits in raw_known_list.items()
         }
         engine_kwargs["known_list"] = sanitized_known_list
+        # Device traits (class/alias/faked/bound/scheme) are consumed by
+        # ramses_rf DeviceRegistry via GatewayConfig.known_list.
+        gateway_kwargs["known_list"] = sanitized_known_list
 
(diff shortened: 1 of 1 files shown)


silverailscolo:
Thanks for your PR. You fixed the empty `known_list` for now.
I will try to remove one of the duplicates in Config later.

silverailscolo (approved):

PR #707, opened 2026-06-07
Fix VirtualRF test race conditions and update coordinator configuration schema

PWhite-Eng:
### The Problem:

The test suite currently suffers from intermittent race conditions in the VirtualRF module due to the use of arbitrary, hardcoded sleep delays when waiting for hardware buffers to populate. Additionally, `test_coordinator.py` fails when attempting to assert against the `known_list` attribute due to an outdated object traversal path on the gateway config.

### Consequences:

If left unfixed, the CI pipeline will continue to fail unpredictably (flaky tests), causing developer friction and false negatives. Furthermore, outdated configuration schema tests will incorrectly block valid architectural updates to the main application.

### The Fix:

Implemented a dynamic asynchronous polling helper for VirtualRF serial reads, and updated the configuration attribute paths in the coordinator tests to match the current gateway architecture.

### Technical Implementation:
- **VirtualRF**: Introduced `async_wait_for_serial` in `test_virtual_rf.py`. This function continuously polls `ser.in_waiting` within a tight loop until data arrives, rather than relying on a blind `await asyncio.sleep(0.01)`. It includes a built-in timeout to prevent infinite hangs.
- **Coordinator**: Updated `test_coordinator.py` to access `kwargs["config"].known_list` directly, removing the deprecated intermediate `.engine` traversal.

### Testing Performed:

Ran the `pytest` suite locally against the targeted files (`tests/tests_new/test_virtual_rf/test_virtual_rf.py` and `tests/tests_new/test_coordinator.py`). Confirmed that the VirtualRF packet routing now synchronises correctly without race conditions, and that the coordinator mocks correctly validate the new configuration schema.

Flaky tests will continue to undermine developer confidence in the CI/CD pipeline, and broken mock paths will prevent accurate testing of the gateway's known device lists.

### Risks of Implementing:

The new polling loop in the VirtualRF tests could theoretically mask performance regressions in serial hardware emulation if the timeout window is too generous.

### Mitigation Steps:

A strict default timeout of `0.5` seconds was implemented within `async_wait_for_serial`. This ensures that while we accommodate normal asynchronous event loop variances, any genuinely stalled or broken processes will still correctly raise a `TimeoutError` and fail the test.

### AI Assistance Disclosure:

### DIFF
--- a/tests/tests_new/test_coordinator.py
+++ b/tests/tests_new/test_coordinator.py
@@ -377,10 +377,11 @@ async def test_create_client_strips_commands_from_known_list(
         mock_coordinator._create_client({})
 
         _, kwargs = cast(Any, mock_gwy).call_args
-        known_list = kwargs["config"].engine.known_list
 
-        assert known_list["37:168270"]["class"] == "REM"
-        assert CONF_COMMANDS not in known_list["37:168270"]
+        gwy_config = kwargs["config"]
+
+        assert gwy_config.known_list["37:168270"]["class"] == "REM"
+        assert CONF_COMMANDS not in gwy_config.known_list["37:168270"]
 
 
 async def test_async_update_discovery(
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo, this fixes a couple of the failing tests, but there are still more tests that fail. just trying to keep the PR's small

silverailscolo (approved):

PR #708, opened 2026-06-07
Fix: Update Gateway binary sensor known_list resolution for ramses_rf Phase 2.77 compatibility

PWhite-Eng:
### The Problem:

Following Phase 2.77 of the OSI Decoupling refactor in `ramses_rf`, the legacy `gwy.known_list` dictionary attribute was deprecated. L7 domain traits are now strictly housed within a `GatewayConfig` facade. Because the Home Assistant Gateway binary sensor was still attempting to parse the missing legacy data for its UI diagnostic attributes, an `IndexError` was triggered, causing immediate test failures (`test_gateway_binary_sensor_attrs`).

### Consequences:

If left unfixed, the integration will throw exceptions during entity state generation. Users will lose diagnostic visibility of the `known_list` and `block_list` within the Home Assistant frontend, and the test suite will remain broken. Furthermore, failing to strictly manage the configuration passed to the backend risks schema validation crashes during startup.

### The Fix:

Updated the Gateway binary sensor to point to the new OSI-compliant configuration facade. Additionally, introduced a robust end-to-end test suite to validate the lifecycle and sanitisation of the `known_list` payload prior to backend ingestion.

### Technical Implementation:
- Modified `RamsesGatewayBinarySensor.extra_state_attributes` to utilise a targeted `getattr` sequence (`getattr(gwy, "config", getattr(gwy, "_gwy_config", None))`) to safely extract the `known_list`, bypassing the deprecated property entirely.
- Refactored `test_binary_sensor.py` using the Arrange-Act-Assert (AAA) pattern to mock the modern `gwy.config.known_list` structure instead of relying on legacy duck-typing.
- Created a new, isolated test file (`test_known_list.py`) utilising the native `hass` fixture. This validates that `RamsesCoordinator._create_client()` correctly strips HA-specific keys (like `commands`) from the dictionary, and accurately injects the required virtual HGI (`18:xxx`) with its mandatory classes when operating over the MQTT transport layer.

### Testing Performed:
- **`test_binary_sensor.py`**: Validated that the Gateway diagnostic attributes format correctly using the new `gwy.config.known_list` data shape without triggering an `IndexError`.
- **`test_known_list.py`**:
    - `test_known_list_sanitised_for_serial`: Proved that custom Home Assistant parameters (e.g., `commands`) are successfully stripped to prevent `ramses_rf` schema validation failures.
    - `test_known_list_mqtt_hgi_injection`: Proved that the coordinator dynamically and correctly injects the virtual Gateway (`class: HGI`, `alias: ramses_esp`) into the data structure when the MQTT bridge is active.

The Gateway sensor will continuously throw indexing exceptions, breaking diagnostic views. Additionally, without strict sanitisation tests, future `ramses_rf` updates that tighten initialisation schemas could result in hard crashes if Home Assistant-specific dictionary keys leak into the backend.

### Risks of Implementing:

Users operating on heavily outdated firmwares or unsupported forks of `ramses_rf` that lack the `config` facade architecture may lose `known_list` visibility if the new attribute path cannot be resolved.

### Mitigation Steps:

The `getattr` extraction chain retains deep fallback support (checking `_gwy_config`, `engine._include`, and `gwy._include`) to gracefully handle edge cases and legacy environments without crashing the sensor. Strict static casting and Mypy compliance were applied to the new tests to prevent future type regressions.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -236,7 +236,7 @@ class RamsesGatewayBinarySensor(RamsesBinarySensor):
         """
         gwy: Gateway = self._device._gwy
         engine = getattr(gwy, "_engine", None)
-        gwy_config = getattr(gwy, "_gwy_config", None)
+        gwy_config = getattr(gwy, "config", getattr(gwy, "_gwy_config", None))
 
         # TODO Q3 2026: return await gwy._config() (only) instead of all below code
         # not yet working: self._cached_attrs = await gwy._config()
--- a/tests/tests_new/test_binary_sensor.py
+++ b/tests/tests_new/test_binary_sensor.py
@@ -47,30 +47,26 @@ async def test_async_setup_entry(
     :param mock_coordinator: The mock coordinator fixture.
     :type mock_coordinator: MagicMock
     """
+    # Arrange
     entry = MagicMock()
     entry.entry_id = "test_entry"
     hass.data[DOMAIN] = {entry.entry_id: mock_coordinator}
 
     mock_add_entities = MagicMock()
-
-    with patch("custom_components.ramses_cc.binary_sensor.entity_platform"):
-        await async_setup_entry(hass, entry, mock_add_entities)
-
-    assert mock_coordinator.async_register_platform.called
-
-    # Extract the callback passed to async_register_platform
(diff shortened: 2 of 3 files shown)


PWhite-Eng:
@silverailscolo, there are still 3 failing tests in ramses_cc, but this PR fixed another test

silverailscolo (approved):

PR #718, opened 2026-06-12
test: add deterministic queue flushing to prevent test teardown race conditions

PWhite-Eng:
### The Problem:

Tests in both the legacy (`tests_old`) and new (`tests_new`) suites occasionally suffer from intermittent failures and dropped connections during teardown. This happens because background Command-Query Responsibility Segregation (CQRS) queues—such as the gateway message queue, the engine message queue, and various Phase 2.95+ central dispatcher queues—are not cleanly or completely drained before the test environment is destroyed.

### Consequences:

If this isn't fixed, it results in flaky CI/CD test runs, lost packets, and noise in test logs due to connection teardown drops. These silent failures hide legitimate regressions and reduce developer confidence in the test suite.

### The Fix:

Introduced a hardcoded async helper function (`async_flush_queues`) to both test modules to explicitly and deterministically drain backend CQRS queues before the gateway is stopped or the Home Assistant components wrap up. Long comment blocks and tuple unpacking lines were also wrapped to comply with project linting guidelines.

### Technical Implementation:
- **Targeted Queue Identification:** The `async_flush_queues` function safely inspects the gateway object via `hasattr` and `getattr` to find `asyncio.Queue` instances across three layers: Legacy/Top-level Gateway (`msg_queue`), Engine Layer (`_msg_queue`), and Central Dispatcher (`_in_queue`, `ssot_queue`, `discovery_queue`, `binding_queue`, `faked_queue`).
- **Safe Awaiting:** Every isolated queue is joined using `asyncio.wait_for(q.join(), timeout=5.0)` wrapped in a `contextlib.suppress(asyncio.TimeoutError)` block to ensure the test suite never hangs indefinitely.
- **Event Loop Micro-ticks:** Added a sequence of 50 `await asyncio.sleep(0)` iterations at the end of the helper to ensure the event loop processes any lingering immediate tasks produced by synchronous `TopologyBuilder` iterations.
- **Hooks Added:** Called the helper inside `test_entities` immediately after `async_setup_component` and inside `instantiate_entities` right before `gwy.stop()`.

### Testing Performed:

Manually verified that running `pytest tests/tests_new/test_init.py` and `pytest tests/tests_old/test_evo_control.py` execute successfully. The added timeout suppression guarantees that even if a queue fails to join within 5 seconds, the tests will gracefully fall back rather than hanging the test process.

Continued intermittent test flakiness, particularly in local environments and automated CI pipelines, causing developer friction and unreliable merge signals.

### Risks of Implementing:
- Minor performance overhead added to test run execution times due to the intentional 50-tick loop sleep and queue waits.
- Since the helper uses hardcoded internal string paths (e.g., `_msg_queue`, `ssot_queue`), future refactoring of the core gateway queue names could silently render this cleanup helper ineffective until updated.

### Mitigation Steps:
- Addressed the test hang risk by wrapping the queue joins inside an explicit `asyncio.wait_for` timeout block alongside `contextlib.suppress`.
- Used defensive `hasattr`, `getattr`, and `isinstance(..., asyncio.Queue)` checks to prevent the helper from throwing `AttributeError` or crashing if core internals change.

### AI Assistance Disclosure:

### DIFF
--- a/tests/tests_new/test_init.py
+++ b/tests/tests_new/test_init.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import asyncio
+import contextlib
 from typing import Any
 from unittest.mock import AsyncMock, MagicMock, PropertyMock, patch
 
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo, we still have 3 failing tests, but they are now running to completion and failing, rather than crashing and failing. the work in ramses_rf (ramses-rf/ramses_rf#707) is also part of this work (so until there is a version bump this may not show up/help)

silverailscolo (approved):

PR #724, opened 2026-06-20
Refactor System Status Entities and Fix EvoControl Compatibility

PWhite-Eng:
### The Problem:

Following recent backend refactoring, the `RamsesSystemBinarySensor` was evaluating to `True` (as it returned the string ID). Because this sensor uses the `PROBLEM` device class, Home Assistant translated this to `STATE_ON` (Fault). Furthermore, stuffing a massive JSON routing schema into a fault-based binary sensor violates Home Assistant's architectural best practices.

### Consequences:

Physical EvoControl Wi-Fi displays drop offline or display false errors because their hardcoded API fetches explicitly expect `STATE_OFF` from `binary_sensor.{cid}_status` for a healthy system. Additionally, Home Assistant dashboard users are misled by false-positive fault states.

### The Fix:

Implemented a "Hybrid Approach" to satisfy both backward compatibility and modern architectural standards. The legacy binary sensor logic was corrected to accurately report system health, and a new dedicated `sys_info` sensor was introduced to host the schema payload moving forward.

### Technical Implementation:
- Updated `RamsesSystemBinarySensor.is_on` to return `not is_on` (yielding `STATE_OFF` for active systems) and explicitly documented it as a legacy shim.
- Added a new `sys_info` entity to `SENSOR_DESCRIPTIONS` in `sensor.py` to naturally expose the System ID string and host the unbounded `working_schema` JSON attribute.
- Updated `test_evo_control.py` to verify the legacy boolean namespace while routing schema checks to the new `sys_info` sensor.
- Incremented the `NUM_ENTS_AFTER` baseline in `test_services.py` and `test_init_data.py` to account for the new sensor.
- Resolved strict Pylance typing constraints by utilising `typing.get_args` for Union type matching in `test_sensor.py`, and `cast(Any, entity)` to safely bypass `@final` method overrides during `test_services.py` mock assignments.

### Testing Performed:

Ran the full `pytest` suite locally. `test_evo_control.py` now passes successfully. Iteration and service tests in `test_init_data.py` and `test_services.py` successfully assert against the new 48-entity baseline. Pylance and Mypy report zero errors. (Note: Snapshot test updates in `test_init.py` are deferred to a subsequent commit).

End-users relying on third-party EvoControl displays will experience completely broken integrations. The codebase will continue to abuse the diagnostic binary sensor class, risking future Home Assistant deprecation warnings and confusing custom dashboard developers.

### Risks of Implementing:

Very minimal. The entity count increases by one per system. Advanced users who built custom dashboard cards relying on the previously buggy `STATE_ON` behaviour of the legacy status sensor will see their dashboards revert to `STATE_OFF`.

### Mitigation Steps:

Retained the legacy `binary_sensor` specifically to ensure existing EvoControl tablets and basic automations do not instantly break (e.g., throwing HTTP 404 Entity Not Found errors). The new sensor uses a distinct `sys_info` key to strictly prevent `unique_id` collisions in the Home Assistant entity registry.

### AI Assistance Disclosure:

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -76,7 +76,9 @@ async def async_setup_entry(
     platform = entity_platform.async_get_current_platform()
 
     @callback
-    def add_devices(devices: RamsesRFEntity | Sequence[RamsesRFEntity]) -> None:
+    def add_devices(
+        devices: RamsesRFEntity | Sequence[RamsesRFEntity],
+    ) -> None:
         """Add new devices to the platform.
 
         :param devices: A list of RAMSES RF devices to be added.
@@ -204,20 +206,24 @@ class RamsesLogbookBinarySensor(RamsesBinarySensor):
 
 
 class RamsesSystemBinarySensor(RamsesBinarySensor):
-    """Representation of a system (a controller)."""
+    """Legacy representation of a system for EvoControl compatibility.
+
+    NOTE: This entity exists purely to serve the working_schema JSON to the
+    EvoControl Wi-Fi display. It evaluates to False (STATE_OFF) to indicate
+    a healthy system. For actual fault detection, use the active_fault sensor.
+    """
 
     _device: System
 
     @property
     def is_on(self) -> bool | None:
-        """Return True if the system has a problem.
+        """Return False (STATE_OFF) to satisfy EvoControl's health check.
 
-        :return: True if a problem exists, None if unknown.
+        :return: False if system is present, None if unknown.
         :rtype: bool | None
         """
         is_on = super().is_on
-        return None if is_on is None else is_on
-        # no status sensor exposed in _rf 0.57.0 gwy
+        return None if is_on is None else not is_on
 
 
 class RamsesGatewayBinarySensor(RamsesBinarySensor):
@@ -238,7 +244,8 @@ class RamsesGatewayBinarySensor(RamsesBinarySensor):
         engine = getattr(gwy, "_engine", None)
         gwy_config = getattr(gwy, "config", getattr(gwy, "_gwy_config", None))
 
-        # TODO Q3 2026: return await gwy._config() (only) instead of all below code
+        # TODO Q3 2026: return await gwy._config() (only) instead of all below
+        # code
         # not yet working: self._cached_attrs = await gwy._config()
         known_list: Any = getattr(gwy_config, "known_list", None)
         if not isinstance(known_list, dict):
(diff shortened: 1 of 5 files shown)


PWhite-Eng:
**@silverailscolo - Context on the `RamsesSystemBinarySensor.is_on` logic inversion:**

If the `not is_on` logic feels backwards, it is because we are dealing with a collision between Home Assistant's strict device class semantics and Python's boolean evaluation of strings.

Here is exactly why this logic is correct, and why the recent `ramses_rf` refactor exposed the bug in `test_evo_control.py`.

#### 1. The Home Assistant Semantic Layer (Why it feels backwards)

This entity uses `device_class=BinarySensorDeviceClass.PROBLEM`.
In Home Assistant's state machine, a `PROBLEM` sensor has strict rules:
- `is_on = True` ➔ `STATE_ON` ➔ UI shows **"Problem Detected"**
- `is_on = False` ➔ `STATE_OFF` ➔ UI shows **"System OK / Healthy"**

The EvoControl Wi-Fi tablet is hardcoded to look at this entity and expects `STATE_OFF`. If it sees `STATE_ON`, it assumes the Evohome network has crashed. Therefore, our mathematical goal is to force `is_on` to equal `False`.

#### 2. The Python Evaluation

In `BINARY_SENSOR_DESCRIPTIONS`, this entity is mapped to `ramses_rf_attr="id"`.
When `super().is_on` runs, it fetches the System ID (e.g., the string `"01:145038"`).
In Python, evaluating a populated string always returns `True`:
`bool("01:145038") == True`

Because it evaluated to `True`, HA set the state to `STATE_ON` ("Problem Detected"), failing the EvoControl test.

By applying `not is_on`, we invert the `True` (System ID exists) to `False`. Home Assistant translates this `False` to `STATE_OFF`, which the `PROBLEM` class correctly renders as "Healthy", keeping EvoControl happy. (Note: This exact same inversion is already used successfully in `RamsesGatewayBinarySensor` just below it).

#### 3. Why did `test_evo_control.py` pass BEFORE the `ramses_rf` refactor?

The test previously passed because of a "happy accident" regarding how async properties evaluated to `False`.

Before the refactor tightened up the backend, it is highly likely that `resolve_async_attr` was failing to instantly resolve the attribute during the strict execution window of the Pytest run (e.g., pulling an un-awaited coroutine, an empty status dictionary `{}`, or simply hitting the `default=None` fallback).

Because `bool({})`, `bool(None)`, and `bool(False)` all evaluate to `False`, the sensor accidentally defaulted to `STATE_OFF` ("Healthy"), and the test passed.

Now that `ramses_rf` has been refactored and the engine is highly deterministic, `resolve_async_attr` correctly and instantly resolves the `"id"` string synchronously during the test. This deterministic fix exposed the underlying logic bug—because the string is finally evaluating as `True` (`STATE_ON`).

By explicitly inverting it with `not is_on`, we are making the "Healthy" state deliberate rather than relying on an accidental falsy evaluation.

silverailscolo (approved):

PR #735, opened 2026-06-26
Chore: Remove _cc filter_remaining RQ hack

silverailscolo:
Root issue has been fixed in _rf
Also correct sensor docstr


silverailscolo:
Will only pass once some ramses_rf issues have been fixed and released.

Note: Filter_remaining not yet updating without this patch, so on hold

silverailscolo:
Better squash_merge this one!

PR #737, opened 2026-06-26
fix: defer SIGNAL_UPDATE to avoid stale hvac_state reads

wimpie70:
~Don't merge this yet or only after proper testing... I'm running out of time...prolly not before monday i can have a look...~ Edit: See results below > ready to merge

The protocol's extra handler list (which includes the learn_event's async_process_msg) is called synchronously from _msg_received, before the Gateway's async _msg_handler coroutine (scheduled via create_task) has a chance to run.  The _msg_handler coroutine contains await points (topology_builder.consume, process_msg) that yield to the event loop before _cqrs_ingestion_engine executes and updates hvac_state.

This means entities receiving SIGNAL_UPDATE would call async_write_ha_state and read stale hvac_state values.

Fix: defer the SIGNAL_UPDATE send by 50ms via async_create_task, ensuring all pending coroutines in the current batch (including CQRS ingestion) have completed before entities read device state.

NOTE: This is a pragmatic fix. The 50ms delay is a heuristic that works in practice but needs further testing. A more robust solution would be to have ramses_rf expose a post-ingestion hook that ramses_cc can register for, eliminating the timing dependency entirely.

### DIFF
--- a/custom_components/ramses_cc/event.py
+++ b/custom_components/ramses_cc/event.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import re
 from collections.abc import Callable
@@ -186,10 +187,23 @@ class RamsesLearnEvent(RamsesEvent):
             except PacketInvalid:
                 return
 
-            # what is the purpose of next block?
-            async_dispatcher_send(self._hass, f"{SIGNAL_UPDATE}_{msg.src.id}")
-            if msg.dst and msg.dst.id != msg.src.id:
-                async_dispatcher_send(self._hass, f"{SIGNAL_UPDATE}_{msg.dst.id}")
+            # Defer SIGNAL_UPDATE so that the Gateway's async _msg_handler
+            # (CQRS ingestion) has completed.  The protocol calls extra
+            # handlers synchronously before the Gateway's create_task'd
+            # coroutine runs, and the coroutine has internal await points
+            # that yield to the event loop before _cqrs_ingestion_engine
+            # executes.  Using a small delay ensures we run after all
+            # pending coroutines in the current batch have completed.
+            src_id = msg.src.id
+            dst_id = msg.dst.id if msg.dst and msg.dst.id != msg.src.id else None
+
+            async def _send_signals() -> None:
+                await asyncio.sleep(0.05)
+                async_dispatcher_send(self._hass, f"{SIGNAL_UPDATE}_{src_id}")
+                if dst_id:
+                    async_dispatcher_send(self._hass, f"{SIGNAL_UPDATE}_{dst_id}")
+
+            self._hass.async_create_task(_send_signals())
 
             if (
                 coordinator.learn_device_id is not None
--- a/tests/tests_new/test_event.py
+++ b/tests/tests_new/test_event.py
@@ -1,5 +1,6 @@
 """Tests for the RamsesEvent class."""
 
+import asyncio
 import re
 from collections.abc import Callable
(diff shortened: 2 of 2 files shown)


wimpie70:
i'm seeing different values between my production and dev HA...no time to check that now....
But at least you can see what we are trying to do...

silverailscolo:
@wimpie70 Confident on this, or better leave this out of 0.57.4?

Will go into 0.57.5 with _rf counterpart.

wimpie70:
The analysis is correct and the fix is sound. The protocol flow is accurately described: _msg_received calls the Gateway's _msg_handler coroutine via create_task, then immediately invokes the extra handlers (including async_process_msg) synchronously. Since the coroutine has await points that yield before _cqrs_ingestion_engine runs, SIGNAL_UPDATE fires before hvac_state is updated.

This is a pragmatic fix, not a structural one. The 50ms delay is a heuristic that works in practice — CQRS ingestion is just `dataclasses.replace`, so it completes well within 50ms under normal conditions. However, it is not deterministic: under heavy event loop load there is no guarantee the coroutine completes within the delay window. This should be replaced in the future with a proper post-ingestion hook in ramses_rf that ramses_cc can register for, eliminating the timing dependency entirely. The current approach is acceptable as an interim fix.

Important dependency: This PR is complementary to ramses_rf [PR 745](https://github.com/ramses-rf/ramses_rf/pull/745), which fixes a separate bug where SZ_REQ_REASON was passed as 'req_reason' (instead of 'request_reason') to `HvacState.__init__()` in `dispatcher.py`, causing a TypeError on every 2`210` packet.
That error silently failed the entire `hvac_state` update — not just request_reason. Without _rf [PR 745](https://github.com/ramses-rf/ramses_rf/pull/745), deferring SIGNAL_UPDATE doesn't help because the state was never correctly updated in the first place.

Recommendation: Safe to merge after testing with _rf [PR 745](https://github.com/ramses-rf/ramses_rf/pull/745). The combination should resolve #742: _rf [PR 745](https://github.com/ramses-rf/ramses_rf/pull/745) ensures state is correctly written, _cc #737 ensures entities read it after the write.

@PWhite-Eng can you give some input on this ?

wimpie70:
Tested on dev HA alongside ramses_rf [PR 745](https://github.com/ramses-rf/ramses_rf/pull/745) — confirmed working together.

After installing both fixes:

- 0 CQRS state extraction errors
- 0 ramses_rf/cc/tx errors at startup or during message processing
- Number platform loads correctly (was previously blocked by an unrelated import error that is also resolved by the updated ramses_rf)
- 2210 packets are processed and SIGNAL_UPDATE is deferred correctly.

The 5 new tests in test_event.py verify:

- SIGNAL_UPDATE is not sent synchronously (scheduled via async_create_task)
- Both src and dst signals are sent when the deferred coroutine runs
- No duplicate signal when src == dst
- `asyncio.sleep` is awaited before any signal is sent
- The learn_ event update_data call remains synchronous

Both PRs are complementary:
- _rf [PR 745](https://github.com/ramses-rf/ramses_rf/pull/745) ensures hvac_state is correctly written,
- _cc #737 ensures entities read it after the write.

The combination resolves the stale-state read pattern described in the PR.

silverailscolo (approved):

wimpie70:
Note, since this is a pragmatic fix, we may need to open a new issue to remind us this could use a more solid fix and should be replaced in the future with a proper post-ingestion hook in ramses_rf that ramses_cc can register for, eliminating the timing dependency entirely.

PR #741, opened 2026-06-27
Fix failing tests following PR #738 gateway state changes

PWhite-Eng:
### The Problem:

PR #738 updated the gateway binary sensor to return its actual `is_active` state without flipping the boolean value. However, the associated assertions in `test_binary_sensor.py` and the Syrupy snapshots in `test_init.py` were not updated to reflect this new logic, causing the test suite to fail.

### Consequences:

If left unaddressed, the continuous integration (CI) pipelines will consistently fail, masking potential regressions, hindering future development, and preventing successful releases.

### The Fix:

Updated the test suite to expect the correct, un-inverted gateway states, and regenerated the snapshot baselines.

### Technical Implementation:
- Modified the assertions in `test_gateway_binary_sensor_state` so that `is_active = True` now asserts `is_on is True`.
- Executed `pytest --snapshot-update` to capture the new baseline state (`'off'` instead of `'on'`) for the gateway entity in `test_init.py`.
- Replaced ambiguous `Any` type hints on sensor variables with explicit domain classes (e.g., `RamsesGatewayBinarySensor`) to ensure strict Mypy compliance.

### Testing Performed:
- Executed the complete `pytest` suite locally; all tests now pass.
- Ran Mypy in strict mode, confirming zero type errors.

A persistently broken CI pipeline, which degrades developer trust in the test suite and blocks automated deployment workflows.

### Risks of Implementing:

Minimal to none, as these changes exclusively target the test suite and do not alter production logic.

### Mitigation Steps:

Comprehensive local test executions and static analysis were performed to verify that the updated assertions accurately map to the latest codebase state.

### AI Assistance Disclosure:

### DIFF
--- a/tests/tests_new/snapshots/test_init.ambr
+++ b/tests/tests_new/snapshots/test_init.ambr
@@ -104,7 +104,7 @@
       'last_changed': <ANY>,
       'last_reported': <ANY>,
       'last_updated': <ANY>,
-      'state': 'on',
+      'state': 'off',
     }),
     StateSnapshot({
       'attributes': ReadOnlyDict({
@@ -228,7 +228,7 @@
       'last_changed': <ANY>,
       'last_reported': <ANY>,
       'last_updated': <ANY>,
-      'state': 'on',
+      'state': 'off',
     }),
     StateSnapshot({
       'attributes': ReadOnlyDict({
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
Hi @silverailscolo,
Thanks for pushing the updates in PR #738. It looks like the removal of the boolean inversion on the gateway sensor caused a few assertions and Syrupy snapshots to fail in the test suite.
I have quickly put together this PR to align the tests with your new logic, update the snapshots, and restore a green CI build. I also took the opportunity to upgrade a few lazy type hints in the test file whilst I was there.
Phil

silverailscolo (approved):

PR #743, opened 2026-06-27
ci: Implement dual-workflow architecture for safe PR coverage comments (#726)

### LINKED ISSUE #726: CI: Implement Dual-Workflow Architecture for Safe PR Coverage Comments

PWhite-Eng:
### The Problem:

Currently, the `check-cov.yml` workflow relies on the `pull_request_target` trigger to secure permissions for posting coverage comments. However, this intentionally forces GitHub to check out and test the `master` branch instead of the incoming PR branch. If underlying dependencies (like Home Assistant core or `ramses_rf`) update, the old `master` code fails against the new requirements, causing false CI failures and completely ignoring the actual PR code (Resolves #726).

### Consequences:

If left unaddressed, the CI pipeline will continuously suffer from false failures due to dependency mismatches. Furthermore, any coverage reports that do successfully post are actually reporting the metrics of the `master` branch, providing entirely inaccurate feedback for the developer's PR.

### The Fix:

Split the coverage reporting process into a secure, decoupled dual-workflow architecture using GitHub Actions' upload/download artifact pattern.

### Technical Implementation:
- **Workflow A (Generator):** Modified `check-cov.yml` to use standard `pull_request` triggers with strict `contents: read` permissions. This securely tests the actual incoming PR code.
- **Artifact Bridge:** Configured `actions/upload-artifact@v7` to export the resulting `coverage.xml` alongside a newly generated `pr-number.txt` file to establish context.
- **Workflow B (Consumer):** Introduced a new `post-cov.yml` workflow, triggered by `workflow_run`. This runs in the trusted base repository context, downloads the data via `actions/download-artifact@v8`, reads the PR number, and securely posts the comment using elevated `pull-requests: write` permissions.
- **Bundled Fix:** Aligned `test_init.ambr` Syrupy snapshots with Home Assistant 2026.7 Enum attributes to restore a green test matrix.

### Testing Performed:

Verified the generation of the `coverage-artifact.zip` containing the exact target PR number and correct XML schema. Executed local Pytest matrix following a dependency upgrade to validate the snapshot fixes.

Persistently broken continuous integration pipelines, lack of accurate code coverage metrics on new features, and the continued use of `pull_request_target`, which is flagged by GitHub as a potential security risk when handling fork code.

### Risks of Implementing:

The decoupled nature of `workflow_run` means failures in the bot commenting step will occur asynchronously, making it slightly harder to debug at a glance compared to a single unified workflow.

### Mitigation Steps:

Explicitly pinned the major versions of the artifact actions (`v7` and `v8`) to ensure API compatibility across the boundary. Added a `continue-on-error` outcome check to gracefully handle skipped paths if no Python files are altered in a PR.

### AI Assistance Disclosure:

---

### DIFF
--- a/.github/workflows/post-cov.yml
+++ b/.github/workflows/post-cov.yml
@@ -1,0 +1,47 @@
+name: Post Coverage Comment
+
+on:
+  workflow_run:
+    workflows: ["Coverage"]
+    types:
+      - completed
+
+permissions:
+  pull-requests: write
+  actions: read
+
+jobs:
+  post-comment:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+
+    steps:
+      - name: Download coverage artifact
+        id: download
+        uses: actions/download-artifact@v8
+        continue-on-error: true
+        with:
+          name: coverage-artifact
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+
+      - name: Read PR number
+        if: steps.download.outcome == 'success'
+        id: pr-number
+        run: echo "PR_NUMBER=$(cat pr-number.txt)" >> $GITHUB_ENV
+
+      - name: Add Pytest Coverage comment
+        if: steps.download.outcome == 'success'
+        # can only run on pr_target repository originally, now runs
+        # safely via artifact decoupling
+        uses: MishaKav/pytest-coverage-comment@v1
+        with:
+          pytest-xml-coverage-path: ./coverage.xml
+          issue-number: ${{ env.PR_NUMBER }}
+          # custom settings
+          report-only-changed-files: true
+          remove-links-to-files: true
+
+      - run: echo "🍏 This job's status is ${{ job.status }}."
--- a/.github/workflows/check-cov.yml
+++ b/.github/workflows/check-cov.yml
@@ -12,7 +12,7 @@ on:
       "tests/**",
     ]
 
-  pull_request_target:
+  pull_request:
     branches: [ "master" ]
     paths: [
       ".github/workflows/check-cov.yml",
@@ -28,7 +28,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  pull-requests: write
+  contents: read
 
 jobs:
 
@@ -104,14 +104,18 @@ jobs:
           output: both
           thresholds: '90 97'
 
-      - name: Add Pytest Coverage comment
-        # can only run on pr_target repository
-        if: github.event_name == 'pull_request_target'
-        uses: MishaKav/pytest-coverage-comment@v1
+      - name: Save PR number
+        if: github.event_name == 'pull_request'
+        run: echo ${{ github.event.pull_request.number }} > pr-number.txt
+
+      - name: Upload coverage artifact
+        if: github.event_name == 'pull_request'
+        uses: actions/upload-artifact@v7
         with:
-          pytest-xml-coverage-path: ./coverage.xml
-          # custom settings
-          report-only-changed-files: true
-          remove-links-to-files: true
+          name: coverage-artifact
+          path: |
+            coverage.xml
+            pr-number.txt
+          overwrite: true
 
       - run: echo "🍏 This job's status is ${{ job.status }}."
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -16,6 +16,8 @@ import sys
 from types import ModuleType
 from typing import Any
 
+# TEST: Triggering CI workflow
+
 # from collections.abc import Callable
 #
 # from homeassistant.components.event import EventEntity
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -16,8 +16,6 @@ import sys
 from types import ModuleType
 from typing import Any
 
-# TEST: Triggering CI workflow
-
 # from collections.abc import Callable
 #
 # from homeassistant.components.event import EventEntity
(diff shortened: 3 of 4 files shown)


.github/workflows/post-cov.yml
```diff
@@ -0,0 +1,47 @@
+name: Post Coverage Comment
+
```
silverailscolo, line 2: You can put this at the bottom of the initial workflow.
No need for an additional file, see example I pointed to earlier (ramses_rf/gh/publish).

PWhite-Eng:
Hi @silverailscolo,

This implements the dual-workflow architecture we agreed upon in #726.

A couple of quick notes for your review:
1. **The Failing Check:** You will notice one failing check at the bottom of the list (`pull_request_target`). This is a "ghost" run of the old YAML file currently residing on `master`, failing against the new HA 2026.7 dependencies. You can completely ignore it; it will vanish forever the moment this PR merges and replaces the old file on `master`.
2. **Where is the Coverage Comment?:** Because GitHub imposes a strict security rule, a new `workflow_run` listener (`post-cov.yml`) will not actively trigger until it has been merged into the default branch. However, I have verified the zip artifact is generating perfectly with the correct PR number bridged inside it. Once this merges, the commenting bot will activate on all future PRs.
3. **Bundled Snapshot Fix:** My local environment was slightly out of date (HA 2026.6). When the CI spun up here on 2026.7, the new strict Home Assistant Enum attributes caused the `test_init.ambr` snapshots to fail. I have upgraded my local environment, regenerated the snapshots, and bundled them into this PR. I acknowledge it should technically be a separate PR, but it seemed pragmatic to bundle them to secure a green test run here!

Let me know if you need any adjustments.
Phil

silverailscolo (approved):
Nice! Tests are all passing now

.github/workflows/post-cov.yml
PWhite-Eng: Hi @silverailscolo,

I completely agree with the spirit of your suggestion—keeping it consolidated in a single file using a `needs:` job (exactly like your `publish-hatch` pipeline) is much cleaner and was my first thought too.

The catch is how GitHub handles security for external forks. Because `publish-hatch` runs on a `release` trigger, it natively gets a `Write` token. However, `check-cov` triggers on `pull_request` events. When a PR comes from an untrusted external fork, GitHub strictly forces the entire workflow file into a `Read-Only` state to prevent malicious code from exfiltrating secrets.

If we append the commenting job to the bottom of that exact same file, it inherits that `Read-Only` token and is completely blocked from posting the comment to the PR.

To bypass this securely, GitHub Security explicitly recommends this two-file split. File A safely tests the untrusted fork code (Read-Only), and File B triggers completely separately via `workflow_run`. Because File B is a detached run triggered from our base repository, it safely gains the `pull-requests: write` token necessary to consume the artifact and post the comment.

Phil

PR #744, opened 2026-06-27
fix: use is_active attr and invert for PROBLEM device class

### LINKED ISSUE #665: Gateway status sensor stays On/Problem
Since installing Ramses RF 0.56.6, the MQTT Gateway/ramses_esp bridge Status binary_sensor is always showing _Problem_. Might be even on older releases.

No errors show up in HA system log. Tx and Rx work fine.

Note change to `GATEWAY_MESSAGE_TIMEOUT` mentioned in PR #537

wimpie70:
The gateway binary sensor used `ramses_rf_attr="status"` which returns an empty dict from `DeviceBase.status()`, not a boolean.  This caused the sensor to always show "OK" (False) regardless of gateway health.

Fix: change `ramses_rf_attr` back to `"is_active"` (the async property on `HgiGateway` that checks recent message timestamps) and invert the result so that `is_on=True` means "problem" (matching the `BinarySensorDeviceClass.PROBLEM` device class semantics).

Resolves #665.

depends on https://github.com/ramses-rf/ramses_rf/pull/736

@PWhite-Eng

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -321,11 +321,16 @@ class RamsesGatewayBinarySensor(RamsesBinarySensor):
     def is_on(self) -> bool | None:
         """Return True if the gateway has a problem (no recent messages).
 
+        `is_active` returns True when the gateway is healthy (recent
+        messages received).  Since this sensor uses
+        `BinarySensorDeviceClass.PROBLEM`, we invert so that `is_on=True`
+        means "problem" and `is_on=False` means "OK".
+
         :return: True if there is a problem, None if unknown.
         :rtype: bool | None
         """
         is_on = super().is_on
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
@wimpie70, can you just run `pip install -r ./requirements/requirements_dev.txt`, as I found this morning that that file had been updated but I had not updated my local system. I found that, pre updating, tests all passed locally, but then failed in the CI, once I updated, the tests started to fail locally, so I could fix them.
thanks
Phil

silverailscolo:
Thanks @wimpie70 
I added the ramses_rf ~and Hold~ labels as this PR will need a ramses_rf bump to pass. Running it locally now, works perfectly!

wimpie70:
only some cleanup to pass the test...

silverailscolo (approved):

PR #756, opened 2026-06-29
fix: find_param_entity unique_id mismatch with number platform

wimpie70:
The number platform creates parameter entities with unique_id format "{device.id}-param_{id}" (e.g., "32:153289-param_01"), but find_param_entity in fan_handler.py was looking for the old format "{device_id}_param_{id}" with underscores (e.g., "32_153289_param_01").

This mismatch meant find_param_entity never found the entities, so:
- The hourglass/pending state was never set when requesting parameters
- The refresh-all-parameters button appeared not to work (the backend was fetching values, but the UI entities never showed pending state or visually updated)

Fix: try the new unique_id format first, fall back to the old format for any entities that haven't been migrated yet.

I did have a lookup on what happened on changing to the new format:

The naming convention change history (all in number.py):

PR 581 (Apr 3, merged) — introduced new format {device.id}-{key} = 32:153289-param_01, with a migration step that converts old → new via ent_reg.async_update_entity(existing_entity_id, new_unique_id=new_unique_id). Resolves issue 516.
Commits 8a75a15 / 79a0b6e — temporarily reverted to underscore, then 2fc8b9c / ae168c5 — changed back to dash. Final state: dash format is mandatory.
The migration code at number.py:1148-1170 handles converting old unique_ids to new ones on startup.
Your prod entity registry confirms: all 16 param entities are already migrated to the new format (32:153289-param_01), zero old-format entries remain.

Why hourglasses show on startup but not on refresh button:

Startup: The number platform calls entity._request_parameter_value() directly on each entity object (number.py:268-274), which calls self.set_pending() directly — no registry lookup needed.
Refresh button: Goes through services.py → fan_handler.find_param_entity() which does a registry lookup by unique_id (fan_handler.py:47-66). This was using the old format 32_153289_param_01 — never found, so set_pending() was never called.
My fix is correct: tries the new format (32:153289-param_01) first, falls back to the old format (32_153289_param_01) for any entities that haven't been migrated yet. This matches exactly what _has_existing_param_entities in number.py already does (lines 94-111) — it checks both legacy_prefix and new_prefix.

### DIFF
--- a/custom_components/ramses_cc/fan_handler.py
+++ b/custom_components/ramses_cc/fan_handler.py
@@ -40,16 +40,45 @@ class RamsesFanHandler:
         Helper Method that searches for a number entity corresponding to a specific
         parameter on a device.
 
-        :param device_id: The ID of the device (e.g., '30:123456').
-        :param param_id: The 2-character hex ID of the parameter.
+        The callers pass a *normalized* device_id (colons replaced with
+        underscores, e.g. ``32_153289``) and an *uppercased* param_id
+        (e.g. ``3D``).  The number platform, however, stores entities with
+        the following unique_id formats:
+
+        * **New** (since PR 581): ``"{device.id}-param_{key}"`` where
+          ``device.id`` keeps its colons and ``key`` preserves the schema
+          case — e.g. ``32:153289-param_3D``.
+        * **Old** (pre-migration): ``"{normalized_id}_param_{param.lower()}"``
+          — e.g. ``32_153289_param_3d``.
+
+        We try the new format first and fall back to the old one so that
+        entities that have not yet been migrated are still found.
+
+        :param device_id: The ID of the device (normalized or raw,
+            e.g. ``32_153289`` or ``32:153289``).
+        :param param_id: The 2-character hex ID of the parameter (uppercased,
+            e.g. ``3D``).
         :return: The found entity or None if not found in the registry/platform.
         """
-        safe_device_id = str(device_id).replace(":", "_").lower()
-        unique_id = f"{safe_device_id}_param_{param_id.lower()}"
+        # Restore colons for the new unique_id format (device.id keeps colons)
+        colon_device_id = str(device_id).replace("_", ":")
+        normalized_device_id = str(device_id).replace(":", "_").lower()
+
+        # New format: "32:153289-param_3D" (colons, schema case)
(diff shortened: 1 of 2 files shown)


wimpie70:
Bugs in the first version
- Device ID format mismatch: The callers (services.py lines 248, 442) pass normalized_device_id = device.id.replace(":", "_") = 32_153289 (underscores). But the new unique_id format uses device.id with colons: 32:153289-param_3D. My first fix constructed 32_153289-param_3D — neither format, would never match.
- Param ID case: The entity key preserves the schema's original case (param_3D, uppercase), but my first fix lowercased it (param_3d). The old format used lowercase, the new format uses the schema case.

The correct fix
fan_handler.py:37-76
- New format: restores colons from underscores (replace("_", ":")) and uses param_id as-is (already uppercased by _get_param_id) → 32:153289-param_3D
- Old format: normalizes to lowercase underscores → 32_153289_param_3d (for any un-migrated entities)

Verification
- Your prod entity registry confirms: all 16 entities are in new format 32:153289-param_XX with colons and mixed case
- The prod logs confirmed the first fix was broken: Entity (unique_id=32_153289-param_75 or 32_153289_param_75) not found in registry
- Added Test 4 to verify the normalized device_id path (underscores → colons restoration)


- Device ID format: RAMSES device IDs are always XX:YYYYYY (digits + colon, no underscores in raw form). The replace("_", ":") only affects the normalized ID passed by callers, safely restoring colons. Verified against DEVICE_ID_REGEX in ramses_rf.
- Param ID case: All 16 keys in _2411_PARAMS_SCHEMA are uppercase. Callers uppercase param_id via _get_param_id. Entity keys preserve schema case. All three align — no case mismatch possible.
- Both formats coexisting: If both old and new unique_ids were in the registry (shouldn't happen post-migration, but defensively), we'd find the new one first. No conflict.
- Test coverage: Added Test 4 verifying the normalized device_id path (underscores → colons restoration) that services.py actually uses. Full suite: 407 passed, 10 skipped, 0 failures.
- No behavior change for existing callers: Both callers in services.py (lines 248, 442) pass normalized_device_id + uppercased param_id — exactly the inputs the fix is designed for. The only change is that the lookup now succeeds instead of silently failing.

silverailscolo (approved):
Any thoughts on the lingering timer test in Coverage failure? Will restart > No cov error on 2nd run

Issue #627, opened 2026-04-17
Finding devices on a clean system


wimpie70:
I'm not sure if the following will help with fixing the lost devices, but it might...

I'm still working on this sim, and what i found with it so far is that devices are now only recognized and found properly after a real (cached) msg from that device was processed.

On a fresh system (no cached msgs), when configuring a bound REM on the known devices, the bound REM will not be found (and wont be set) because no msg was sent by it yet.
Only after sending a msg from the REM, AND a restart, it will be set as bound_rem...

So, it doesnt matter if the device is specified in the known list or the schema (_cc) ....until a msg gets to us, it doesnt exist.
Shouldn't we iterate over the known device and schema and send detection msg (if we have them) ? Or create the devices anyways, cause the user says they exist ?

silverailscolo:
> Shouldn't we iterate over the known device and schema and send detection msg (if we have them) ? Or create the devices anyways, cause the user says they exist ?

Yes: iterate over the known devices and schema/orphans and send detection msg - from HGI ?
No: create the devices anyway, because the user says they exist - as long as we have manually typed schema I would want to check the actual address returned by the (real) device. For that, in Sim they would have to reply to the discovery code.
Let's collect the codes here:

## Signatures

| slug  | code           | meaning |
| ----- | -----------| -------- |
| FAN  | 31DA | fan state |
| REM  | 1060 | battery_state |

wimpie70:
@silverailscolo I'll release  again soon. Probably have most of these codes already, including replies.
The sim can play pre defined or imported 'conversations'. You can choose to filter out the replies and let the sim reply as if a real device would.
It can also load different profiles with most RF configs included.
And it will log msgs for each device separate (from/to)

wimpie70:
@silverailscolo, I noticed you are going to release soon. I'm preparing a fix for this, but probably won't have it ready today...so just choose if you want this in or on the next.

wimpie70:
# Proposed plan for issue 627 — discover_known_devices service call

## Summary

Add a `discover_known_devices` service call to ramses_cc that sends RQ
discovery messages from the HGI to devices in the known_list and schema
that are not yet registered. This follows silverailscolo's direction
("Yes: iterate over the known devices and schema/orphans and send
detection messages from HGI").

## How it works

1. Iterate over known_list + schema orphan device IDs
2. For each device, check if it's in `gwy.device_registry.device_by_id`:
[...]

Issue #677, opened 2026-05-13
0.56.8 Evohome device discovery inconsistent


peternash:
**Describe the bug**
Running 0.56.8 with Evohome the heating zone names are not being populated consistently and some OpenTherm sensors are `Unknown`.  After a cache clear all heating zones are `Unavailable`

**To Reproduce**
Steps to reproduce the behaviour:
1. Run 0.56.8 with Evohome, schema and state caches intact
2. Most (10 out of 12) heating zone names revert to the "RAD 01:216136_xx" form instead of their friendly names. Also, about half of the OpenTherm sensors are `Unknown`
3. Clear the schema and state caches and restart
4. All heating zones are now `Unavailable`. DHW control and Evohome climate control are discovered correctly. OpenTherm sensors discovered correctly although the same set of `Unknown` sensors apply.  Also, the `gateway_status` sensor displays `Problem` permanently.

Debug and packet logs when starting with intact caches:

[home-assistant.log](https://github.com/user-attachments/files/27706873/home-assistant.log)
[packet_log.log](https://github.com/user-attachments/files/27706874/packet_log.log)

Debug and packet logs when starting with cleared caches:

[home-assistant (1).log](https://github.com/user-attachments/files/27706933/home-assistant.1.log)
[packet_log.log](https://github.com/user-attachments/files/27706934/packet_log.log)

peternash:
@silverailscolo - looks like this may overlap with [#665](https://github.com/ramses-rf/ramses_cc/issues/665), [#627](https://github.com/ramses-rf/ramses_cc/issues/627) and [ramses_rf #649](https://github.com/ramses-rf/ramses_rf/issues/649).  Would you prefer if I closed this and added my logs to [#627](https://github.com/ramses-rf/ramses_cc/issues/627)?

silverailscolo:
@peternash thanks for the logs and profile.
No problem to find them here. You already created a link to the related items.

peternash:
Also, unfortunately the CPU ramping up issue is still present although it seems to have improved somewhat - see [#666](https://github.com/ramses-rf/ramses_cc/issues/666#issuecomment-4439354659)

silverailscolo:
@peternash Could you try running ramses_cc test release 0.56.9 and see how it behaves? If have just created a test release 0.56.9 for you. The library ramses_rf 0.56.9 is already up on PyPI.

peternash:
[...]

Issue #666, opened 2026-05-11
0.56.7 CPU ramp up and discovery issues


peternash:
**Describe the bug**
Testing 0.56.7 with Evohome shows that CPU usage ramps up over time, eventually causing HA to restart.  Also, heating zones are not discovered after a cache clear.  Some OpenTherm sensors continue to be `Unknown`

**To Reproduce**
Steps to reproduce the behaviour:
1. Run 0.56.7 with an Evohome system
2. Observe CPU usage over several hours
3. Also, after a cache clear heating zones are all `Unavailable`

1 hour profile, debug log and packet log attached for first hour after a restart.  On this run all heating zones were `Available` as they were restored from cache.  Some OpenTherm sensors  such as `boiler_setpoint` were `Available`, whereas others such as `water_pressure` remain `Unknown`

I've not attached a screenshot of CPU usage but it ramps up over a period of hours, eventually triggering a restart after about 15 hours.  This seems to be an improvement on 0.56.6 where a restart typically occurred after about 8-12 hours.

[home-assistant.log](https://github.com/user-attachments/files/27589326/home-assistant.log)
[packet_log.log](https://github.com/user-attachments/files/27589325/packet_log.log)
[profile.1778488228937614.cprof.zip](https://github.com/user-attachments/files/27589327/profile.1778488228937614.cprof.zip)

silverailscolo:
@PWhite-Eng Please look into this pattern.
A lot of repeating blocks that look like reboot/crash:

```
2026-05-11 10:25:06.096 DEBUG (RamsesStorage) [ramses_rf.sqlite_worker] Deleted specific message via queue.
2026-05-11 10:25:06.247 DEBUG (MainThread) [ramses_rf.gateway] Gateway: Resuming engine...
2026-05-11 10:25:06.260 INFO (MainThread) [custom_components.ramses_cc.coordinator] Saving the client state cache (packets, schema)
2026-05-11 10:25:07.740 INFO (MainThread) [custom_components.ramses_cc.coordinator] Discovery: Devices=27, Systems=1
[...]

PR #759, opened 2026-06-30
feat: add discover_known_devices service

wimpie70:
Add a new service call that sends RQ discovery messages from the HGI to devices in the known_list that are not yet registered or not yet promoted to their expected class. This fixes the issue where devices in known_list/schema aren't discovered until they send RF traffic on their own (issue 627).

The service iterates over known_list devices and classifies each as:
- not in registry → send detection RQ
- in registry but not promoted → send promotion RQ
- already OK → skip

Type-specific RQ codes are chosen based on the device class (if specified in known_list) or the address prefix. Codes that trigger both detection AND promotion are preferred (e.g. 31DA for FAN, 3220 for OTB, 10A0 for DHW).

Also adds a "discover_known_devices" boolean switch to the schema config step so users can trigger discovery on next startup without calling a service manually.

https://github.com/ramses-rf/ramses_cc/issues/627


silverailscolo:
@wimpie70 You coded this in ramses_cc, but I would prefer to use what is already in our code: DRY
Did you look at the `ramses_rf/discovery.py` provider? That code took care of discovery up to 0.55.3.

I can't get it to work yet, nor for polling the change_filter per device, but that could perhaps be fixed between us.

wimpie70:
I'll look into this tomorrow, just pushed the draft so you could have a look :)
I wanted to avoid _rf on this , not to cross phil's work...
W'll have a look

silverailscolo:
@PWhite-Eng You most certainly will have given `discover.py` some thought. Any idea why it has stopped doing what its name suggests?
Running discovery from ramses_rf allows the CLI/library to provide this service to more platforms than just HA.

wimpie70:
ah...that's not complete yet, toggle to discover didnt work yet...
tomorrow...

wimpie70:
long read, i did some extra digging:
# Plan: Fix issue 627 — Finding devices on a clean system

https://github.com/ramses-rf/ramses_cc/issues/627

## Problem

On a fresh system (no cached messages), devices configured in the
known_list or schema are not found until they send RF traffic. Only after
a real message from the device is processed AND a restart does the device
get recognised. This blocks new installations and recovery after cache
clear.

## Maintainer's direction

silverailscolo's comment on issue 627 (Apr 20, 2026):

> **Yes:** iterate over the known devices and schema/orphans and send
> detection messages from HGI.
> **No:** create the devices anyway, because the user says they exist —
> as long as we have manually typed schema I would want to check the
> actual address returned by the (real) device. For that, in Sim they
> would have to reply to the discovery code.

Later feedback (Jun 2026, PR review): **DRY principle** — don't
reimplement discovery code tables in ramses_cc. Use the existing
`ramses_rf` `DiscoveryService` and `_setup_discovery_cmds()` which
already know the right RQ codes per device class.

## Related issues & PRs

- https://github.com/ramses-rf/ramses_rf/issues/649 — CLOSED (0.57.3).
  silverailscolo found the DiscoveryService/polling was broken (never
  sending RQ commands) since 0.55.3 (PR 466). Fixed by PR 730 which
  re-adds explicit `_setup_discovery_cmds()` call after device creation
  in `DeviceRegistry.get_device()`, `_promote_device_class()`, and
  `_handle_promote_class()`.
- https://github.com/ramses-rf/ramses_rf/pull/707 — HGI bootstrapping.
  The active HGI is now explicitly registered in the device registry
  during `lifecycle.start()`, bypassing topology filters. **Only
  registers one HGI** (the first from known_list or discovered by
  transport).
- https://github.com/ramses-rf/ramses_rf/issues/691 — CLOSED (0.57.0).
  `known_list` not passed from ramses_cc to ramses_rf.
- https://github.com/ramses-rf/ramses_cc/issues/666 — references 627 and
  649 as known discovery issues.
- https://github.com/ramses-rf/ramses_cc/issues/677 — overlaps with 627
  (discovery after cache clear broken).

## What exists now (ramses_rf 0.57.3)

### ramses_rf side

#### DiscoveryService (`src/ramses_rf/discovery.py`)

- Composed into each entity (device, zone, system) via `self.discovery`
- `add_cmd(cmd, interval, delay, timeout)` — schedules a periodic RQ
- `discover()` — core loop: for each due command, checks if a recent
  message already exists, sends the RQ if needed, with exponential
  backoff on failure (1m → 4m → 7.5m → 15m → 30m → 1h)
- `deprecate_code_ctx()` — stops polling after 2 failures, can be
  reinstated
- **Only polls devices that already exist** — it does not discover new
  devices from known_list
- Disabled if `gwy.config.disable_discovery` is True or if the engine
  has sending disabled

#### `_setup_discovery_cmds()` per device class — coverage gaps

Each device class overrides this to register its own RQ codes.
**Critical finding: most HEAT device types have ZERO discovery
commands.** This is why HEAT device discovery doesn't work as expected.

| Class | File | Codes registered | Can respond to RQ? |
|---|---|---|---|
| `DeviceBase` | `dev_base.py:163` | none (pass) | — |
| `DeviceInfo` | `dev_base.py:270` | 10E0 (24h) | yes (RP) |
| `Controller` (01:) | `heat_controllers.py:65` | 2E04 (60min) | yes (RP 0006, 000C, 0100, 10E0, 1260, 1290) |
| `UfhController` (02:) | `heat_controllers.py:160` | 0005, 000C×8, zone cmds (6h) | yes (RP 0001, 0005, 000A, 000C, 2309) |
| `BdrSwitch` (13:) | `heat_actuators.py:193` | TPI params (6h), 3EF1 (60s/300s) | yes (RP 0008, 0016, 1FC9, 3EF1) |
| `OtbGateway` (10:) | `opentherm_bridge.py:152` | OT data IDs (5m/1h/6h) | yes (extensive RP list) |
| `HvacVentilator` (32:) | `hvac_ventilators.py:113` | 10D0 (24h) | yes (RP) |
| `RfsGateway` | `hvac_ventilators.py:495` | 22F1, 2411, etc. (24h) | yes |
| **`TrvActuator` (04:)** | `heat_actuators.py` | **ZERO** | **yes** (RQ 0004, 0016, 0100, 1F09, 313F) |
| **`DhwSensor` (07:)** | `heat_sensors.py` | **ZERO** | **yes** (RQ 0016, 10A0) |
| **`THM` (..)** | `heat_controllers.py` | **ZERO** | **yes** (RQ 0016, 1090, 2349, 3EF0, 3EF1) |
| **`OutSensor`** | `heat_sensors.py` | **ZERO** | **no** (no RQ codes at all) |
| **`Temperature`/`Setpoint`** | `heat_sensors.py` | **ZERO** | **no** |
| `HgiGateway` (18:) | `dev_base.py:456` | **none** — no override | no (gateway, not responder) |
| `Programmer` (23:) | | **ZERO** | yes (RP 1090, 10A0, 3EF1) |

**Key insight:** `_setup_discovery_cmds()` is called automatically by
`DeviceRegistry.get_device()` after device creation (PR 730 fix). So
creating a device via `get_device()` also sets up its discovery poller.

**Coverage gap:** TRV (04:), DHW sensor (07:), THM, and other HEAT
device types have ZERO discovery commands despite being able to respond
to RQ for certain codes. They are only detected via passive
eavesdropping. This is a ramses_rf limitation that should be addressed
(see Phase 2 below).

#### HgiGateway (`dev_base.py:456`)

- No `_setup_discovery_cmds()` override — 0 discovery commands
- Has `is_active()` (async) — checks if the gateway has received
  messages recently (uses `message_timeout` from config)
- Has `is_available` (property) — same check, inherited from DeviceBase
- **Does not respond to RQ 10E0** — confirmed by testing. HGIs are
  gateways, not responders. They broadcast `1F09` (system_sync) and
  relay messages, but don't answer RQs directed at them.

#### Multi-HGI support

- **Not supported.** `GatewayConfig.hgi_id` is a single string.
- `config.py:__post_init__()` takes the **first** HGI from known_list
  (`explicit_hgis[0]` or `implicit_hgis[0]`)
- `lifecycle.start()` only bootstraps one HGI (the configured one + the
  transport-discovered one)
- Other HGIs in known_list are **ignored** — they would only be created
  if they appear in message traffic
- The `gwy.hgi` property returns the **active** HGI (from transport's
  `SZ_ACTIVE_HGI` extra info)

#### Device creation flow

When a packet arrives from an unknown device:
1. `dispatcher.py:instantiate_devices()` calls
   `device_registry.get_device(device_id)`
2. `get_device()` determines class via `device_factory()` →
   `best_dev_role()` (uses traits, address type, message codes)
3. Device is created, `_setup_discovery_cmds()` is called (PR 730 fix)
4. Device added to registry
5. `TopologyBuilder` evaluates message and may emit
   `TopologyChangedEvent` (PROMOTE_CLASS, BIND_DEVICE, etc.)

#### TopologyChangedEvent and Phase 3 CQRS

- Immutable dataclass in `models/state_base.py` (lines 67-117)
- Actions: PROMOTE_CLASS, UPDATE_TRAITS, BIND_DEVICE, CREATE_CONTROLLER,
  CREATE_CIRCUIT
- Emitted by `TopologyBuilder` via a **single direct callback** to
  `DeviceRegistry.handle_topology_event()` (no event bus)
- The registry has an event router dict mapping each action to a handler
  method (`dev_registry.py:86-112`)

**Phase 3 CQRS status:**
- `_ACTIVE_DEGRADATION = False` in `dev_registry.py:753` — the CQRS
  shadow state is built but not yet active (guarded by this flag)
- Phil (PWhite-Eng) is actively working on **Phase 2.95**: migrating
  internal state getters to CQRS read-models (UFH, actuators,
  OpenTherm done; recent commits: `87762ed5`, `321a5fc0`, `078994e5`)
- Phase 3 (external event bus) has not been started — no branches, no
  commits, no TODOs assigned to Phil

**Could we propose a PR for the event bus?**
A minimal PR would add `register_topology_event_callback()` on
`Gateway` and support multiple callbacks in `TopologyBuilder`. This
touches:
- `gateway.py` (constructor wiring)
- `topology_builder.py` (callback list instead of single callback)

This does **not** cross Phil's active work — he's working on internal
state migration (`devices/`, `state/`), not the external event API.
However, the event bus is architecturally Phase 3, and Phil may have
opinions on its design. **Recommendation: ask Phil before proposing a
PR.** The minimal callback registration is low-risk, but the full event
bus design (typed events, filtering, lifecycle) should be coordinated.

#### Schema and zone discovery

The schema (from config or cache) contains structured zone information:

```
SCH_TCS = {
    SZ_SYSTEM:     { SZ_APPLIANCE_CONTROL: "<10: or 13: device_id>" }
    SZ_DHW_SYSTEM: { SZ_SENSOR: "<07: device_id>", SZ_DHW_VALVE: "<13:>", SZ_HTG_VALVE: "<13:>" }
    SZ_UFH_SYSTEM: { "<02: ufc_id>": { SZ_CIRCUITS: { "01": { SZ_ZONE_IDX: "01" } } } }
    SZ_ZONES:      { "01": { SZ_CLASS: "valve", SZ_SENSOR: "<08:>", SZ_ACTUATORS: ["<04:>", ...] } }
    SZ_ORPHANS:    ["<device_id>", ...]
}
SZ_ORPHANS_HEAT: ["<device_id>", ...]  # global, outside TCS
SZ_ORPHANS_HVAC: ["<device_id>", ...]
SZ_MAIN_TCS:     "<01: ctl_id>"
```

**How zones are discovered in ramses_rf:**
- The TCS (`systems/tcs.py:157`) sends `RQ 000C` to the CTL with 3
  payloads: `00F1` (appliance_control), `00F0` (hotwater_valve),
  `01F0` (heating_valve). The CTL responds with `RP 000C` containing
  the device map for those roles.
- **This only queries domain-level actuators/valves** — it does NOT
  query individual zone sensors or TRVs.
- The CTL does NOT respond to RQ for zone-level codes: `000A` (zone
  params), `2309` (setpoint), `30C9` (temperature) — these are
  broadcast-only (I packets).
- Zone sensor and TRV discovery is **entirely passive**: relies on the
  CTL broadcasting `000C` I packets (which include zone→device
  bindings) and on devices broadcasting telemetry.

**What this means for discovery:**
- We can force-create the CTL from the schema/known_list — its TCS
  will then send `RQ 000C` to discover bound actuators/valves.
- We can force-create zone sensors and TRVs from the schema — but they
  have ZERO discovery commands, so they won't be actively probed.
- These devices will only be "verified" when they send traffic, or
  when the CTL's `000C` response reveals them as bound to a zone.
- **The schema is the authoritative source for zone structure** — it
  tells us which devices are sensors, actuators, and which zones they
  belong to. We should use it, not just the known_list.

### ramses_cc side

- `coordinator.py:async_start()` — runs initial discovery, then
  schedules periodic `_async_discovery_task` every 60s. Discovery only
  picks up devices that already exist in
  `gwy.device_registry.device_by_id`.
- `services.py:async_send_packet()` — sends a raw command via
  `client.async_send_cmd(cmd)`.
- **Current implementation (branch `feature/discover-known-devices`):**
  - `services.py:async_discover_known_devices()` — iterates known_list
    only, force-creates non-HGI devices via `device_registry.get_device()`
    (which calls `_setup_discovery_cmds()`), then triggers
    `dev.discovery.discover()` for immediate polling. HGI devices are
    probed with RQ 10E0 (but this doesn't work — HGIs don't respond to
    RQs, and MQTT echoes the RQ back as a false positive).
  - **Does NOT iterate the schema** — zone sensors, actuators, and
    orphans defined in the schema are not force-created.
  - Service call only — no button entity (tried, didn't show up
    reliably in HA UI).

## Revised approach

### Principles

1. **known_list AND schema are ramses_cc concepts** — iterate both.
   The schema contains zone structure (sensors, actuators, zones) that
   the known_list doesn't. Use both sources to ensure all configured
   devices are created.
2. **DRY** — don't reimplement discovery code tables. Each device class
   already knows its RQ codes via `_setup_discovery_cmds()`. Creating a
   device via `get_device()` sets up its poller automatically.
3. **Don't blindly create devices** — silverailscolo's direction. For
   non-HGI devices, force-creation is acceptable because the discovery
   poller will verify they're real (failed polls get deprecated). For
   HGIs, skip entirely — they're gateways, not probes.
4. **Don't cross Phil's work** — Phil (PWhite-Eng) is actively working
   in `pipeline/`, `state/`, `devices/` on Phase 2.95 CQRS state
   migration. We only use public APIs:
   `device_registry.get_device()`, `dev.discovery.discover()`,
   `gwy.async_send_cmd()`.
5. **Accept the passive discovery limitation** — devices with zero
   discovery commands (TRV, DHW sensor, OUT sensor, THM) cannot be
   actively probed. They will be created from schema/known_list but
   only "verified" when they send traffic or when the CTL's 000C
   response reveals them. This is a ramses_rf limitation to address
   separately (Phase 2).

### Phase 1: Service call (current implementation, extended)

**`ramses_cc/services.py:async_discover_known_devices()`**

1. **Iterate over known_list** from coordinator options.
2. **Also iterate over the schema** to find devices not in known_list:
   - `SZ_MAIN_TCS` → the CTL (01:)
   - `SZ_APPLIANCE_CONTROL` → heat source (10: or 13:)
   - `SZ_DHW_SYSTEM` → DHW sensor (07:), DHW/HTG valves (13:)
   - `SZ_UFH_SYSTEM` → UFC (02:) and circuit zone indices
   - `SZ_ZONES` → zone sensors (08:) and actuators (04:, 13:)
   - `SZ_ORPHANS`, `SZ_ORPHANS_HEAT`, `SZ_ORPHANS_HVAC` → orphan devices
3. For each unique device_id from both sources:
   - **Skip the active HGI** (`client.hgi.id`)
   - **Skip HGI-class devices** — HGIs are gateways, not responders.
     They have 0 discovery commands and don't respond to RQs. They'll
     be detected naturally when they send traffic. Log a TODO for
     future multi-HGI support.
   - **Already in registry** → trigger `dev.discovery.discover()` to
     force an immediate poll cycle
   - **Not in registry** → force-create via
     `device_registry.get_device(device_id)` (this calls
     `_setup_discovery_cmds()` automatically), then trigger
     `dev.discovery.discover()`
   - **Special case: CTL (01:)** — when created, its TCS will
     automatically send `RQ 000C` to discover bound actuators/valves.
     This is the key mechanism for zone-level device discovery.
4. After probing, trigger `_discover_new_entities()` to update the HA
   entity registry.
5. Log a summary: N from known_list, N from schema, N already present,
   N created, N probed, N skipped (HGI), N with zero discovery cmds.

**What was removed from the original implementation:**
- Custom discovery code tables (`DISCOVERY_CODES_BY_CLASS`,
  `DISCOVERY_CODES_BY_PREFIX`, etc.) — replaced by
  `_setup_discovery_cmds()` in ramses_rf
- Config-flow switch (`CONF_DISCOVER_KNOWN_DEVICES`) — didn't persist
  reliably, and toggle-then-restart UX was clunky
- Button entity — didn't show up reliably in HA UI
- HGI probing with RQ 10E0 — HGIs don't respond to RQs, and MQTT echoes
  create false positives

### Phase 2: ramses_rf discovery command gaps (separate PR)

**Problem:** TRV (04:), DHW sensor (07:), THM, and other HEAT device
types have ZERO `_setup_discovery_cmds()` despite being able to respond
to RQ for certain codes:

| Device type | RQ codes it can respond to | Has discovery cmds? |
|---|---|---|
| TRV (04:) | 0004, 0016, 0100, 1F09, 313F | NO |
| DHW (07:) | 0016, 10A0 | NO |
| THM | 0016, 1090, 2349, 3EF0, 3EF1 | NO |
| Programmer (23:) | (RP 1090, 10A0, 3EF1) | NO |
| OUT sensor | (none) | NO — purely passive |
| Temperature/Setpoint | (none) | NO — purely passive |

**Proposed fix (ramses_rf PR):** Add `_setup_discovery_cmds()` overrides
for TRV, DHW, and THM that register the RQ codes they can respond to.
This would allow active probing of these devices during discovery.

**Does this cross Phil's work?** The `_setup_discovery_cmds()` method
is in the device class files (`heat_actuators.py`, `heat_sensors.py`,
`heat_controllers.py`). Phil has been touching some of these for CQRS
state migration, but the discovery command setup is separate from the
state getter migration. Low risk, but should be coordinated.

**Not blocking Phase 1** — even without these discovery commands,
devices are created from schema/known_list and will be detected when
they send traffic or when the CTL's 000C response reveals them.

### Phase 3: Event-driven discovery (future work, depends on Phil's Phase 3)

When Phil's CQRS event bus is ready, ramses_cc can listen to
`TopologyChangedEvent` directly and trigger entity creation reactively,
instead of polling `_discover_new_entities()` on a timer. This would
make device discovery instant when a new device sends traffic.

**Could we propose a PR for the event bus?**
A minimal PR would add `register_topology_event_callback()` on
`Gateway` and support multiple callbacks in `TopologyBuilder`. This
does not cross Phil's active Phase 2.95 work (internal state migration),
but the event bus is architecturally Phase 3. **Recommendation: ask
Phil before proposing a PR** — he may have design opinions on the
event bus API (typed events, filtering, lifecycle management).

**Not now** — depends on ramses_rf Phase 3 CQRS work and Phil's input.

### Phase 4: HGI handling (future work)

(I added this for some extra info)

**Current status:** Multi-HGI is not supported in ramses_rf. Only one
HGI is bootstrapped (the first from known_list). Other HGIs are ignored.

**TODO for future:**
- Add support in ramses_rf for multiple HGIs (config change: `hgi_id`
  → `hgi_ids` list, lifecycle bootstraps all of them)
- Or: add a method to ramses_rf to check if an HGI is online (e.g., send
  `1F09` sync and listen for response, or check message store for recent
  traffic from that HGI ID)
- For now: **skip HGI-class devices in the service call** with a log
  message: `"Skipping HGI 18:149488 (multi-HGI not supported, HGI will
  be detected when it sends traffic)"`

## What needs to change

### ramses_cc (Phase 1 — service call, extended)

| File | Change |
|------|--------|
| `services.py` | `async_discover_known_devices()`: iterate **both** known_list AND schema (zones, orphans, DHW, UFC, appliance_control), skip HGIs, force-create non-HGI devices via `get_device()`, trigger `discover()`, log summary. Remove the HGI RQ 10E0 probing code (doesn't work). |
| `services.yaml` | `discover_known_devices` service definition (already done) |
| `const.py` | `SVC_DISCOVER_KNOWN_DEVICES` constant (already done) |
| `schemas.py` | `SCH_DISCOVER_KNOWN_DEVICES` schema (already done) |
| `__init__.py` | Register/unregister the service (already done) |
| `coordinator.py` | Delegate service call to `ServiceHandler` (already done) |

### ramses_rf (Phase 2 — separate PR, not blocking)

| File | Change |
|------|--------|
| `heat_actuators.py` | Add `_setup_discovery_cmds()` to `TrvActuator` (04:) — register RQ 0004, 0016, 0100 |
| `heat_sensors.py` | Add `_setup_discovery_cmds()` to `DhwSensor` (07:) — register RQ 0016, 10A0 |
| `heat_controllers.py` | Add `_setup_discovery_cmds()` to THM class — register RQ 0016, 1090 |

**Coordinate with Phil** — he's active in these files for CQRS state
migration, but discovery command setup is separate from state getters.

### ramses_rf (Phase 3 — future, depends on Phil)

| File | Change |
|------|--------|
| `gateway.py` | Add `register_topology_event_callback()` method |
| `topology_builder.py` | Support multiple emit callbacks |

## Does this cross Phil's work?

**Phase 1 (ramses_cc only): No.** We use only public APIs:
- `device_registry.get_device()` — stable, Phil's PR 730 actually
  improved it by adding the `_setup_discovery_cmds()` call
- `dev.discovery.discover()` — public method on DiscoveryService
- `gwy.async_send_cmd(cmd)` — public method
- We do NOT touch `discovery.py`, `dev_registry.py`,
  `topology_builder.py`, `pipeline/`, `state/`, or any of Phil's active
  files

**Phase 2 (ramses_rf discovery commands): Low risk, coordinate.** The
`_setup_discovery_cmds()` method is in device class files that Phil has
been touching for CQRS state migration. However, discovery command
setup is separate from state getter migration. Should be a small,
focused PR that Phil can review.

**Phase 3 (event bus): Ask Phil first.** The event bus is
architecturally Phase 3 in ramses_rf. Phil is on Phase 2.95. A minimal
callback registration PR is low-risk, but the full design should be
coordinated.

## Testing

- Tested on live system with 7 known_list devices (2 HGI, 1 REM, 1 FAN,
  1 CO2, 1 DIS, 1 bound REM)
- Service correctly: skips HGIs, creates missing non-HGI devices,
  triggers discovery polling
- HGI probing with RQ 10E0 confirmed not working (HGIs don't respond,
  MQTT echoes create false positives) — removed
- All 511 ramses_cc tests pass, `make local-ci` passes
- **TODO:** extend tests to cover schema iteration (zones, orphans,
  DHW system, UFC system)

## Open questions for Phil

1. **Event bus (Phase 4):** Would a minimal PR adding
   `register_topology_event_callback()` on Gateway be welcome now, or
   should we wait for Phase 3? Does this cross your planned work?
2. **Discovery command gaps (Phase 2):** Would a PR adding
   `_setup_discovery_cmds()` to TRV (04:), DHW (07:), and THM be
   welcome? These device types can respond to RQ but have zero
   discovery commands. Would this conflict with your CQRS state
   migration in `heat_actuators.py` / `heat_sensors.py`?

## References

- https://github.com/ramses-rf/ramses_cc/issues/627 — the issue
- https://github.com/ramses-rf/ramses_rf/issues/649 — DiscoveryService
  broken (closed, 0.57.3, PR 730)
- https://github.com/ramses-rf/ramses_rf/pull/707 — HGI bootstrapping
- https://github.com/ramses-rf/ramses_rf/issues/691 — known_list not
  passed from ramses_cc (closed, 0.57.0)
- `ramses_rf/discovery.py` — DiscoveryService (entity-level polling)
- `ramses_rf/devices/dev_registry.py:get_device()` — device creation +
  `_setup_discovery_cmds()` hook
- `ramses_rf/devices/dev_registry.py:753` — `_ACTIVE_DEGRADATION` flag
  (Phase 3 CQRS gate)
- `ramses_rf/pipeline/topology_builder.py` — TopologyBuilder (heuristic
  device detection from messages, emits TopologyChangedEvent)
- `ramses_rf/models/state_base.py:67-117` — TopologyChangedEvent
  dataclass
- `ramses_rf/systems/tcs.py:157` — TCS `_setup_discovery_cmds()` (sends
  RQ 000C to CTL for appliance/valve discovery)
- `ramses_rf/schemas.py:82-194` — schema structure (SZ_MAIN_TCS,
  SZ_ZONES, SZ_ORPHANS_HEAT, etc.)
- `ramses_rf/protocol/ramses.py:809-849` — CTL protocol schema (which
  codes CTL responds to)
- `ramses_rf/lifecycle.py` — HGI bootstrapping (single HGI only)
- `ramses_rf/config.py` — GatewayConfig (single `hgi_id`, no multi-HGI)

wimpie70:
@PWhite-Eng There are some questions for you at the end of the previous post :smiley:

wimpie70:
I ran the new action call on the device sim, and 2 devices were created extra.
a TRV was created from the schema, but wouldn't be actively probed (Phase 2 gap).

wimpie70:
So current status: this PR and https://github.com/ramses-rf/ramses_rf/pull/758 (phase 2) are tested as much as I can with the device simulator.
So far I found only improvements in discovering devices with these 2 PR's in place.
Calling the action: ramses_cc.discover_known_devices isn't for all users, but we should keep it to test it on systems with missing devices.
I'll change this PR to run in the background. So it could be called on startup or on updating the config.

custom_components/ramses_cc/const.py
```diff
@@ -129,3 +129,7 @@ class ZoneMode(StrEnum):
     PERMANENT = "permanent_override"  # indefinitely
     COUNTDOWN = "countdown_override"  # for a number of minutes (max 1,215)
     TEMPORARY = "temporary_override"  # until a given date/time
+
+
+# Service name
```
silverailscolo: Put this above the first class definition

custom_components/ramses_cc/services.py
```diff
@@ -646,3 +663,284 @@ def _normalize_service_call(
...
+            zero_cmds,
+            len(zero_cmds_skip or []),
+        )
+
+        # TODO: Phase 3 — when ramses_rf exposes TopologyChangedEvent via an
+        # external callback API, listen to it here to trigger entity creation
+        # reactively instead of polling _discover_new_entities() on a timer.
+        # The minimal API would be:
+        #   client.register_topology_event_callback(self._on_topology_event)
+        # This depends on Phil's Phase 3 CQRS event bus work.
```
silverailscolo: Wouldn't mention a name in code/comment, just the task (clear label)

silverailscolo (changes requested):

wimpie70:
good catch on both

silverailscolo:
@wimpie70 nice feature for HVAC and debugging too. But to use in for CH, isn't there a loop/short?

Users of CH (heat) setups are told to "keep your schema minimal". In practice, that means: leave it empty in config, except for the hgi.

Users often ask: where do I find the address of my xxx-device. I tell them to look at a HA system log and spot some device messages using mqtt. But what they hope is the device address is printed on the package or on the product. They have to enter this in their schema.

Ramses RF up to 0.55.3 was able to fill in the schema using a set of strategies (eavesdropping, polling, ...).
That gave you a schema and yes, you could use it to generate devices, but without that egg...

@PWhite-Eng might be out of town atm

wimpie70:
Thanks for that question. I don't have a CH system myself (yet) so my info comes from the codebase and from user posts.
I did see a few users that had full schema's defined so that's sent me moving into this direction. 
Users with a minimum schema have less to gain from this, but not nothing. The service can still be useful in 3 scenarios:
So the service is useful in three scenarios:

- HVAC users — know device IDs from the hardware, put in known_list, service creates + probes
- CH users with a pre-built schema — have a full schema from a previous setup, service creates all devices at once
- CH users with minimal schema — service creates the CTL, but the rest still depends on 000C polling (less useful, but not useless — it gets the CTL started faster)

I investigated a bit more on eavesdropping and where it broke:

The eavesdropping flow exists in ramses_rf (dispatcher.py:204) but is disabled by default (i have an option on the device sim to enable it):

- enable_eavesdrop defaults to False in GatewayConfig
- ramses_cc does not expose enable_eavesdrop in its config schema at all
- enforce_known_list=True (the recommended setting) blocks any device not in known_list, even if eavesdropping were enabled

So on a clean CH system with minimal known_list, ramses_rf cannot discover devices from traffic. The chain that used to work (up to 0.55.3) was:
1. Eavesdrop on RF traffic → create devices from traffic (TRVs, relays, etc.)
2. CTL responds to RQ 000C → zone mapping learned
3. Schema built automatically

To restore that flow, we'd need to:
1. Expose enable_eavesdrop as a ramses_cc config option (or enable it by default)
2. Relax enforce_known_list to allow eavesdropped devices to be created (perhaps with a "pending" status until confirmed)
3. Or: keep enforce_known_list strict but have eavesdropping add to known_list automatically after first sighting

That's a bigger change in ramses_rf/ramses_cc, separate from this PR. This PR (discover_known_devices) is a pragmatic shortcut for users who already know their device IDs — it doesn't solve the eavesdropping gap.

But why was eavesdropping set to false ?:

It was intentionally set to False as part of Phil's CQRS refactoring (Phase 2.9, PR #639, May 2026). The refactoring extracted all eavesdropping heuristics from device classes (_handle_msg in base.py, heat.py, hvac.py) into a centralized TopologyBuilder engine.

The key design decision: heuristics were split into two categories:

1. Explicit bindings (000C zone device maps from CTLs) — these are authoritative configuration data from the CTL, so they're processed regardless of enable_eavesdrop
2. Heuristic eavesdropping (class promotion from broadcast codes, directed telemetry inference, passive prefix-based inference) — these are guesses, so they're gated behind enable_eavesdrop and disabled by default

The reason for disabling heuristics by default is the SRP (Single Responsibility Principle) concern: the old code had Read-Models (device entities) mutating the Write-Models (topology/graph) during packet parsing via self.__class__ = cls class-swapping. This was fragile and hard to reason about. The new TopologyBuilder centralizes this, but the heuristics are considered unreliable enough that they're opt-in.

What this means for CH users:

- 000C zone binding still works without eavesdrop — if the CTL is known and polled, its zone device maps are processed
- Class promotion from traffic doesn't work — a device broadcasting controller-only codes won't be promoted to CTL class
- Directed telemetry binding doesn't work — a TRV sending I 3150 to a CTL won't be auto-bound to a zone
- Passive prefix inference doesn't work — devices won't be created from address prefix alone

So the eavesdrop=False default is a deliberate architectural decision in the CQRS refactor, not a bug. The heuristics can be re-enabled by setting enable_eavesdrop=True, but ramses_cc doesn't expose that option. 

The fix would be to expose it in ramses_cc config, with appropriate documentation about the trade-offs. 


But I don't think that's good enough. It would add devices to HA that you don't want. So...how about a passive discovery scan ? Collect devices from the RF traffic ? See next comment (so it can be ref.).
(yeah, i let my assistant wrap it all up)

wimpie70:
# Plan: Passive device scan for unknown devices (Phase 4)

> Note: See follow up in #762

## Problem

On a clean CH system with minimal schema, users have no way to discover
unknown device addresses. Eavesdropping is disabled by default (CQRS refactor,
Phase 2.9), and `enforce_known_list=True` blocks unknown devices. Users must
manually spot device IDs in MQTT logs — which is tedious and error-prone.

## Concept

A passive scan that listens to RF traffic, identifies devices not yet in
known_list/schema/HA, and presents them to the user for review — without
creating any HA entities or touching the topology.

## Flow

```
RF traffic → msg_handler → filter unknown → classify → store → notify user
                                                              ↓
                                              user reviews list
                                                  ↓         ↓
                                             add to KL   discard
```

### 1. Start scan (automatic, always running)

Register a `msg_handler` on the ramses_rf client (same mechanism as
`event.py:139`). For every packet:

- Extract src, dst, addr3 device IDs
- Check if device is already in: known_list, schema, HA device registry
- If unknown → add to discovery list with metadata

### 2. Classify without adding

For each unknown device, gather info **without creating it in the registry**:

- **Address prefix** → likely device type (01:=CTL, 04:=TRV, 07:=DHW,
  10:=BDR, 13:=OTB, 18:=HGI, 22:=THM, 32:=FAN, 37:=REM, etc.)
- **Codes seen** → what messages this device sends/receives
  (e.g. I 3150 = heat demand → likely TRV, I 1060 = battery → battery-powered)
- **Binding hints** → if device sends to a known CTL, note the CTL ID
  and zone_idx from the payload (this is the 000C/3150/30C9 correlation)
- **Promotion check** (optional, careful): check if the device would be
  promoted by the TopologyBuilder heuristics, but **do not actually emit
  the event** — just record what class it would be promoted to

### 3. Store findings

Persist the discovery list to HA storage (`.storage/ramses_cc/`) so it
survives restarts. Each entry:

```json
{
  "device_id": "04:056053",
  "first_seen": "2026-07-01T13:06:14",
  "last_seen": "2026-07-01T13:42:08",
  "likely_type": "TRV",
  "codes_seen": ["3150", "1060", "30C9", "12B0"],
  "bound_to": "01:145038",
  "zone_idx": "02",
  "status": "new" | "reviewed" | "added" | "discarded"
}
```

### 4. Notify user

- **Persistent notification** in HA when new unknown devices are found
  ("Ramses CC found 2 unknown devices — click to review")
- **WebSocket command** for ramses_extras card to show a discovery list
  with add/discard buttons
- **Service call** `ramses_cc.get_discovered_devices` to retrieve the list
  programmatically

### 5. User action

User reviews the list and for each device chooses:

- **Add to known_list** — opens config flow with pre-filled device ID and
  suggested class, user confirms, device is added to known_list, then
  `discover_known_devices` can be called to create + probe it
- **Discard** — device stays in the list with status "discarded" so it
  won't trigger another notification (spam prevention). User can review
  discarded devices later and undo if needed.

## Key design decisions

### No entity creation
The scan never calls `get_device()` or adds to the registry. It only
observes packets and collects metadata. This means:
- No `enforce_known_list` conflict
- No `enable_eavesdrop` needed
- No topology mutations
- No HA entities created until the user explicitly adds to known_list

### Always running (not a timed scan)
The scan runs continuously as a msg_handler. This is lightweight — it's
just a dict lookup per packet. Devices appear as soon as they send traffic.
A timed scan would miss devices that only broadcast infrequently (e.g.
battery devices that wake every few minutes).

### Classification is best-effort
The classification from address prefix + codes seen is a hint, not
authoritative. The user makes the final decision. This avoids the
fragility problems that led to eavesdrop heuristics being disabled.

### Binding correlation
When a TRV sends `I 3150` to a known CTL, the payload contains `zone_idx`.
We can record "04:056053 is likely in zone 02 of CTL 01:145038". This is
the same info the CTL's 000C response gives, but obtained passively without
needing to poll the CTL.

## Implementation scope

### ramses_cc
- `services.py`: add `get_discovered_devices` and `add_discovered_device`
  service calls
- `coordinator.py`: add msg_handler for passive scan, store/load discovery
  list from `.storage/`
- `schemas.py`: add config option for scan (enable/disable, notification
  preference)
- New storage helper for the discovery list

### ramses_extras (optional, later)
- WebSocket command to retrieve discovery list
- Card UI with add/discard buttons and device classification display

### ramses_rf
- No changes needed — the scan uses the existing `add_msg_handler` API

## Relationship to issue 627

This is Phase 4 of the issue 627 plan:
- Phase 1: `discover_known_devices` service (PR 759) — create known devices
- Phase 2: discovery command gaps (PR 758) — probe devices that can respond
- Phase 3: CQRS event bus (future) — reactive entity creation
- **Phase 4: passive scan (this plan)** — find unknown device addresses

Phase 1 and 2 help users who already know their device IDs. Phase 4 helps
users who don't — the "chicken" that comes before the "egg".

silverailscolo (approved):

wimpie70:
i'm writing a new issue right now for this phase 4 plan

silverailscolo:
Thanks so far @wimpie70 & friends. Will merge and release.

The decision to turn off discovery since 0.55.3 wasn't made explicitly, or we would have known why evohome showed a few issues since. But no second thoughts on that now.

PR #775, opened 2026-07-05
feat: fresh-start flag for reliable cache clearing

wimpie70:
I'm testing on PR 764 with _extras simulator and run into a hard to tackle (without this PR) race condition. _extras runs after RF is started, but I cannot reload an empty (no schema, knownlist, dbase, etc) .

## Summary

Add a CONF_FRESH_START flag in the config entry options that tells
async_setup_entry to call Store.async_remove() before creating the
coordinator.  This invalidates the in-memory StorageManager cache AND
deletes the .storage file, ensuring the integration starts from a
clean slate.

The clear_cache flow now also removes stale HA device registry entries
when clear_schema is selected — without this, ramses_cc recreates all
old devices from the device registry on reload even though .storage
was wiped.

The check runs in async_setup_entry (before the coordinator is created)
so it can't be interrupted by a reload.  The flag is reset immediately
via async_update_entry so subsequent normal reloads preserve state.

### Why

When testing with the ramses_extras simulator (PR 764), there's a race
condition: the unload save (async_save_client_state via async_on_unload)
re-populates .storage after an external clear-cache operation has already
wiped it.  This flag solves that race by wiping .storage *after* the reload
starts, in async_setup_entry.

### Changes
- `const.py`: add CONF_FRESH_START constant
- `__init__.py`: check the flag in async_setup_entry, call
  Store.async_remove(), reset the flag
- `config_flow.py`: clear_cache step removes stale HA device registry
  entries and sets CONF_FRESH_START when clear_schema or clear_packets
  is selected
- `tests`: fresh-start removes .storage + flag is reset; normal setup
  preserves .storage; clear_cache removes devices + sets the flag

#### Test plan
- [x] `pytest tests/tests_new/` — 408 passed, 10 skipped
- [x] `ruff check` — all passed
- [x] Tested on ha-sim: clear_cache with all checkboxes → logs show
      `Clear cache: removed N stale HA device(s)`, `Fresh start requested`,
      `Fresh start: cleared .storage cache`, `Starting with 0 cached packets`
- [x] Normal restart preserves .storage (no fresh-start trigger)
- [x] Flag is reset after use (no reload loop)
- [x] AI Coding Assistance disclosure: GLM 5.2-High

**PR Summary**


silverailscolo (approved):
Fine tool, you found a good hook

PR #779, opened 2026-07-06
Fix VirtualRf race condition, add async system_mode cooldown

wimpie70:
The ramses_rf system_mode getter (PR 769) was changed from a sync property to an async method that dispatches a 2E04 RQ command when unpopulated. This caused ramses_cc to call it as a sync property, getting back a coroutine and retrying in an infinite loop, plus unbounded task creation from the RQ dispatch side-effect.

Fixes:
- virtual_rf.py: yield to event loop before cleanup, make _cleanup() robust (check closed, handle errors, clear maps), add FD guard in _read_ready, add _rx_buffer for frame fragmentation, handle BlockingIOError in _push_frame_to_dst_port
- helpers.py: add cooldown to resolve_async_attr to prevent tight retry loops when an async attribute returns None/unpopulated
- test_init_data.py: move rf.stop() into finally block for guaranteed cleanup
- test_virtual_rf.py: updated for VirtualRf changes

created with AI, not tested (yet)

### DIFF
--- a/custom_components/ramses_cc/helpers.py
+++ b/custom_components/ramses_cc/helpers.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import inspect
 import logging
+import time
 from contextlib import suppress
 from datetime import datetime as dt
 from typing import Any, cast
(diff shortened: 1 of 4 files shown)


silverailscolo:
Thanks for the fix.
Restarted the Coverage test, passed second time.

silverailscolo (approved):

PR #783, opened 2026-07-06
fix(tests): stop test_services_entry_ hanging on perpetual QoS retransmit

### LINKED ISSUE #774: Chore: Fix `services_entry` init test intermittent  failing
The CI failure in `tests/tests_old/test_init_data.py::test_services_entry_` is a pre-existing flaky test.

> As of release 0.57.9 this set of tests regularly times out after > 45 minutes 👎 

## Evidence

The test times out on clean master locally (both 60s and 100s timeouts) — git checkout master && pytest tests/tests_old/test_init_data.py::test_services_entry_ hangs the same way

A second CI run (28733407628) failed on test_init_data.py with the same code — purely timing-dependent

## Root cause

The test calls hass.async_block_till_done() after casting packets, but the ramses_rf QoS layer continuously retransmits RQ 2E04 to the CTL every ~1s, creating a perpetual stream of pending tasks that never lets async_block_till_done() settle. On a fast CI runner it finishes within the 60s timeout; on a slower run it doesn't.

## Fix
The test needs to either mock the QoS retransmit loop or not depend on async_block_till_done() with a perpetual transport.

_Originally posted by @wimpie70 in https://github.com/ramses-rf/ramses_cc/pull/764#issuecomment-4885295732_

AshleyAHuang:
Closes #774.

## Root cause

As diagnosed in the issue (thanks @wimpie70): the gateway's QoS layer keeps retransmitting any RQ the virtual RF never replies to (e.g. discovery's `RQ 2E04` to the CTL). Each retransmit schedules another pending task, so `hass.async_block_till_done()` can never fully settle — which is why the test occasionally hangs for 45+ minutes on CI instead of failing or passing quickly.

## Fix

- Set `disable_qos: True` in `TEST_CONFIG` — an existing `EngineConfig` field that already flows through the coordinator into the `Gateway`. With QoS disabled, sent commands aren't tracked or retried, removing the perpetual-retransmit source entirely.
- Wrapped the specific `hass.async_block_till_done()` call named in the issue in `asyncio.wait_for(..., timeout=30)`, so if anything else ever causes a similar stall, the test fails loudly within 30s instead of consuming the whole CI timeout budget.

## Testing

I set up the full dev/test environment locally and tried to reproduce the reported hang directly. I hit a different, 100%-reproducible failure in my environment (`assert 9 == 13`, fails in under a second, and the device count doesn't change even after several seconds of polling) — I confirmed this is a pre-existing, environment-specific issue unrelated to #774: it occurs identically with and without this fix, and it fails immediately rather than hanging, so it can't be the same mechanism. I did not attempt to fix that separate issue here to keep this PR scoped to what was actually reported.

I could not reproduce the actual 45-minute hang locally since it depends on CI-specific timing/load, so I can't personally confirm this resolves it end-to-end — but the fix directly targets the retransmit mechanism described in the issue via an existing, already-supported config option, and the change is a net-neutral no-op for every other assertion the test makes (verified identical pass/fail behavior on the unrelated local issue with and without this change).

ruff and mypy are clean on the changed file.

### DIFF
--- a/tests/tests_old/test_init_data.py
+++ b/tests/tests_old/test_init_data.py
@@ -56,7 +56,11 @@ NUM_DEVS_SETUP = 13  # Updated from 1 due to better packet decoding
 
 # Clean config to prevent HA schema validation failures
 TEST_CONFIG = {
-    CONF_RAMSES_RF: {"disable_discovery": True},
+    # disable_qos: without it, the gateway perpetually retransmits any RQ that
+    # the virtual RF never replies to (e.g. discovery's RQ 2E04 to the CTL),
+    # which keeps scheduling new pending tasks and can prevent
+    # hass.async_block_till_done() below from ever settling (see #774).
+    CONF_RAMSES_RF: {"disable_discovery": True, "disable_qos": True},
     SZ_SERIAL_PORT: {SZ_PORT_NAME: "/dev/ttyACM0"},
 }
 
@@ -141,7 +145,9 @@ async def test_services_entry_(
     try:
         gwy = list(hass.data[DOMAIN].values())[0].client
         await cast_packets_to_rf(rf, f"{TEST_DIR}/system_1.log", gwy=gwy)
-        await hass.async_block_till_done()
+        # Bounded so an unexpected perpetual-transport stall fails fast and
+        # clearly instead of hanging for the full CI timeout (see #774).
+        await asyncio.wait_for(hass.async_block_till_done(), timeout=30)
 
         await _test_common(hass, entry, rf)
     finally:


silverailscolo:
Testing CI failed once with 1 lingering task. Restarted, passed second run.

silverailscolo:
@wimpie70 Can you approve this change? It touches the same file `tests/tests_old/test_init_data.py` you updated this morning in PR #779

silverailscolo:
@AshleyAHuang Thanks for your PR. As the integration is actively being worked on and this is a crucial file, I want to make sure your changes align with the other devs.

wimpie70:
Nice finds,
small problem, it doesn't have our finally block. Here's what we found:

The two changes:

-1. disable_qos: True in TEST_CONFIG — Good. This is a valid existing config option that prevents the QoS layer from retransmitting unanswered RQ commands. This directly addresses the root cause of the hang: the 2E04 RQ dispatched by the system_mode side-effect (PR 774) gets retransmitted forever by QoS, creating endless pending tasks that prevent async_block_till_done() from settling.
2. asyncio.wait_for(..., timeout=30) around async_block_till_done() — Good. Defense-in-depth so a future similar issue fails fast instead of hanging for 45 minutes.

The problem:

The PR was branched before our PR 779 merge. It reverts our finally block fix for the rf fixture (moving rf.stop() back outside finally). If merged as-is, the cleanup race condition we fixed would come back.

Recommendation: The changes are good and safe to approve, but the merge needs to preserve our finally block. A rebase onto current master would resolve this automatically.

silverailscolo:
@AshleyAHuang In your IDE or using git, can you _rebase_ your branch onto current master ? Thanks again

AshleyAHuang:
Thanks for the careful review @wimpie70 — confirmed, the branch was cut before #779 merged, so it was carrying the old (non-`finally`) version of the `rf` fixture as its base and made it look like a revert. Rebased onto current `master` (force-pushed) — the diff is now scoped to just the two changes you called out (`disable_qos: True` + the bounded `asyncio.wait_for`), and #779's `finally: await rf.stop()` fix is preserved. Re-ran the full `test_init_data.py` suite plus ruff/mypy after the rebase, all clean.

silverailscolo (approved):

silverailscolo:
Testing run failed with
```
ERROR tests/tests_new/test_init.py::test_entities[minimal] - Failed: Lingering task after test <Task pending name='PortTransport._create_connection()' coro=<PortTransport._create_connection() running at /opt/hostedtoolcache/Python/3.14.6/x64/lib/python3.14/site-packages/ramses_tx/transport/port.py:259> wait_for=<Future pending cb=[Task.task_wakeup()] created at /opt/hostedtoolcache/Python/3.14.6/x64/lib/python3.14/asyncio/base_events.py:459> created at /opt/hostedtoolcache/Python/3.14.6/x64/lib/python3.14/site-packages/ramses_tx/transport/port.py:206>
```
Restarted, passed on second run
(Still some timing issue in test_init.py::test_entities)

PR #803, opened 2026-07-13
Fix: persist device_comments when sync_learned_topology returns None

wimpie70:
When the scan engine captures zone bindings from broadcast traffic (e.g. 3150 packets), refresh_device_comments updates the comments in a local config_schema variable.  However, when sync_learned_topology returns None (no topology changes), these updated comments were never persisted to self.options — they were silently lost.

This meant zone binding info from broadcast TRV traffic never survived to the next sync cycle, so step 0b/1g could never create zones from passive scan observations.

Fix: track a comments_refreshed flag and persist the config schema (with updated comments) even when sync_learned_topology returns None.

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -1737,6 +1737,7 @@ class RamsesCoordinator(DataUpdateCoordinator):
             # (where dst is --:------) that wasn't captured when the device was
             # first accepted.  This ensures sync_learned_topology has up-to-date
             # zone info in the comments.
+            comments_refreshed = False
             if self.discovery_manager and isinstance(config_schema, dict):
                 existing_comments = config_schema.get(SZ_DEVICE_COMMENTS, {})
                 if isinstance(existing_comments, dict):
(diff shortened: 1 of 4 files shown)


wimpie70:
Added a 'do not edit' remark to the root comments. This will need a better fix later.

The comments serve dual purpose and shouldn't. The structured data (zone_idx, bound_to) should be proper schema traits (like _zone_idx, _bound_to) rather than regex-parsed from free text.

They do get overwritten every by refresh_device_comments that runs every checkpoint, but there can be a race issue when users overwrite them. Still want to keep it for now, since it is also informative on what the system finds.

silverailscolo (approved):

silverailscolo:
> This will need a better fix later.

When we get to that, can existing stored info be converted?

Speaking about versions, a 0.58.0 schema will be rejected by 0.57.12, won't it? Just step forward, not back.

Really good to put it as a comment (of course I tried!). Manuals are often hard to find when you need them.

wimpie70:
see https://github.com/ramses-rf/ramses_cc/issues/809 2.5 and migration

PR #811, opened 2026-07-13
Feature/phase3a commands in schema

wimpie70:
Implements Phase 3a from issue https://github.com/ramses-rf/ramses_cc/issues/809
Depends on ramses_rf PR: strip_and_map_traits() + list bound in SCH_TRAITS_HVAC

## Summary

- Move learned RF commands from `.storage[remotes]` and `known_list[dev][commands]` into the schema as `_commands` per-device trait (SSOT)
- `.storage[remotes]` kept as cache/fallback, `known_list[dev][commands]` kept as legacy fallback
- **No storage version bump** — runtime migration via `_sync_remotes_to_schema` (bumping would break downgrade to 0.58.0/0.58.1: HA raises `UnsupportedStorageVersionError` when stored version > `max_readable_version`)
- `climate.set_fan_mode` reads custom commands from schema `_commands` (via `coordinator._remotes`) first, falls back to `known_list[commands]` (legacy)
- `fan_handler.setup_fan_bound_devices` accepts `str | list[str]` for `_bound` (multi-REM binding)
- `_derive_known_list_from_schema` uses ramses_rf's `strip_and_map_traits()` as base for trait extraction
- REM entity exposes `bound_to_fan` state attribute (derived from `fan_handler._fan_bound_to_remote`)

### ramses_rf dependency

This PR requires ramses_rf PR (feature/strip-and-map-pipeline) which adds:
- `strip_and_map_traits()` / `strip_and_map_schema()` — two-stage pipeline (strip `_` keys ramses_rf doesn't need, map `_bound`→`bound` etc.)
- `SCH_TRAITS_HVAC` `bound` accepts `str | list[str]` (multi-REM binding)

Both the HA integration (ramses_cc) and the CLI (ramses_cli) call the same ramses_rf function — no duplicate stripper. The CLI loads `config.json` directly into ramses_rf, so without this function `_`-prefixed keys like `_commands` would reject the entire schema (`PREVENT_EXTRA`).

### Changes

| File | What |
|------|------|
| const.py | `SZ_TR_COMMANDS = "_commands"` constant |
| coordinator.py | `_sync_remotes_to_schema`: copies remotes → schema `_commands` (after `sync_learned_topology`) |
| coordinator.py | `_async_update_schema_commands`: writes `_commands` to config entry (called by remote.py) |
| coordinator.py | Startup: merge schema `_commands` into `_remotes` (highest precedence) |
| coordinator.py | Backup before remotes migration (`reason="ssot_phase3a"`) |
| coordinator.py | `_derive_known_list_from_schema` uses ramses_rf's `strip_and_map_traits()` as base |
| remote.py | `learn_command`/`add_command`/`delete_command` now persist to schema |
| remote.py | `extra_state_attributes` exposes `bound_to_fan` (from `fan_handler._fan_bound_to_remote`) |
| climate.py | `async_set_fan_mode` reads from `coordinator._remotes` (schema `_commands`) first, falls back to `known_list[commands]` |
| fan_handler.py | `setup_fan_bound_devices` accepts `str | list[str]` for `_bound`, loops over bound REMs |
| store.py | `STORAGE_VERSION` stays at 1 (no bump), `_async_migrate_func` stays no-op identity — runtime migration via coordinator's `_sync_remotes_to_schema` |
| tests | tests: no-op identity migration, `_sync_remotes_to_schema`, strip `_commands`, `bound_to_fan` attr, `set_fan_mode` reads from `_remotes`, fan_handler list `_bound` |

### Precedence (highest wins)
1. Schema `_commands` (SSOT — user edits, `learn_command` writes here)
2. `.storage[remotes]` (cache — synced to schema on 5-min cycle)
3. `known_list[dev][commands]` (DEPRECATED legacy fallback)

### Config flow
`async_step_schema` already supports editing `_commands` — it strips `_` keys before validation, so no UI change needed.

#### Test plan
- [x] 956 tests pass (`pytest tests/tests_new/`)
- [x] mypy clean (19 source files, 0 errors)
- [x] ruff check + format pass
- [x] ramses_rf PR merged (strip_and_map_traits + list bound): https://github.com/ramses-rf/ramses_rf/pull/820
- [ ] Manual: learn a command, verify it appears in schema `_commands`
- [ ] Manual: reload, verify command persists from schema
- [ ] Manual: verify runtime migration copies remotes → schema _commands on upgrade
- [ ] Manual: verify `bound_to_fan` attribute on REM entity
- [ ] Manual: verify `set_fan_mode` intercepts custom command from schema

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -1079,34 +1079,44 @@ class RamsesHvac(RamsesEntity, ClimateEntity):
             )
 
         try:
-            # 1. Check for user-defined custom commands in the config
+            # 1. Check for user-defined custom commands.
+            # Priority: schema _commands (SSOT, via coordinator._remotes)
+            #          > known_list[bound_rem][commands] (legacy fallback)
             bound_rem = self._bound_rem or self._device.get_bound_rem()
+            commands: dict[str, str] = {}
             if bound_rem:
-                rem_config = self.coordinator.options.get(SZ_KNOWN_LIST, {}).get(
-                    str(bound_rem), {}
+                # Schema _commands (Phase 3a SSOT — populated from
+                # config_entry.options[schema][rem_id][_commands])
+                remotes = getattr(self.coordinator, "_remotes", {}) or {}
+                if isinstance(remotes, dict):
+                    commands = remotes.get(str(bound_rem), {})
+                if not commands:
+                    # Legacy fallback: known_list[bound_rem][commands]
+                    rem_config = self.coordinator.options.get(SZ_KNOWN_LIST, {}).get(
+                        str(bound_rem), {}
+                    )
+                    commands = rem_config.get(CONF_COMMANDS, {})
+
+            if fan_mode in commands:
+                cmd_str = commands[fan_mode]
+                _LOGGER.info(
+                    "Intercepted fan_mode '%s'; sending custom command: %s",
+                    fan_mode,
+                    cmd_str,
                 )
-                commands = rem_config.get(CONF_COMMANDS, {})
 
-                if fan_mode in commands:
-                    cmd_str = commands[fan_mode]
-                    _LOGGER.info(
-                        "Intercepted fan_mode '%s'; sending custom command: %s",
-                        fan_mode,
-                        cmd_str,
-                    )
+                # Users might enter a CLI shorthand OR a raw frame string
+                try:
+                    cmd = Command.from_cli(cmd_str)
+                except (ValueError, RamsesException):
+                    # Fallback for raw packet frames copied from logs
+                    cmd = Command(cmd_str)
 
-                    # Users might enter a CLI shorthand OR a raw frame string
-                    try:
-                        cmd = Command.from_cli(cmd_str)
-                    except (ValueError, RamsesException):
-                        # Fallback for raw packet frames copied from logs
-                        cmd = Command(cmd_str)
-
-                    await self._device._gwy.async_send_cmd(
-                        cmd, num_repeats=2, priority=Priority.HIGH
-                    )
-                    self.async_write_ha_state()
-                    return
+                await self._device._gwy.async_send_cmd(
+                    cmd, num_repeats=2, priority=Priority.HIGH
+                )
+                self.async_write_ha_state()
+                return
 
             # 2. Fallback to standard ramses_rf implementation
             await self._device.set_fan_mode(fan_mode)
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -1028,6 +1028,26 @@ class RamsesHvac(RamsesEntity, ClimateEntity):
         return self._get_cached_fan_info()
 
     @property
+    def fan_modes(self) -> list[str] | None:
+        """Return the list of available fan modes.
+
+        Extends the standard fan modes with custom command names from
+        the bound REM's schema ``_commands`` (Phase 3a SSOT) so that
+        ``climate.set_fan_mode`` accepts custom command names like
+        ``boost`` or ``speed_1``.
+        """
+        base_modes = list(self._attr_fan_modes or [])
+        bound_rem = self._bound_rem or self._device.get_bound_rem()
+        if bound_rem:
+            remotes = getattr(self.coordinator, "_remotes", {}) or {}
+            commands = remotes.get(str(bound_rem), {})
+            if isinstance(commands, dict):
+                for cmd_name in commands:
+                    if cmd_name not in base_modes:
+                        base_modes.append(cmd_name)
+        return base_modes
+
+    @property
     def hvac_action(self) -> HVACAction | str | None:
         """Return the current running hvac operation if supported.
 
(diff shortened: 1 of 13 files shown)


silverailscolo (approved):

wimpie70:
this depends on https://github.com/ramses-rf/ramses_rf/pull/820

Note the downgrade path:

```
0.58.0 has:

- STORAGE_VERSION = 1 (no migration logic — identity no-op)
- _strip_schema_extensions with _strip_traits that recursively strips all _-prefixed keys (including _commands)

What happens on downgrade 0.58.2 → 0.58.0:

1. HA Store: stored data is v2, code expects v1. HA only calls _async_migrate_func when old_major < self.version. Since 2 > 1, it loads data as-is — no migration, no error.
2. remotes in .storage: still present (Phase 3a migration only copies to _commands, never deletes remotes). 0.58.0 reads remotes normally — commands work.
3. _commands in schema: 0.58.0's _strip_traits strips it (recursively removes any key starting with _). ramses_rf never sees it. Harmless.
4. No data loss: learned commands are in both remotes (old format) and _commands (new format). Downgrade reads remotes.

```

And:

```
There are two different strippers:

ramses_cc's _strip_schema_extensions — in coordinator.py, has been there since before 0.58.0. This is what matters for the downgrade path.
ramses_rf's strip_and_map_traits — new in PR 820, goes into ramses_rf 0.58.2. This is the shared function that both ramses_cc and ramses_cli will eventually use.
The downgrade path to 0.58.0 relies on #1 (ramses_cc's own stripper), not #2. Let me verify that 0.58.0 actually has it:

Command cd in ~/…/ramses_extras

- cd /home/willem/dev/ramses_cc && git show 0.58.0:custom_components/ramses_cc/coordinator.py | grep -B2 -A15 "def _strip_traits"
        """

 ```
       def _strip_traits(obj: Any) -> Any:
            """Recursively strip _ prefixed keys from dicts."""
            if isinstance(obj, dict):
                return {
                    k: _strip_traits(v)
                    for k, v in obj.items()
                    if not str(k).startswith("_")
                }
            return obj

        # First pass: collect _disabled/_skipped device IDs so we can remove
        # them from orphan lists, and collect un-disabled trait-only devices
        # so we can add them to orphans (so ramses_rf creates them)
        ctl_id = schema.get(SZ_MAIN_TCS)
        root_owner = schema.get(SZ_OWNER)
        disabled_ids: set[str] = set()
```

Confirmed — ramses_cc 0.58.0 already has _strip_traits that strips all _-prefixed keys (including _commands). This is ramses_cc's own code, nothing to do with ramses_rf PR 820.

So the downgrade path works like this:

- 0.58.2 (Phase 3a): writes _commands to schema, storage migrates v1→v2, remotes kept as cache
- 0.58.0 (downgrade): storage v2 loaded as-is (no migration triggered), _commands stripped by ramses_cc's own _strip_traits, remotes read normally from .storage
PR 820 (strip_and_map_traits in ramses_rf) is irrelevant to the downgrade path. It only matters for the upgrade — moving the stripping logic from ramses_cc to ramses_rf so ramses_cli also benefits.
```

**Comment by wimpie70** [comment]
Had to revert the version bump.

### Correction: the downgrade path does NOT work with a storage version bump.

I was wrong in my previous comment. HA's Store._async_load_data checks:

```python
if data["version"] > self._max_readable_version:
    raise UnsupportedStorageVersionError(...)
```
_max_readable_version defaults to the code's version (1 for 0.58.0/0.58.1). Neither sets max_readable_version, so they cannot read v2 data — the integration would fail to start on downgrade.

Fix applied (commit e44d43d): STORAGE_VERSION stays at 1. The _migrate_v1_to_v2 storage migration is removed — it was redundant anyway. The Phase 3a command migration (remotes → schema _commands) is handled at runtime by the coordinator's _sync_remotes_to_schema, which runs on every save cycle regardless of storage version.

### Corrected downgrade path (0.58.2 → 0.58.0):

Storage stays v1 — no version mismatch, HA loads it normally
remotes in .storage — still present (Phase 3a only copies to _commands, never deletes). 0.58.0 reads remotes normally
_commands in schema — stripped by 0.58.0's own _strip_traits (recursively removes _-prefixed keys). ramses_rf never sees it
No data loss — learned commands exist in both remotes and _commands

wimpie70:
Downgrade path clarification

The STORAGE_VERSION bump to v2 was reverted because HA's Store raises UnsupportedStorageVersionError when the stored version exceeds max_readable_version (defaults to the code's version). Since 0.58.0/0.58.1 have STORAGE_VERSION = 1 and don't set max_readable_version, they cannot read v2 data.

For users who ran the briefly-released v2 code (me), max_readable_version=2 is now set on the store. This allows the current code (v1) to read v2 files: HA's Store calls the no-op migrate function (v2 format is identical to v1) and saves the data back as v1. After the first load, the file is v1 and downgrade to 0.58.0/0.58.1 works normally.

Migration paths:

v1 → v1: no migration (exact match)
v2 → v1: no-op identity, saved back as v1
v1 → 0.58.0/0.58.1: no migration (both v1)

silverailscolo:
@wimpie70 This is also intended for 0.58.2? Enough fixes already to not wait too long. Migration seems not a very big issue, but well developed.

wimpie70:
I think this is complete...i have 1 or 2 fixes coming up

Issue #812, opened 2026-07-13
0.57.12 - Entity states are not restoring from cache


peternash:
When restarting HA with ramses_cc 0.57.12 and Evohome it would appear that the entity states are not being restored from cache.  Entities such as `temperature`, `window_open`, Opentherm sensors etc. all initially come up as `Unknown` and slowly populate over a matter of minutes, presumably as they receive new data.  However, the battery sensors only seem to get updates once every 24 hours so they all go `Unknown` and slowly repopulate over 24 hours.

I suspect all cached entities are affected but it's most noticeable and inconvenient for the battery sensors.

I'm also seeing a lot of these errors in the log on startup:

```
2026-07-13 15:52:40.945 DEBUG (MainThread) [ramses_rf.lifecycle] Gateway: Restoring a cached packet log...
2026-07-13 15:52:40.945 DEBUG (MainThread) [ramses_rf.lifecycle] Gateway: Pausing engine...
2026-07-13 15:52:40.946 DEBUG (MainThread) [ramses_rf.lifecycle] Gateway: Failed to restore packet 2026-07-13T15:38:00.653000+01:00: Bad frame: Invalid structure: >>><<<
2026-07-13 15:52:40.971 DEBUG (MainThread) [ramses_rf.lifecycle] Gateway: Failed to restore packet 2026-07-13T15:38:00.669000+01:00: Bad frame: Invalid structure: >>><<<
2026-07-13 15:52:40.971 DEBUG (MainThread) [ramses_rf.lifecycle] Gateway: Failed to restore packet 2026-07-12T15:54:40.340000+01:00: Bad frame: Invalid structure: >>><<<
2026-07-13 15:52:40.971 DEBUG (MainThread) [ramses_rf.lifecycle] Gateway: Failed to restore packet 2026-07-13T10:34:49.635000+01:00: Bad frame: Invalid structure: >>><<<
2026-07-13 15:52:40.971 DEBUG (MainThread) [ramses_rf.lifecycle] Gateway: Failed to restore packet 2026-07-13T10:34:44.331000+01:00: Bad frame: Invalid structure: >>><<<
```

I've tried clearing the cache with no improvement.

Attached is a startup log showing the `Failed to restore packet` errors:

[home-assistant (1).log](https://github.com/user-attachments/files/29972157/home-assistant.1.log)

silverailscolo:
@peternash This was automatically closed via merge of _rf PR 813, but I would like to hear if it works for you.
[...]

PR #816, opened 2026-07-14
Bump version, dependency to 0.58.0, restore eco assert

silverailscolo:


silverailscolo:
@peternash In ramses_rf 0.58.0, out on Pypi, I included a "magic" fix to create a TCS for a zone, if there is not yet one.

As a result of this - I think - I had to adjust 2 evohome test values, included in this PR. In fact, these exact 2 lines were changed from the new results to None in [this commit](https://github.com/ramses-rf/ramses_cc/commit/bd1e43e284ff400330846632b2cfdcd5766ac171)

When an element suddenly returns None instead of a value, it Is often a real bug. But the other way around, as here, the fix in _rf may have repaired something that came back empty in the test while it shouldn't.

I will wait releasing _cc until you have had a chance to run current ramses_cc master with _rf 0.58.0 set as dependency.

peternash:
Hi @silverailscolo - I'm running ramses_cc master with ramses_rf 0.58.0 as a dependency.  I don't know if anything else has changed but the only difference I'm seeing from yesterday is that the OpenTherm sensors are very reluctant to start populating - stuck at `Unknown`.  Everything resets to `Unknown` after a restart anyway due to the current issue around not restoring the cache but usually the OpenTherm sensors start populating quite quickly.  That's not happening at the moment but I'll leave it running a little longer.  I just noticed that [#812](https://github.com/ramses-rf/ramses_cc/issues/812) is marked as closed but that's still an issue for me.

Update: OpenTherm sensors have started to populate again after about 20 minutes - I think it's just a symptom of [#812](https://github.com/ramses-rf/ramses_cc/issues/812) that everything gets reset to `Unknown` after a restart.

PR #818, opened 2026-07-15
fix(813): improve main_tcs_id fallback for multiple CTL keys

wimpie70:
When main_tcs is not set in the config schema, the fallback logic only worked when there was exactly one 01: key. With multiple 01: keys (CTL + zone sensors), the fallback failed, preventing comment_device_zones from being built for broadcast TRVs.

The fix: when there are multiple CTL keys, prefer the one that has a "zones" dict — zone sensors don't have zones, only the CTL does.

AI used: GLM 5.2 high


wimpie70:
not sure if this will make it to 0.58.0, can be in the follow up as well.

found the issue running tests with ha_sim_test.py on _extras device simulator.

silverailscolo (approved):

Issue #608, opened 2026-04-10
evohome: Faking not working in 0.56.1


iMiMx:
This is not a very helpful post - so far - for that I apologise, I will update it with more information over the weekend.

... I'm just hoping to potentially spare someone else some pain and a complaining wife ;)

Temperature sensor faking, is not working for me in 0.56.1 - but is when downgrading back to 0.55.3

```
Installation method
Home Assistant OS
Core
2026.4.1
Supervisor
2026.03.3
Operating System
17.2
Frontend
20260325.6
```

- Updated to 0.56.1
- Performed  the instructions from the change log:

"To complete the 0.56.1 setup after restart, open HA > Integration > Ramses RF > PacketLog tab and click Submit. Everything should load instantly."

- All zones on the evohome controller, remain on 'waiting for sync'.  I only noticed a few hours later.
- I tried restarting Home Assistant, no change

Downgrading back to 0.55.3 and all is well again.

iMiMx:
Few others in the HA forum seeing the same as I believe I was:

https://community.home-assistant.io/t/honeywell-ch-dhw-via-rf-evohome-sundial-hometronics-chronotherm/151584/5007

https://community.home-assistant.io/t/honeywell-ch-dhw-via-rf-evohome-sundial-hometronics-chronotherm/151584/5011

iMiMx:
Upgraded again - automation runs/completes without errors:

```
alias: Front Room Evohome Faked Sensor
mode: single
triggers:
  - entity_id: sensor.ble_temperature_front_room_evohome
    trigger: state
  - value_template: >-
      {{ this.attributes.last_triggered is none or (as_timestamp(now()) -
      as_timestamp(state_attr('automation.front_room_evohome_faked_sensor',
      'last_triggered'))) > 900 }}
    trigger: template
conditions:
  - condition: not
    conditions:
      - condition: state
        entity_id: climate.front_room
        state: unavailable
      - condition: state
        entity_id: sensor.ble_temperature_front_room_evohome
        state: unavailable
actions:
  - data:
      entity_id: climate.front_room
      temperature: "{{ states(\"sensor.ble_temperature_front_room_evohome\") | float }}"
    action: ramses_cc.fake_zone_temp
  - delay: "00:00:01"
  - data: {}
    target:
      entity_id: sensor.03_055596_temperature
    action: homeassistant.update_entity
  - data: {}
    target:
      entity_id: climate.front_room
    action: homeassistant.update_entity
```

[...]

PR #823, opened 2026-07-15
fix(608): use `await set_temperature()` instead of sync property setter

wimpie70:
The ramses_rf 0.56.1 async refactor changed temperature from a sync property with a setter to an async method. The old code `sensor.temperature = value` silently overwrote the async method with a float, doing nothing — faking appeared to succeed but the temperature never changed.

Changes:
- climate.py: async_fake_zone_temp is now async, calls `await sensor.set_temperature(temperature)`, and also updates the zone's temp_state directly (the zone has its own separate temp_state in ramses_rf's CQRS model)
- sensor.py: async_put_dhw_temp and async_put_room_temp are now async and call `await ...set_temperature(temperature)`
- All three methods call async_write_ha_state() for immediate UI update

see https://github.com/ramses-rf/ramses_cc/issues/608

AI used: GLM-5.2 high

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import json
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, replace as dc_replace
 from datetime import datetime as dt, timedelta as td
 from typing import Any, Final, cast
 
@@ -765,8 +765,7 @@ class RamsesZone(RamsesEntity, ClimateEntity):
 
     # the following are integration-specific method service calls
 
-    @callback
-    def async_fake_zone_temp(self, temperature: float) -> None:
+    async def async_fake_zone_temp(self, temperature: float) -> None:
         """Cast the room temperature of this zone (if faked).
 
         :param temperature: The temperature to fake.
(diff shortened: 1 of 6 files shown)


silverailscolo:
Both Coverage and Testing fail on `File "/home/runner/work/ramses_cc/ramses_cc/tests/tests_new/test_climate.py", line 659, in test_zone_methods_and_services
    zone.async_fake_zone_temp(20.0)`

Will check test_climate, related to the change, probably caused by changing  sync to/from async, as described in the PR

@wimpie70 np probably needs _rf bump, now on Pypi, will rerun

silverailscolo:
@wimpie70 Had a look and ran into:

1. A similar `sensor.temperature = temperature` call is still present in `custom_components/ramses_cc/water_heater.py` method `async_fake_dhw_temp()`. Should be replaced - and awaited - in `tests/tests_new/test_water_heater.py` too?
2. Trouble mocking the data_temp model in the test_zone_methods_and_services under 5. (L662). Got as far as:
```
from ramses_rf.models import TemperatureState
...
mock_device.sensor = AsyncMock()
zone.temp_state = TemperatureState()  # << not a dataclass model, cast?
await zone.async_fake_zone_temp(22.5)
...
```
The await threw: `TypeError: replace() should be called on dataclass instances` from `async_fake_zone_temp`.

Tomorrow, will wait with release until I receive your updates to this PR.

silverailscolo (approved):

Issue #825, opened 2026-07-16
0.58.1 DiscoveryScan adds HVAC REM to TCS


silverailscolo:
Although it was in my Known List with a bound trait to my HRU:
```
"32:022222":
  class: FAN
  bound: "29:091138"
```
upon hearing it after a day and asking me to Accept, the Discovery inserts it in the config schema under TCS:
```
...
"01:088175":
  _class: CTL
  _owner: ours
  zones:
    "00":
      actuators:
        - "04:143181"
  remotes:
    - "29:091138"
...
```
Next is an error:
```
Logger: custom_components.ramses_cc
Bron: custom_components/ramses_cc/__init__.py:231
Integratie: RAMSES RF (documentatie, problemen)
Eerst voorgekomen: 14:32:55 (6 gebeurtenissen)
Laatst gelogd: 14:34:51

Unexpected error during setup of entry 01J8SP7MFT2Q8793JP11BMK46X: Failed to initialise RAMSES client: not a valid value @ data['01:088175']['remotes']
Traceback (most recent call last):
  File "/config/custom_components/ramses_cc/coordinator.py", line 543, in async_setup
    self.client = self._create_client(config_schema)
                  ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/config/custom_components/ramses_cc/coordinator.py", line 1671, in _create_client
    return Gateway(
        port_name=port_name,
        config=gwy_config,
        loop=self.hass.loop,
    )
  File "/usr/local/lib/python3.14/site-packages/ramses_rf/gateway.py", line 140, in __init__
    self._schema: dict[str, Any] = SCH_GLOBAL_SCHEMAS(self._gwy_config.schema or {})
                                   ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/voluptuous/schema_builder.py", line 205, in __call__
    return self._compiled([], data)
           ~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/voluptuous/schema_builder.py", line 574, in validate_dict
    if not schema:
           ^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/voluptuous/schema_builder.py", line 407, in validate_mapping
voluptuous.error.MultipleInvalid: not a valid value @ data['01:088175']['remotes']

[...]

PR #828, opened 2026-07-16
fix(825): don't place remotes[] under a CTL/TCS in discovery schema

wimpie70:
generate_schema_entry fell back to ctl_id (the TCS) when bound_to (the FAN) was unknown, placing remotes[] under a CTL entry. SCH_TCS rejects 'remotes' (PREVENT_EXTRA), so the next setup failed with:
  not a valid value @ data['01:088175']['remotes']

remotes[] is only valid under a FAN/VCS entry. Now place it under bound_to only when it is not a CTL prefix (01:/23:); otherwise orphan the REM/CO2 to orphans_hvac (schema-safe) instead of corrupting the TCS.

Tests updated: the CO2-with-ctl fallback test now expects orphans_hvac, and regression tests cover the issue scenario (REM with ctl_id only) and a non-FAN bound_to.

Issue: https://github.com/ramses-rf/ramses_cc/issues/825

AI used: GLM 5.2 high

### DIFF
--- a/custom_components/ramses_cc/discovery.py
+++ b/custom_components/ramses_cc/discovery.py
@@ -688,10 +688,20 @@ class DiscoveryManager:
         #  in remotes[] for now — the sensors[] list is reserved for the
         #  future when load_fan is implemented and the Builder pattern
         #  can distinguish dual-role devices (CO2 sensor + REM).
+        #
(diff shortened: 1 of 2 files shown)


wimpie70:
Yeah, I'm done for 0.58.2 after this one.
Think most real bugs that were reported are fixed now.

silverailscolo (approved):

PR #831, opened 2026-07-17
Phase 3c: flagging mismatches — persistent notification, entity attrs new mismatch types

wimpie70:
Extends the existing class mismatch detection with:

1. Persistent notification — a HA persistent notification is sent when any mismatch is detected, with per-device details. Auto-dismissed when all mismatches are resolved. Separate notification ID from the "new devices" notification.

2. Diagnostic entity attributes — mismatch flags are surfaced as state attributes (class_mismatch, bound_mismatch, missing_class, orphaned) on every entity whose device has a flag. Visible in Developer Tools without opening the config flow.

3. _bound mismatch detection — compares schema _bound with scan engine's bound_to. Logs warning + sets bound_mismatch metadata.

4. Missing _class detection — flags devices where the scan engine has a likely_type but the schema entry has no _class at all.

5. Orphaned device detection — flags schema devices not seen by the scan engine for longer than the lost threshold (default 7 days). Skips HGI, structural keys, and NEW (never accepted) devices.

6. Unified check_all_mismatches() — called from the coordinator's periodic checkpoint, runs all four checks and sends the notification.

22 new unit tests (6 bound, 5 missing_class, 7 orphaned, 4 unified). 133 tests pass in test_discovery_manager.py. ruff/mypy clean.

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -1060,8 +1060,13 @@ class RamsesHvac(RamsesEntity, ClimateEntity):
                     base_modes.append(cmd_name)
 
         # Phase 3a: bound REM's _commands (packet strings, fallback)
+        # Only offer REM commands if the REM is faked — sending as a
+        # non-faked REM would collide with the physical remote's transmissions.
         bound_rem = self._bound_rem or self._device.get_bound_rem()
         if bound_rem:
+            rem_dev = self.coordinator._get_device(str(bound_rem))
+            if rem_dev is not None and not rem_dev.is_faked:
+                return base_modes  # REM not faked — don't offer its commands
             rem_commands = remotes.get(str(bound_rem), {})
             if isinstance(rem_commands, dict):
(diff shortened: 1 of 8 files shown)


wimpie70:
The four mismatch types added

- class_mismatch: Schema _class vs scan likely_type -> Config flow review + notification + entity attr
- bound_mismatch:Schema _bound vs scan bound_to -> Notification + entity attr
- missing_class: Scan has likely_type but schema has no _class -> Notification + entity attr
- orphaned: In schema but not seen by scan for >7 days -> Notification + entity attr

silverailscolo (approved):

silverailscolo:
@wimpie70 Can I include this in the next release? (_rf is already out with the zone name fix)

wimpie70:
I didn't check it myself yet, only automated tests

wimpie70:
checking https://github.com/ramses-rf/ramses_cc/issues/837

silverailscolo (approved):

PR #833, opened 2026-07-18
Refactor: Migrate HVAC commands to CQRS Intent architecture for `ramses_rf` 0.58.3 compatibility

PWhite-Eng:
### The Problem:
Following the release of `ramses_rf` 0.58.3, the monolithic `Command` class was purged of legacy HVAC factory methods (`get_fan_param`, `set_fan_param`) as part of Phase 1 of the CQRS migration. Because `ramses_cc` explicitly relied on these deleted factory methods to construct packets, the integration and its boundary tests started failing entirely with `AttributeError: type object 'Command' has no attribute 'get_fan_param'` and `set_fan_param`.
### Consequences:
If not fixed, the `ramses_cc` integration will be unable to control or query HVAC ventilation systems. The service calls will throw fatal exceptions locally, and it completely prevents `ramses_cc` from integrating with `ramses_rf` versions >= 0.58.3.
### The Fix:
Transitioned the command construction architecture in `services.py` away from legacy factory methods towards the new CQRS `Intent` building pipeline, and fundamentally refactored the test suite to validate against these architectural changes.
### Technical Implementation:
- **`services.py`**: Replaced direct `Command.get_fan_param` and `Command.set_fan_param` invocations with the new `Intent(..., action=Action.*)` dataclass and routed them through `build_dto(intent)`.
- **Address Casting**: Wrapped raw string IDs in `ramses_rf.address.Address` before passing them to `Intent` to satisfy validation.
- **Test Schema Validation (`tests_new`)**: Updated boundary tests that previously used invalid dummy parameter IDs (e.g., `"0A"`) to use valid IDs (e.g., `"01"`) since `build_dto` enforces `_2411_PARAMS_SCHEMA` strict validation. 
- **Mock Adjustments (`tests_new` & `tests_old`)**: Removed patches to the legacy `Command` class. Mocks now target `build_dto` and assertions have been remapped to validate `CommandDTO` structures (i.e. checking `addr1` and `addr2` instead of `src.id` and `dst.id`).
### Testing Performed:
Ran both the legacy suite (`pytest tests/tests_old`) and the new suite (`pytest tests/tests_new`). Verified that all `1137` tests (including `test_services.py` and `test_coordinator.py`) execute successfully. The tests thoroughly cover device capability checks, fallback bindings, remote overrides, and exception/timeout catching inside the `_async_run_fan_param_sequence` handlers.
Total breakage of HVAC ventilator controls in Home Assistant for users running `ramses_rf` 0.58.3 and above.
### Risks of Implementing:
The introduction of strict schema validation via `build_dto` means that if end-users attempt to send unsupported parameter IDs, the service will now throw a `ValueError` earlier in the pipeline. We might see an increase in strict validation errors if edge-case ventilation systems exist that use undocumented parameter IDs.
### Mitigation Steps:
Exception catching (`try/except`) for schema and service validation exists internally. Tests explicitly verify that generic exceptions and out-of-range ValueErrors are correctly caught, gracefully logged, and converted into valid `HomeAssistantError` states without silently crashing the coordinator.
### AI Assistance Disclosure:
This contribution was developed with the assistance of Google Gemini 3.1 Pro for code generation and documentation. All logic and implementations were reviewed and verified by the author.

### DIFF
--- a/custom_components/ramses_cc/manifest.json
+++ b/custom_components/ramses_cc/manifest.json
@@ -14,9 +14,9 @@
       "aioesphomeapi>=45.3.1",
       "aiousbwatcher>=1.1.2",
       "pyserial-asyncio-fast>=0.16",
-      "ramses-rf==0.58.2",
+      "ramses-rf==0.58.3",
       "serialx>=1.6.0"
     ],
     "single_config_entry": true,
-    "version": "0.58.2"
+    "version": "0.58.3"
   }
--- a/custom_components/ramses_cc/services.py
+++ b/custom_components/ramses_cc/services.py
@@ -13,7 +13,11 @@ from homeassistant.exceptions import HomeAssistantError, ServiceValidationError
 from homeassistant.helpers import device_registry as dr, entity_registry as er
 from homeassistant.helpers.event import async_call_later
 
+from ramses_rf.address import Address
+from ramses_rf.commands.builders import build_dto
+from ramses_rf.commands.core import Command as Intent
 from ramses_rf.devices import Fakeable
+from ramses_rf.enums import Action
 from ramses_rf.exceptions import BindingFlowFailed
 from ramses_rf.protocol.ramses import _2411_PARAMS_SCHEMA as _2411_PARAMS_SCHEMA
 from ramses_rf.schemas import (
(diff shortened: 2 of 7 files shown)


PWhite-Eng:
**Update on CI Failures:**

I've just pushed a follow-up commit to resolve the CI test failures.

The tests were failing with `ModuleNotFoundError: No module named 'ramses_rf.commands.builders'` because the GitHub Actions environment was still pulling `ramses-rf==0.58.2` (as specified in `manifest.json`), which predates the new CQRS `build_dto` logic.

I have bumped the dependency in `manifest.json` and `requirements_dev.txt` to `0.58.3` so the CI uses the correct version with the new builders.

silverailscolo (approved):

Issue #822, opened 2026-07-15
0.58.0+ - Evohome zone names not populating


peternash:
Running master of both ramses_cc and ramses_rf (0.58.0 with the "restore from cache" fix, which seems to work very well [#812](https://github.com/ramses-rf/ramses_cc/issues/812)).  However, I've noticed that now the Evohome zone names are not being populated correctly, they are all defaulting to the form "01:216136_xx".  Prior to the restore from cache fix the Evohome zone names were being used.  I've tried a full cache and schema clear and rebuild but still the same.

Obviously the zone names can be overridden in HA but it's more convenient if the Evohome names are used by default.

Startup logs attached:

[home-assistant (1).log](https://github.com/user-attachments/files/30061318/home-assistant.1.log)
[packet_log.log](https://github.com/user-attachments/files/30061319/packet_log.log)

peternash:
I'm assuming that restoring from cache is now somehow preventing the zone names from being discovered.  

I was also wondering if it makes sense to store the zone names in the discovered schema?

silverailscolo:
> I was also wondering if it makes sense to store the zone names in the discovered schema?

It does, at least to me. But we have an old issue #67 that wants to prevent users editing a Zone's `friendly_name` (supposed to be based on the `zone._name`. We should confirm showing it in the discovered schema doesn't facilitate that.

Note: looking at the current UI, it makes sense to call the Config Schem "discovered schema", as it is not the same as the gateway status sensor Details "schema" contents. The latter is what ramses_rf is actually using.

silverailscolo:
The last run of Discovery reports:
```
[...]

PR #836, opened 2026-07-18
fix: exclude foreign HGIs (18:) from block_list

wimpie70:
Foreign HGIs (18: devices with _owner != root_owner) were being put in the block_list alongside other foreign devices.  This caused ramses_rf's protocol filter to silently drop all packets to/from the foreign HGI, including the controller's 0004 zone name responses that are addressed to the foreign HGI.

The active gateway eavesdrops on these responses to populate zone names and other state.  Blocking the foreign HGI prevented this eavesdropping, so zone._name stayed None and the device registry showed fallback names like "01:216136_xx" (issue 822).

Now 18: devices are exempt from the block_list.  ramses_rf's protocol filter already has foreign-HGI handling (warn but pass through for receiving) — this fix ensures those packets actually reach that logic instead of being dropped by the block_list check.

Requires the companion ramses_rf fix (foreign-HGI exception moved before block_list check in _is_wanted_addrs) for full defense-in-depth.

see https://github.com/ramses-rf/ramses_cc/issues/822

AI used: GLM 5.2 High

### DIFF
--- a/custom_components/ramses_cc/config_flow.py
+++ b/custom_components/ramses_cc/config_flow.py
@@ -1507,20 +1507,27 @@ class RamsesOptionsFlowHandler(BaseRamsesFlow, OptionsFlow):
         config_schema_check = self.options.get(CONF_SCHEMA, {})
         if isinstance(config_schema_check, dict):
             coordinator.discovery_manager.check_class_mismatches(config_schema_check)
+            coordinator.discovery_manager.check_missing_class(config_schema_check)
 
         new_devices = coordinator.discovery_manager.get_devices(
             status=DiscoveryStatus.NEW
         )
         mismatched_devices = coordinator.discovery_manager.get_mismatched_devices()
-        # Deduplicate: a device could be both NEW and mismatched (unlikely
-        # but safe) — only show it once, in the NEW section.
+        missing_class_devices = (
+            coordinator.discovery_manager.get_missing_class_devices()
+        )
+        # Deduplicate: a device could appear in multiple categories — only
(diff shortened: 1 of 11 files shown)


wimpie70:
## Summary

- Exclude 18: devices (HGIs) from the block_list in `_create_client`, even when `_owner != root_owner`
- Sync `_name` from learned_schema to config_schema in `sync_learned_topology` (was syncing sensor/actuators/class but not _name)
- Add tests: `test_create_client_foreign_hgi_not_in_block_list`, `test_sync_learned_topology_adds_zone_name`, `test_sync_learned_topology_preserves_existing_zone_name`

## Root cause of issue 822

Two bugs prevented zone names from appearing:

1. Foreign HGIs with `_owner: not-me` were put in the block_list. The protocol filter dropped all packets to/from the foreign HGI, including 0004 zone name RPs addressed to it. The active gateway could not eavesdrop on these responses.

2. Even when the 0004 packet reached ramses_rf and the zone name was learned, `sync_learned_topology` did not copy `_name` from the learned_schema to the config_schema. The zone merge logic synced `sensor`, `actuators`, and `class` but omitted `_name`.

Requires the companion ramses_rf fix (foreign-HGI exception moved before block_list check in `_is_wanted_addrs`).

#### Test plan
- [x] `pytest tests/tests_new/test_coordinator.py tests/tests_new/test_schemas.py` — 372 passed
- [x] `ruff check` + `ruff format` — clean
- [x] ha-sim recipe 28: 0004 RP to foreign HGI → zone name persisted to schema

silverailscolo (approved):

Issue #841, opened 2026-07-19
0.58.3 - Evohome controller `active_fault` is showing Unknown state


peternash:
When there are no active faults the `active_fault` binary sensor on the Evohome controller returns a state of `unknown`.  On earlier releases this returned a state of `off`.  When a fault is present the sensor correctly returns the `problem` state.

Ideally the state should report a known, non-problem state when there are no faults present.

Ramses RF 0.58.3:

<img width="1088" height="832" alt="Image" src="https://github.com/user-attachments/assets/061bd56c-3b5e-4b50-9130-7a5e757a01a3" />

Ramses RF 0.54.3:

<img width="1088" height="832" alt="Image" src="https://github.com/user-attachments/assets/ec311b94-2054-4146-819d-402af19e0814" />

peternash:
Currently testing master of both ramses_cc and ramses_rf (as of 2026-07-19 20:30) and this issue appears to be resolved.

silverailscolo:
Is in pre-release 0.58.4, just out. Thanks for the detailed feedback.

If you find this version is stable enough to drop the "pre-", I can drop that and people will see it pop ip on HACS.

peternash:
@silverailscolo - 0.58.4 is looking very stable and I've not seen any significant issues so far, definitely worth dropping the "pre-" as it fixes some issues in 0.58.3. I'll continue testing but will aim to update my primary HA system later today so I can check out actions and automations etc.  Many thanks to everyone who's contributing to the project code.

peternash:
Fixed in 0.58.4

PR #842, opened 2026-07-19
fix: active_fault sensor reports `off` instead of `unknown` when no faults (issue 841)

wimpie70:
RamsesLogbookBinarySensor.is_on kept _last_known_state at its initial None when ramses_rf's active_faults returned None (empty/unloaded fault log), so the sensor showed 'unknown' instead of 'off'. This was a regression introduced by the _last_known_state refactor (commit 51eaddf) — 0.54.3 returned bool(active_faults), which mapped None to False.

Coerce faults to bool unconditionally so None/() both report off, matching the 0.54.3 truth table and the BinarySensorDeviceClass.PROBLEM semantic.

see https://github.com/ramses-rf/ramses_cc/issues/841
AI used: GLM 5.2 high

### DIFF
--- a/custom_components/ramses_cc/binary_sensor.py
+++ b/custom_components/ramses_cc/binary_sensor.py
@@ -224,12 +224,15 @@ class RamsesLogbookBinarySensor(RamsesBinarySensor):
     def is_on(self) -> bool | None:
         """Return the state of the binary sensor.
 
-        :return: True if faults are active, None if unknown.
+        ``active_faults`` returns ``None`` when the fault log is empty or not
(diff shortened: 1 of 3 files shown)


silverailscolo (approved):

silverailscolo:
@peternash both ramses masters contain fixes for the errs you reported. Can't release atm, but you will know how to get them running

peternash:
@silverailscolo - already on it and all looking good so far.  I'll leave it running overnight to get better data on the battery state cache issue and update the relevant open issues tomorrow.

PR #853, opened 2026-07-20
Refactor: Migrate to CommandDTO and stabilise test suite for ramses_rf#877 compatibility

PWhite-Eng:
### The Problem:
The upstream `ramses_rf` library introduced a major architectural change in ramses-rf/ramses_rf#877, deprecating the legacy `Command` interface in favour of the new CQRS `CommandDTO` intents. This caused massive breakages across `ramses_cc`, including failures in message dispatching, entity state initialization, and widespread `pytest` failures.
### Consequences:
If this isn't fixed, `ramses_cc` will be completely incompatible with the upcoming `ramses_rf` release. The integration will crash on startup or when attempting to transmit commands, leading to total system failure for end-users.
### The Fix:
Migrated the entire `ramses_cc` codebase to utilise `CommandDTO` as the sole interface for transmission instructions, completely removing all usages of the legacy `Command` object.
### Technical Implementation:
- Standardised imports across all modules to `from ramses_tx.dtos import CommandDTO` (removing all `as Command` aliases).
- Refactored `async_send_cmd` calls in `climate.py`, `water_heater.py`, and `remote.py` to construct and dispatch `CommandDTO` objects.
- Updated all mock targets in the `tests_new` suite to patch `CommandDTO` instead of `Command`.
- Corrected legacy assertions in `tests_old/test_evo_control.py` to accommodate the improved packet parsing of DHW states introduced in the `ramses_rf` update.
### Testing Performed:
Ran the full `ramses_cc` `pytest` suite (1,209 tests) locally against a linked installation of the updated `ramses_rf` codebase (containing ramses-rf/ramses_rf#877). All tests passed flawlessly. 
*Note: This PR is linked to ramses-rf/ramses_rf#877 and will require a version bump in both repositories. Consequently, the GitHub Actions CI tests for this PR are expected to fail until the `ramses_cc` requirements are bumped to point to the new `ramses_rf` release.*
### Risks of NOT Implementing:
The integration will remain fundamentally broken and incompatible with the next version of the core `ramses_rf` library.
### Risks of Implementing:
Possible edge cases in how `CommandDTO` serialises complex edge-case commands compared to the legacy `Command` object, though the extensive test suite significantly minimises this risk.
### Mitigation Steps:
Verified the exact byte-for-byte serialisation in the `tests_new` suite and ensured 100% test success rate against the new upstream architecture to validate payload parity.
### AI Assistance Disclosure:
[...]

Issue #834, opened 2026-07-18
0.58.3 - Evohome schema discovery incorrectly categorising a BDR as a hot water control


peternash:
I've just updated a system with Evohome from Ramses RF 0.56.0 to 0.58.3 and tested the schema discovery by clearing the cache and schema.

The Evohome system has a BDR as a boiler control but no DHW control.  The schema discovery incorrectly assigned the BDR as a `hotwater_valve` and created a DHW device in HA.

I'm not sure if this is a limitation of automatic discovery or if it's possible to identify the BDR as an `appliance_control` automatically.

Original:
```
  system:
    appliance_control: "13:121025"
```

Discovered:
```
  system:
    appliance_control: null
  stored_hotwater:
    hotwater_valve: "13:121025"
```

Original schema:

[ramses_rf_schema.txt](https://github.com/user-attachments/files/30152265/ramses_rf_schema.txt)

Discovered schema:

[discovered_schema.txt](https://github.com/user-attachments/files/30152272/discovered_schema.txt)

Logs:

[home-assistant_ramses_cc_2026-07-18T14-28-04.648Z.log](https://github.com/user-attachments/files/30152276/home-assistant_ramses_cc_2026-07-18T14-28-04.648Z.log)
[packet_log.log](https://github.com/user-attachments/files/30152275/packet_log.log)

wimpie70:
We found some fixes, just testing it with ha_sim_test.py now...
ramses_rf (discovery_scan.py)
- Added _APPLIANCE_CONTROL_CODES = {"3B00", "3EF0"} constant — the TPI loop codes a BDR/OTB broadcasts as I when acting as the boiler relay.
- Added domain_id: str | None field to DiscoveredDevice (FC = appliance_control).
- Added _is_appliance_control_signal() helper that returns True only when a 13:/10: device broadcasts (verb I) 3B00/3EF0 as the src — not for RP/RQ replies (e.g. 3EF1), and not for other device types.
- Wired domain_id through _process_packet → _process_device in all three paths (known-device creation, known-device update, new-device creation, existing-device update), setting confidence="high" when detected.

ramses_cc (discovery.py)
- Added domain_id parameter to generate_schema_entry().
- Added an appliance_control branch in the BDR section: when ctl_id is set, zone_idx is None, and domain_id == "FC", the BDR is placed under system.appliance_control instead of dhw_system.dhw_valve. Zone binding still takes priority over the FC domain.
[...]

PR #856, opened 2026-07-21
sync_learned_topology clears hotwater_valve after re-parenting (issue 834)

wimpie70:
When a BDR is re-parented from hotwater_valve to appliance_control,
the learned schema has stored_hotwater: {} (empty dict).  The step 1c
code checked `if learned_dhw:` which is falsy for empty dicts, so it
never entered the block to clear hotwater_valve from the config.
 
Changed the condition to `if isinstance(learned_dhw, dict):` and added
logic to set valve keys to None when the learned schema has them
absent/None.

merge on top of #853
see issue https://github.com/ramses-rf/ramses_cc/issues/834

AI used: GLM 5.2 high

### DIFF
--- a/.github/workflows/check-cov.yml
+++ b/.github/workflows/check-cov.yml
@@ -69,7 +69,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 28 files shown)


wimpie70:
I'll check

silverailscolo (commented):
You fixed the evohome test differently than the fix now in master. But it is already fixed, so please take that file out

silverailscolo (approved):
Can't edit the fix part, and I see some DHQ item in the same test is new and should go in.
Earlier comment still stands. I didn't understand what you meant with "will check".

Done, via conflict resolving

PR #860, opened 2026-07-22
fix: prevent discovery loop when two relays contend for single-slot r…

wimpie70:
…oles

Issue 834 comment 5044906835: when both an OTB and a BDR broadcast 3B00/3EF0, the scan engine flags both as appliance_control (FC domain). Accepting one device into the appliance_control slot displaces the other, which then gets re-discovered and re-prompted — an infinite discovery loop.

Add _resolve_single_slot_conflicts() to _apply_schema_entry: before merging a discovery fragment, check if it would overwrite a single-slot role (appliance_control, hotwater_valve, heating_valve) already held by a *different* device.  If so, strip the conflicting key and redirect the new device to orphans_heat instead.

Zone/DHW sensor slots are intentionally excluded — replacing a sensor is a legitimate user action, not a misclassification loop.

Tests: 3 new unit tests (appliance_control displacement, hotwater_valve displacement, idempotent re-accept of same device).

see https://github.com/ramses-rf/ramses_cc/issues/834
AI used: GLM 5.2 high

### DIFF
--- a/custom_components/ramses_cc/services.py
+++ b/custom_components/ramses_cc/services.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import copy
 import logging
 import re
 from typing import TYPE_CHECKING, Any, Final, cast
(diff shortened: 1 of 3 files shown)


silverailscolo (approved):
It just isn't any simpler, but we are improving. Thanks

silverailscolo (commented):
You fixed the evohome test differently than the fix now in master. But it is already fixed, so please take that file out

Issue #864, opened 2026-07-23
0.59.1: `send_packet` action unavailable


Deathbyteacup:
**Describe the bug**

Fixed in 0.59.1: 
Calling the `ramses_cc.send_packet` service in 0.59.0 fails immediately with `AttributeError: 'CommandDTO' object has no attribute 'src'`. This appears to affect all `send_packet` calls regardless of payload content, since the failing code path (`_adjust_sentinel_packet`) is called unconditionally on every `CommandDTO` before any value comparison happens. 

This started immediately after upgrading to 0.59.0; a previously working "keep alive" automation that sends faked zone-temperature packets to a BDR91 relay now fails every time it runs.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Developer tools > Actions
2. Select the `ramses_cc.send_packet` action
3. Switch to YAML mode and run:

```yaml
action: ramses_cc.send_packet
data:
  device_id: "03:004303"
  from_id: "03:004303"
  verb: I
  code: 30C9
  payload: "000834"
```

4. See the error in Settings > System > Logs

**Expected behaviour**
The service call should build and transmit the packet as it did prior to 0.59.0, without raising an exception.

```
File "/config/custom_components/ramses_cc/init.py", line 435, in async_send_packet
await _coordinator.async_send_packet(call)
File "/config/custom_components/ramses_cc/coordinator.py", line 2473, in async_send_packet
await self.service_handler.async_send_packet(call)
File "/config/custom_components/ramses_cc/services.py", line 372, in async_send_packet
self._adjust_sentinel_packet(cmd)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
File "/config/custom_components/ramses_cc/services.py", line 397, in _adjust_sentinel_packet
if cmd.src.id != "18:000730" or cmd.dst.id != hgi.id:
^^^^^^^
AttributeError: 'CommandDTO' object has no attribute 'src'
```

**Screenshots**
N/A

**Additional context**
- Installed version of Ramses RF integration: 0.59.0 (also using `ramses_rf` 0.59.0 as pinned by this release)
- Gateway dongle: nanoCUL clone (CH340/FTDI-based USB-to-RF dongle running evofw3), connected via `/dev/serial/by-id/usb-FTDI_FT232R_USB_UART_A50285BI-if00-port0`
- Gateway interface: USB (serial), not MQTT
[...]

PR #867, opened 2026-07-24
fix(services): migrate _adjust_sentinel_packet to CommandDTO position…

wimpie70:
…al addressing

The CommandDTO migration (PR 853) replaced src/dst (Address objects with .id) with addr1/addr2/addr3 (plain strings), and made the dataclass frozen.  _adjust_sentinel_packet still used cmd.src.id/cmd.dst.id and mutated cmd._addrs in place, causing AttributeError on every send_packet call.

Fix: use cmd.addr1/cmd.addr2 for the sentinel check, and dataclasses.replace() to create a new CommandDTO with swapped addr2/addr3 when validation fails.  _adjust_sentinel_packet now returns the (possibly modified) CommandDTO instead of mutating in place.

Tests updated to use real CommandDTO objects instead of MagicMock with the old src/dst/_addrs API.

see 
https://github.com/ramses-rf/ramses_cc/issues/864
AI used: GLM 5.2 high

### DIFF
--- a/custom_components/ramses_cc/services.py
+++ b/custom_components/ramses_cc/services.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import copy
+import dataclasses
 import logging
 import re
 from typing import TYPE_CHECKING, Any, Final, cast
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):

maartenv86:
@silverailscolo updated to 0.59.1 but still receiving:

```
QosProtocol(IsInIdle, len(queue)=0): Send timed out for W|01FF: <ProtocolContext state=WantEcho cmd_=01FF| W|21:057310, tx_count=3/4>: Expired global timer after 20.0 sec
```
And unable to bind/change thermostat values on Itho .

PR #888, opened 2026-08-02
fix: clear `resolve_async_attr` cooldown cache in `force_update`

wimpie70:
## Summary

The `resolve_async_attr` helper has a 30-second cooldown that prevents re-dispatching async getters (e.g. `zone.setpoint`) when the cached value is not None. This means that when a new packet updates the underlying CQRS state (e.g. a 2349 setpoint update), the entity continues to show the stale cached value until the cooldown expires.

The `force_update` service is meant to force an immediate refresh of all entity states, but it only called `async_refresh` which does not clear the cooldown cache.

Add `clear_async_attr_cache()` helper that removes all `_cooldown_` and `_cached_` attributes from an entity, and call it for all entities in `async_force_update` before triggering the refresh.

#### Test plan
- [x] `ruff check` and `mypy` pass
- [x] `pytest tests/` — 1232 passed, 10 skipped (1 pre-existing snapshot failure on master, unrelated)
- [ ] ha_sim_test R36 (climate target_temperature hydration) passes with this fix

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -111,6 +111,7 @@ from .const import (
 )
 from .discovery import DiscoveryManager
 from .fan_handler import RamsesFanHandler
+from .helpers import clear_async_attr_cache
 from .mqtt_bridge import RamsesMqttBridge
 from .schemas import (
     _HEAT_PREFIXES,
@@ -2328,8 +2329,15 @@ class RamsesCoordinator(DataUpdateCoordinator):
     async def async_force_update(self, _: ServiceCall) -> None:
         """Force an immediate update of all device states.
 
+        Clears the resolve_async_attr cooldown cache on all entities so
+        that freshly-received packet data (e.g. a 2349 setpoint update)
+        is visible immediately rather than waiting for the 30-second
+        cooldown to expire.
+
         :param _: Unused service call argument.
         """
+        for entity in self._entities.values():
+            clear_async_attr_cache(entity)
         await self.async_refresh()
 
     async def async_sync_topology(self, _: ServiceCall) -> None:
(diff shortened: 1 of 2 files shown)


wimpie70:
AI used: GLM 5.2 high

wimpie70:
## Additional context for review

> **Note:** The branch history was rewritten (force-pushed) — the previous commits had a messy revert/reapply cycle (fix → revert → reapply → revert = net zero diff). The branch now has a single clean commit on top of the merge base.

### What changed

The `resolve_async_attr` helper has a 30-second cooldown that prevents re-dispatching async getters (e.g. `zone.setpoint`) when the cached value is not None. This means that when a new packet updates the underlying CQRS state (e.g. a 2349 setpoint update), the entity continues to show the stale cached value until the cooldown expires.

The `force_update` service is meant to force an immediate refresh, but it only called `async_refresh` which does not clear the cooldown cache.

### The fix

`clear_async_attr_cache()` does two things:
1. **Cancels in-flight resolution tasks** — tracked via `_resolving_task_` keys. Without this, clearing only the `_resolving_` flag causes a *second* concurrent getter to be dispatched while the first (up to 20s RQ/RP timeout) is still running, doubling outbound command traffic on every `force_update` call. This caused 113% CPU on ha-sim and 75-minute wall times during parallel test runs.
2. **Clears `_cooldown_`, `_cached_`, and `_resolving_` attributes** so the next property access re-dispatches a fresh getter.

### Test results
- R36 (zone climate setpoint hydration): passes 3x consistently after this fix
- Full parallel run: **381 passed, 3 failed** (R34 pre-existing, R31/R47 transient — fixed in ramses_extras PR 125)
- Wall time reduced from ~75 min to ~22 min, CPU from 113% to ~0.05%

silverailscolo (approved):

silverailscolo:
Thanks, quite an improvement, hidden deep down

wimpie70:
yes, it took a full day to find it, had to get clause sonnet 5 medium to do some analyzing as well.

Ok, I still need to run some tests, but I think this one is done

silverailscolo (approved):

PR #889, opened 2026-08-02
bump version, update bug report schema path, snapshot

silverailscolo:
Sets version, _rf dependency.
Expected to fail until we are ready to release, but stored here to remind before actually releasing

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -23,9 +23,9 @@ Provide code if you can. For example, if you are having an issue with a service 
 A clear and concise description of what you expected to happen.
 
 **Please include the following information:**
- - `ramses_cc:` your integration **config** (copy it from HA > Integrations > Ramses RF > Diagnose > YourEspGatewayDevice > Gateway status sensor (click on it) > Three dots menu at top right > Attributes)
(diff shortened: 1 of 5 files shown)


silverailscolo:
@PWhite-Eng With _rf 0.59.2 set as dependency, I had to update the snapshot (returning `None` in 2 cases) to pass the _cc init tests.

Can you confirm that is ok, given the _rf changes? Cf. https://github.com/ramses-rf/ramses_rf/pull/976

PWhite-Eng:
> @PWhite-Eng With _rf 0.59.2 set as dependency, I had to update the snapshot (returning `None` in 2 cases) to pass the _cc init tests.
> 
> Can you confirm that is ok, given the _rf changes? Cf. [ramses-rf/ramses_rf#976](https://github.com/ramses-rf/ramses_rf/pull/976)

Yes, confirmed — returning `None` in those two snapshot cases is expected and correct with `ramses_rf` 0.59.2.

In [ramses-rf/ramses_rf#976](https://github.com/ramses-rf/ramses_rf/pull/976), legacy eavesdropping fallback heuristics (`_evaluate_appliance_eavesdrop_rules` and `_evaluate_zone_type_eavesdrop_rules`) were removed from the topology handlers in favour of pure CQRS state projection. Because those test captures lack explicit domain configuration (`000C`) or binding (`1FC9`) frames for those sensors, `ramses_rf` correctly refrains from speculatively inferring zone sensor associations, resulting in `None`.

Issue #851, opened 2026-07-20
0.58.3: Orcon HRC-400 fan lost 15 entities


wesleymooiman:
Title: FAN loses configuration entities after upgrading from 0.57.12 to 0.58.3

Description

After upgrading from 0.57.12 to 0.58.3, my Orcon HRC-400 FAN device loses a significant number of entities.

0.57.12: ~40 entities
0.58.3: ~25 entities

Around 15 entities disappear, including all configuration/setting entities such as the comfort temperature and other configurable FAN settings.

What I tried
Cleared the Home Assistant entity/device cache.
Allowed auto-discovery to recreate all entities from scratch.
Removed and rebuilt the entity scheme.

None of these steps made any difference. The configuration entities are simply no longer exposed in 0.58.3.

Verification

To rule out a Home Assistant issue, I downgraded back to 0.57.12 (after restoring the previous entity scheme, since the new one is not backwards compatible).

Immediately after downgrading, all missing entities returned and everything worked as expected again.

Hardware
Unit: Orcon HRC-400 (GRU 400)
Integration version: Works on 0.57.12, broken on 0.58.3.

wimpie70:
@wesleymooiman - Thanks for reporting. It seems the 2411 params are broken. I'll find a fix. A new release is expected on/after thursday.

```
2026-07-20 10:42:01.756 DEBUG (MainThread) [custom_components.ramses_cc.services] Processing update_fan_params service call with data: {'device_id': ['32:153289']}
2026-07-20 10:42:01.757 DEBUG (MainThread) [custom_components.ramses_cc.services] Processing get_fan_param service call with data: {'device_id': '32:153289', 'param_id': '01'}
2026-07-20 10:42:01.757 DEBUG (MainThread) [ramses_rf.devices.hvac_ventilators] Found bound REM device 37:168270 for FAN 32:153289
2026-07-20 10:42:01.759 DEBUG (MainThread) [custom_components.ramses_cc.fan_handler] Found entity number.32_153289_param_01 in entity registry
2026-07-20 10:42:01.764 DEBUG (MainThread) [custom_components.ramses_cc.services] Sending command: CommandDTO(verb=<VerbT.RQ: 'RQ'>, addr1='37:168270', addr2='32:153289', addr3='--:------', code=<Code._2411: '2411'>, payload='000001', priority=<Priority.DEFAULT: 0>, num_repeats=0)
[...]

Issue #904, opened 2026-08-07
HA 2026.8: deprecated const `CONCENTRATION_PARTS_PER_MILLION`


silverailscolo:
Full warning:
```
Logger: homeassistant.const
Bron: helpers/deprecation.py:272
Eerst voorgekomen: 6 augustus 2026 om 14:44:43 (3 gebeurtenissen)
Laatst gelogd: 6 augustus 2026 om 20:26:04

The deprecated constant CONCENTRATION_PARTS_PER_MILLION was used from ramses_cc. It will be removed in HA Core 2027.8. Use UnitOfRatio.PARTS_PER_MILLION instead, please report it to the author of the 'ramses_cc' custom integration
```

PR #916, opened 2026-08-08
fix: add periodic 2411 polling and fix param ID normalization

wimpie70:
## Summary

- Add periodic 2411 parameter polling (every 6 hours) in `fan_handler` to replace the daily discovery poll removed from ramses_rf 0.58.3+
- Fix param ID normalization in `_async_param_updated`: ramses_rf strips leading zeros from param IDs (e.g. `"01"` → `"1"`), but the entity description keeps the original format. The comparison `"1" == "01"` failed silently, so parameter update events were ignored and entities stayed unavailable even after values arrived

## Root cause

Two issues prevented FAN parameter `number` entities from receiving values:

1. **No periodic polling**: ramses_rf 0.58.3+ removed the daily `RQ 2411` discovery poll. ramses_cc only requested parameters once at startup via `on_fan_first_message` (which itself never fired — fixed in ramses-rf/ramses_rf#1011). Without periodic polling, values go stale after the initial sweep.

2. **Param ID mismatch**: `_handle_2411_message` in ramses_rf normalizes param IDs with `lstrip("0")` (e.g. `"01"` → `"1"`), but `RamsesNumberParam` stores `ramses_rf_attr="01"` in the entity description. The event handler compared them directly: `"1" != "01"` → event ignored → entity stayed unavailable.

See #851.

Depends on ramses-rf/ramses_rf#1011 for the initialized callback fix.

#### Test plan

- [x] ramses_cc test suite: 1232 passed, 10 skipped (1 pre-existing snapshot failure unrelated to this change)
- [x] Verified on live hass container: all 20 parameter entities receive values after restart (e.g. `Parameter 75 updated: 15.2 (stored as: 15.2)`)


wimpie70:
note, the polling is every 6 hours now...this could be adapted if you want...

custom_components/ramses_cc/sensor.py
```diff
@@ -16,10 +16,10 @@
     SensorStateClass,
 )
 from homeassistant.const import (
+    CONCENTRATION_PARTS_PER_MILLION,
```
silverailscolo, line 19: This is a regression, restores a deprecated unit. Please remove and leave as in master since #904

custom_components/ramses_cc/sensor.py
```diff
@@ -214,7 +214,7 @@ def async_put_co2_level(self, co2_level: int) -> None:
 
         # TODO: Remove from here...
         assert self.device_class == SensorDeviceClass.CO2
-        assert self.native_unit_of_measurement == UnitOfRatio.PARTS_PER_MILLION
+        assert self.native_unit_of_measurement == CONCENTRATION_PARTS_PER_MILLION
```
silverailscolo, line 217: Same: This is a regression, restores a deprecated unit. Please remove and leve as in master

custom_components/ramses_cc/sensor.py
```diff
@@ -507,7 +507,7 @@ class RamsesSensorEntityDescription(RamsesEntityDescription, SensorEntityDescrip
         key=SZ_CO2_LEVEL,
         ramses_rf_attr=SZ_CO2_LEVEL,
         device_class=SensorDeviceClass.CO2,
-        native_unit_of_measurement=UnitOfRatio.PARTS_PER_MILLION,
+        native_unit_of_measurement=CONCENTRATION_PARTS_PER_MILLION,
```
silverailscolo, line 510: And this one too: This is a regression, restores a deprecated unit. Please remove and leve as in master

We must have passed each other here np

Rebase and this is probably solved

silverailscolo (changes requested):

custom_components/ramses_cc/sensor.py
wimpie70: Already fixed in commit a350c80 — replaced UnitOfRatio with CONCENTRATION_PARTS_PER_MILLION. The branch has been rebased on upstream/master which includes the PR 904 changes. These lines now use CONCENTRATION_PARTS_PER_MILLION as expected.

custom_components/ramses_cc/sensor.py
wimpie70: Same — already fixed in a350c80. Line 217 now asserts against CONCENTRATION_PARTS_PER_MILLION.

custom_components/ramses_cc/sensor.py
wimpie70: Same — already fixed in a350c80. Line 510 now uses CONCENTRATION_PARTS_PER_MILLION. The rebase on upstream/master resolved this as you predicted.

wimpie70:
This is confusing...
UnitOfRatio does NOT exist in the current HA version. CONCENTRATION_PARTS_PER_MILLION does.

wimpie70:
I'll be afk till tomorrow...

silverailscolo:
> This is confusing... UnitOfRatio does NOT exist in the current HA version. CONCENTRATION_PARTS_PER_MILLION does.

It is in 2026.8.0 Core and has been since 2 months (Blame). It starts to warn in the system log since. I want to silence this now, and not wait till ? months.

silverailscolo (approved):
Will restore new ppm

PR #918, opened 2026-08-08
fix: restore latest ppm, bump ha_cc dep

silverailscolo:
Repairs #904 regression introduced in #916

### DIFF
--- a/custom_components/ramses_cc/sensor.py
+++ b/custom_components/ramses_cc/sensor.py
@@ -16,10 +16,10 @@ from homeassistant.components.sensor import (
     SensorStateClass,
 )
 from homeassistant.const import (
-    CONCENTRATION_PARTS_PER_MILLION,
     PERCENTAGE,
     EntityCategory,
     UnitOfPressure,
+    UnitOfRatio,
     UnitOfTemperature,
     UnitOfTime,
 )
@@ -214,7 +214,7 @@ class RamsesSensor(RamsesEntity, SensorEntity):
 
         # TODO: Remove from here...
         assert self.device_class == SensorDeviceClass.CO2
-        assert self.native_unit_of_measurement == CONCENTRATION_PARTS_PER_MILLION
+        assert self.native_unit_of_measurement == UnitOfRatio.PARTS_PER_MILLION
 
         device = self._device
         if not isinstance(device, HvacCarbonDioxideSensor):
@@ -507,7 +507,7 @@ SENSOR_DESCRIPTIONS: tuple[RamsesSensorEntityDescription, ...] = (
         key=SZ_CO2_LEVEL,
         ramses_rf_attr=SZ_CO2_LEVEL,
         device_class=SensorDeviceClass.CO2,
-        native_unit_of_measurement=CONCENTRATION_PARTS_PER_MILLION,
+        native_unit_of_measurement=UnitOfRatio.PARTS_PER_MILLION,
         entity_category=None,
     ),
     RamsesSensorEntityDescription(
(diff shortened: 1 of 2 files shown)


wimpie70:
Upstream Master Test Results
Our branch	Upstream master
ramses_cc	fix/fan-param-polling-851	49a119d (PR 918 merged)
ramses_rf	fix/fan-initialized-callback-851	ce27af41 (PR 1011 merged)
Pass	400	328
Fail	2	54
Root Cause: PR 918 broke the sensor platform
PR 918 ("fix ppm, bump ha_cc dep") reverted CONCENTRATION_PARTS_PER_MILLION back to UnitOfRatio.PARTS_PER_MILLION in sensor.py. But UnitOfRatio doesn't exist in the HA version running in ha-sim:



ImportError: cannot import name 'UnitOfRatio' from 'homeassistant.const'
This causes the sensor platform to fail to load, which cascades into 54 failures across many recipes:

Entity creation failures (sensor platform broken → coordinator degraded):

R01: CTL/TRV/DHW entities not created
R30: FAN climate entity missing
R32: TRV battery sensor has no state
R35/R36: water_heater/climate state unavailable
R40: climate entity for zone 03 missing
R45: entities not reappearing after crash
Service call failures (coordinator degraded):

R02/R04: remove_device returns HTTP 500
R10: invalid main_tcs not sanitised
Schema/binding failures (coordinator can't process):

R14: zones wiped after 000C inject
R23: 0004 packets not processed
R24: class_mismatch attribute missing
R26: missing _class detection not logged
R27: FAN remotes/commands not preserved after accept
R34/R37: BDR appliance_control/hotwater_valve binding fails
R44: hvac_schema not populated, _alias not surviving sync
Other issues (may be pre-existing on master, not caused by PR 918):

R21: CTL comment has zone 05 from 000A (000A zone binding leak)
R33: trait-only HVAC not moved to orphans_hvac
R16: ERROR logs during stress test
Key finding
PR 918 is a regression — it depends on a newer HA version that has UnitOfRatio, but the ha-sim container (and likely many users) are on an HA version that only has CONCENTRATION_PARTS_PER_MILLION. Our branch correctly uses CONCENTRATION_PARTS_PER_MILLION, which works on both old and new HA versions.

wimpie70:
I'll look into this tomorrow

silverailscolo:
> I'll look into this tomorrow

The pytest_homeassistant_custom_component requirement >= 0.13.325 that **ramses_extras** relies on (dev.txt) is from April 25, 2026.
I suggest updating that requirement in ramses_extras to >= 0.13.342 from June 28, 2026 to fix this issue.
Master of ramses_cc requires at least 0.13.342, because any older version has this error:
```
ERROR: Cannot install -r requirements/requirements_dev.txt (line 29) and mypy==2.3.0 because these package versions have conflicting dependencies.
                                                                                                                                                               
The conflict is caused by:
    pytest-homeassistant-custom-component 0.13.341 depends on ast-serialize==0.3.0
    mypy 2.3.0 depends on ast-serialize<1.0.0 and >=0.6.0
```

Since 0.13.342 the `UnitOfRatio` is in place.

wimpie70:
great! after updating we now get 0 fails on current masters:
Results
Upstream Master Test (HA 2026.8.1)
Container	Pass	Fail	Time	Status
ha-sim	232	0	581s	PASS
ha-sim-2	52	0	729s	PASS
ha-sim-3	65	0	646s	PASS
ha-sim-4	52	0	717s	PASS
Total	401	0		ALL PASS

PR #926, opened 2026-08-09
fix: repeating discovery notifications after reload

### LINKED ISSUE #917: Repeated notifications of the same two new devices having been discovered
see https://github.com/wimpie70/ramses_extras/issues/127

Issue reported by @donaldwpage on _extras that belongs here instead.

wimpie70:
## Summary

Fixes https://github.com/ramses-rf/ramses_cc/issues/917

After accepting devices in the **Review Discovered Devices** config flow, the ACCEPTED/DISCARDED metadata was lost on reload because it was only persisted during the 5-minute checkpoint — `_async_save()` triggered a reload before the checkpoint could run. The new coordinator restored from `.storage/` with no metadata for the accepted devices, so `check_for_new_devices` flagged them as NEW and re-notified every cycle.

### Root cause

1. `review_discovered` sets `meta.status = ACCEPTED` in the running coordinator's in-memory `_metadata`
2. `_async_save()` → `async_create_entry()` → update listener → full integration reload
3. New coordinator's `restore_state()` reads from `.storage/` — but the ACCEPTED metadata was never persisted (only happens during the 5-minute `async_save_client_state` checkpoint)
4. 10 seconds after reload, `_async_discovery_checkpoint` fires:
   - `sync_with_schema` creates NEW metadata for the scan device (no metadata exists)
   - `check_for_new_devices` sees `meta.status == NEW` and `device_id not in _notified` (empty after reload) → notification sent
5. Repeat every 5 minutes

### The fix — two layers

**1. Persist before reload** (`config_flow.py`): call `async_save_client_state()` before `_async_save()` in `review_discovered` so the ACCEPTED/DISCARDED status survives the reload. This fixes the root cause for the config flow path.

**2. Schema-membership guard** (`discovery.py`): safety net for any metadata loss scenario (HA restart before checkpoint, other reload triggers, etc.):
- `sync_with_schema` stashes `schema_device_ids` as `self._schema_device_ids`
- `sync_with_schema` skips creating NEW metadata for scan devices that are already in the schema
- `check_for_new_devices` skips devices that are in the schema but have no metadata — they're already configured, not new discoveries
- The config flow's direct call to `check_for_new_devices` now calls `sync_with_schema` first, so the stash is always populated

The guard is based on schema membership (an objective fact), not on fabricating ACCEPTED metadata — `ACCEPTED` keeps its meaning of "user explicitly reviewed and accepted."

### Changes

- **`config_flow.py`**: `await coordinator.async_save_client_state()` before `_async_save()` in `review_discovered`; call `sync_with_schema` before `check_for_new_devices` when opening the review form
- **`discovery.py`**: stash `_schema_device_ids` in `sync_with_schema`; skip schema devices in `check_for_new_devices` and in `sync_with_schema`'s second loop
- **`test_discovery_manager.py`**: 3 regression tests

#### Test plan
- [x] `pytest tests/tests_new/test_discovery_manager.py` — 171 passed
- [x] `pytest tests/` — 1248 passed, 10 skipped
- [x] ruff + mypy --strict + pre-commit hooks pass


wimpie70:
if you want I can run a full ha_sim_test tomorrow morning, that is, if you approve the PR's and merge to master.
But so far the individual parts passed.

silverailscolo (approved):

silverailscolo:
@wimpie70  Commits approved, but another merge conflict flagged. If you can review and fix, ready to merge either in 15 mins or early tomorrow

Issue #937, opened 2026-08-11
0.59.4 - Startup blocked by `fan_handler poll_loop`


silverailscolo:
Rebooting HA end in a 'waiting to complete' screen. Finally, the system log shows:

`
Something is blocking Home Assistant from wrapping up the start up phase. We're going to continue anyway.
[...]
The system is waiting for tasks: {<Task pending name='Task-2787' coro=<RamsesFanHandler._start_param_polling.<locals>._poll_loop() running at /config/custom_components/ramses_cc/fan_handler.py:332> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[set.remove()]>, <Task pending name='Task-1672' coro=<RamsesFanHandler._start_param_polling.<locals>._poll_loop() running at /config/custom_components/ramses_cc/fan_handler.py:332> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[set.remove()]>}
`

wimpie70:
First ~Thirst~ (heat wave!) thought: It's waiting for every param, but your FAN only answers on a few... I'll check in a bit...

--- hahaha...I woke up too early today...staring at that sentence...something is off....thanks :D

silverailscolo:
This blocking is new in 0.59.4

silverailscolo:
[Startup-0.59.4.log.txt](https://github.com/user-attachments/files/30942445/Startup-0.59.4.log.txt)

wimpie70:
yes, these logs confirm the bug we fixed:
- _poll_loop() at fan_handler.py:332 — the while True + asyncio.sleep loop
- Two pending tasks blocking startup (two FAN devices, each with its own poll loop)

silverailscolo:
Issue confirmed fixed in current _cc master/upcoming 0.59.5.

PR #938, opened 2026-08-11
fix(fan_handler): use background task for poll loop to not block startup (issue 937)

wimpie70:
## Summary

The periodic 2411 parameter poll loop introduced in 0.59.4 (commit `a96103f`, issue 851) used a `while True` + `asyncio.sleep` loop created via `hass.async_create_task()`. This blocked HA's startup phase because `async_create_task` creates a **tracked** task that HA waits for, but the loop never completes — HA hangs at "waiting for tasks" and eventually gives up with a warning.

Replace the manual loop with **`async_track_time_interval`**, which is the HA-idiomatic pattern for periodic callbacks (already used 3x in `coordinator.py` for discovery, state-save, and checkpoint timers). The timer callback is fire-and-forget — it never blocks startup, and the cancel callable is registered via `entry.async_on_unload()` for automatic cleanup on config entry unload.

This also removes the now-unnecessary `_fan_param_poll_tasks` dict and `_stop_param_polling` method (cleanup is automatic).

See: https://github.com/ramses-rf/ramses_cc/issues/937

#### Test plan

- [x] `prek run -a` passes (ruff, codespell, mypy --strict, datetime aliases)
- [x] `pytest .` — 1267 passed, 10 skipped (1 pre-existing failure in `test_evo_control::test_namespace` deselected, unrelated to this change)
- [x] All 21 fan handler tests pass (updated 2 tests to clean up timers via `async_on_unload` mock)

### DIFF
--- a/custom_components/ramses_cc/fan_handler.py
+++ b/custom_components/ramses_cc/fan_handler.py
@@ -2,14 +2,15 @@
 
 from __future__ import annotations
 
-import asyncio
 import logging
 from collections.abc import Callable
+from datetime import timedelta as td
 from typing import TYPE_CHECKING, Any
 
 from homeassistant.const import Platform
 from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers.dispatcher import async_dispatcher_send
+from homeassistant.helpers.event import async_track_time_interval
 
 from ramses_rf.const import DevType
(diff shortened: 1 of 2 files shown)


wimpie70:
changing  to HA-idomatic way instead...

silverailscolo (approved):

PR #958, opened 2026-08-14
Fix HCC100 cooling mode fallback

ac-uy:
## Summary
- Preserve the configured system heating/cooling default while the HCC100 controller has no active 2D49 mode.
- Continue to report an active cooling override.
- Companion integration behavior for ramses-rf/ramses_rf#1054.

## Validation
- `python -m compileall -q custom_components/ramses_cc/climate.py tests/tests_new/test_climate.py` passed.
- `python -m pytest tests/tests_new/test_climate.py -q` could not collect because the current Python environment does not include `homeassistant` (`ModuleNotFoundError`).


silverailscolo (changes requested):
Your base is outdated. For a correctly set up developer environment you should not need to adjust _all_ import paths.

I will now close this PR. You should not delete this dev branch, but use it as a pasteboard.

Then, create a fresh new branch from _cc master and apply _only_ the additions that apply to the cooling option.

PR #965, opened 2026-08-15
feat(climate): add HCC100 cooling support and UFC pump relay sensor (ramses-rf/ramses_rf#933)

PWhite-Eng:
### The Problem:
Downstream integration in Home Assistant lacked exposure for the new thermal cooling capabilities, HCC100 cooling demand, and UFC actuator pump relay states introduced in ramses-rf/ramses_rf#1066 (ramses-rf/ramses_rf#933).

### The Fix:
1. **Climate Entity Cooling Support**:
   - Mapped `HVACMode.COOL` to target system and zone modes (`SystemMode.AUTO` / `ZoneMode.PERMANENT`).
   - Updated `RamsesController` and `RamsesZone` `hvac_action` to return `HVACAction.COOLING` when `thermal_demand.mode == ThermalMode.COOL` and demand > 0 (and `HVACAction.IDLE` when 0).
   - Updated `RamsesController` and `RamsesZone` `hvac_mode` to check `thermal_mode` and return `HVACMode.COOL`.
   - Updated `RamsesZone.async_set_hvac_mode` to accept `HVACMode.COOL`.
2. **UFC Pump Relay Sensor**:
   - Registered `RamsesSensorEntityDescription` for `UfhController` `SZ_PUMP_RELAY_STATE` exposing diagnostic enum states (`heating`, `cooling`, `off`).

### Testing Performed:
- Added comprehensive unit tests in `tests/tests_new/test_climate.py` and `tests/tests_new/test_sensor.py`.
- Verified all pre-commit checks (`prek run -a`), ruff formatting/linting, strict docstrings, mypy typing, and full test suite (`pytest -n auto`: 1,305 passed, 10 skipped).

### AI Assistance Disclosure:
This contribution was developed with the assistance of Google Gemini 3.7 Flash for code generation and documentation. All logic and implementations were reviewed and verified by the author.

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -44,6 +44,7 @@ from homeassistant.util import dt as dt_util
 from custom_components.ramses_cc.helpers import parse_packet_string
 from ramses_rf.const import SZ_MODE, SZ_SETPOINT, SZ_SYSTEM_MODE
 from ramses_rf.devices import HvacVentilator
+from ramses_rf.enums import ThermalMode
 from ramses_rf.systems.tcs import Evohome
 from ramses_rf.systems.zones import Zone
 from ramses_tx.const import Priority
@@ -92,6 +93,7 @@ MODE_HA_TO_TCS: Final[dict[HVACMode, str]] = {
     HVACMode.HEAT: SystemMode.AUTO,
     HVACMode.OFF: SystemMode.HEAT_OFF,
     HVACMode.AUTO: SystemMode.RESET,  # not all systems support this
+    HVACMode.COOL: SystemMode.AUTO,
 }
 
 PRESET_TCS_TO_HA: Final[dict[str, str]] = {
@@ -121,6 +123,7 @@ MODE_ZONE_TO_HA: Final[dict[str, HVACMode]] = {
 MODE_HA_TO_ZONE: Final[dict[HVACMode, str]] = {
     HVACMode.HEAT: ZoneMode.PERMANENT,
     HVACMode.AUTO: ZoneMode.SCHEDULE,
+    HVACMode.COOL: ZoneMode.PERMANENT,
 }
 
 PRESET_ZONE_TO_HA: Final[dict[str, str]] = {
@@ -199,7 +202,7 @@ class RamsesController(RamsesEntity, ClimateEntity):
         self._last_known_targ_temp: float | None = None
 
     async def async_added_to_hass(self) -> None:
-        """Called when entity is added to Home Assistant."""
+        """Handle entity addition to Home Assistant."""
         await super().async_added_to_hass()
         if resolve_async_attr(self, self._device, "system_mode") is None:
             try:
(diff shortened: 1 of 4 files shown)


silverailscolo (approved):

ac-uy:
Thanks for landing this. I had independently investigated the HCC100 cooling flow in #958. One behaviour we observed is that, before an active 2D49 cooling update arrives, the climate entity should preserve the configured/default heating-or-cooling state rather than fall back to HEAT. Is that intentionally handled elsewhere in the new thermal_mode model? If not, I can provide the packet context and a focused follow-up test/PR.

PWhite-Eng:
Defaulting to `HVACMode.HEAT` when `thermal_mode` is `None` is intentional for backwards compatibility. Standard evohome systems never transmit `2D49` packets, so their `thermal_mode` remains `None`. Falling back to `HVACMode.HEAT` ensures existing heating-only systems continue to report the correct mode without regression.

For HCC100 systems, the entity updates to `HVACMode.COOL` once `2D49` packets are ingested.

That said, if you have packet logs showing a distinct sequence or trait that allows `ramses_rf` to recognise cooling-capable hardware at startup, please do feel free to share them. We would be very glad to look through the packet context and would certainly welcome a targeted test case or follow-up PR in `ramses-rf/ramses_rf` to refine the initialisation behaviour.

ac-uy:
There may already be a protocol-faithful route here: ramses_cc persists recent raw packets and passes them to client.start(cached_packets=...) before entity discovery. A cached HCC100 2D49 should therefore hydrate thermal_mode before the first live packet.\n\nCould we add a regression test in ramses_rf that starts a fresh gateway from a cached real HCC100 cooling packet (2D49 ... F6C8FF) and asserts thermal_mode == COOL? If that does not hydrate today, I think the fix belongs in the cached-packet restore path rather than in a hardware-capability heuristic.\n\nA separate persisted last-known-mode field could remain a fallback for unclean shutdowns, but raw packet replay would be preferable as the source of truth.

PWhite-Eng:
@ ac-uy, Your analysis is corret. Packet cache replay during `client.start(cached_packets=...)` is the intended protocol-faithful mechanism to hydrate read-models before live RF traffic begins.

We ran an empirical reproduction test against `ramses_rf` with a cached `2D49` packet. While `_restore_cached_packets()` replays `2D49` (and does not discard it as high-volume), state projection currently updates the `Controller` device rather than targeting the `Tcs` system read-model (`tcs.system_state.cooling_mode` remains `None`).

The fix belongs in `ramses-rf/ramses_rf`:
1. Include `Code._2D49` in `state_projector.py` / `ingestion.py` system-level target routing so `tcs` receives the state update.
2. Add the cached packet replay regression test in `tests/tests_rf/test_hcc100_cooling.py`.

PWhite-Eng:
We have implemented this bug fix in ramses-rf/ramses_rf#1078.

The PR registers `Code._2D49` with the `StateProjector` system-level routing so cached cooling packets directly hydrate `tcs.system_state.cooling_mode`, and promotes `thermal_mode()` to `ZoneBase` so all zones inherit the parent TCS thermal mode. With this in place in `ramses_rf`, climate entities in `ramses_cc` hydrate correctly directly from `cached_packets` on startup.

Issue #988, opened 2026-08-17
0.59.7 - Orphaned device notification keeps re-appearing after dismissal (no dismiss tracking for orphaned flag)


wimpie70:
## Summary

The "last seen" (orphaned device) notification keeps coming back every 5-minute checkpoint cycle after the user dismisses it. There is no dismiss tracking for the `orphaned` flag, unlike `class_mismatch_dismissed` and `missing_class_dismissed`.

## Root cause

`check_orphaned_devices()` in `discovery.py` (line 877) sets `meta.orphaned` on every checkpoint cycle for devices whose `last_seen` is older than the 7-day threshold:

```python
if last_seen < threshold:
    meta = self._metadata.get(device_id, DeviceMetadata())
    meta.orphaned = f"last seen {last_seen_str} (>{threshold_days} days)"
    self._metadata[device_id] = meta
    orphaned.append(device_id)
```

Then `check_all_mismatches()` (line 1091) counts the orphaned devices and calls `_send_mismatch_notification(counts)` every time `total > 0`. The notification uses a fixed ID (`_mismatch_notification_id`), so HA updates it in-place — but if the user dismissed it, HA removed it, and the next checkpoint re-creates it.

There IS dismiss tracking for:
- `class_mismatch_dismissed` (line 611: `if existing_meta and existing_meta.class_mismatch_dismissed: continue`)
- `missing_class_dismissed` (line 852: `if existing_meta and existing_meta.missing_class_dismissed: continue`)

But there is NO `orphaned_dismissed` flag. The user has no way to dismiss the orphaned notification permanently.

## Logs

```
2026-08-17 09:48:44.427 DEBUG DiscoveryManager: orphaned device 29:176861 — last seen 2026-08-09T16:39:25
2026-08-17 09:48:44.427 INFO  DiscoveryManager: 1 device(s) in schema appear orphaned (not seen > 7 days): 29:176861
2026-08-17 09:48:44.427 DEBUG DiscoveryManager: _send_mismatch_notification counts={'class_mismatch': 0, 'bound_mismatch': 0, 'missing_class': 0, 'orphaned': 1, 'name_mismatch': 0}
```

This repeats every 5 minutes. The notification is re-created each time.

## Fix

[...]

Issue #987, opened 2026-08-17
HGI (18:) devices re-discovered every cycle: _strip_and_orchestrate drops them from schema_device_ids


wimpie70:
## Summary

Foreign HGI devices (e.g. `18:149488`) keep being re-discovered and re-prompted every 5-minute checkpoint cycle, even after the user accepts them as "not-me" and they appear in the schema with `_owner: not-me`.

## Root cause

`_strip_and_orchestrate()` in `schemas.py` (line 353-354) unconditionally drops ALL `18:` entries from the schema:

```python
# Drop HGI (18:) entries — they are gateways, not heating devices.
if isinstance(k, str) and k.startswith("18:"):
    continue
```

This is correct for feeding ramses_rf's Gateway (HGIs are gateways, not controlled devices). However, `_extract_device_ids_from_stripped()` in `coordinator.py` operates on the *already-stripped* schema, so HGI device IDs are never included in `schema_device_ids`.

The checkpoint cycle (`coordinator.py:778-783`) then:
1. Strips the schema → HGI entries removed
2. Extracts device IDs from stripped schema → HGIs missing
3. Calls `sync_with_schema(schema_device_ids)` → HGIs not in set
4. `sync_with_schema` marks HGI metadata as `REMOVED` (line 514-526)
5. `check_for_new_devices` sees the HGI in the scan, metadata is REMOVED → re-notifies

This repeats every 5 minutes.

## Logs

```
2026-08-17 09:48:44.427 INFO DiscoveryManager: device 18:149488 not in schema, marked as REMOVED
2026-08-17 09:53:44.428 INFO DiscoveryManager: device 18:149488 not in schema, marked as REMOVED
2026-08-17 09:58:44.428 INFO DiscoveryManager: device 18:149488 not in schema, marked as REMOVED
```

The schema clearly contains `18:149488: {'_class': 'HGI', '_owner': 'me'}`, but `schema_device_ids` doesn't include it because it was stripped.

## Fix

`_extract_device_ids_from_stripped()` should extract HGI device IDs from the *original* (unstripped) schema, not from the stripped schema. Alternatively, the checkpoint cycle should pass HGI IDs separately or use `_extract_schema_device_ids()` (which operates on the unstripped schema) instead of `_extract_device_ids_from_stripped()`.

[...]

PR #993, opened 2026-08-17
Fix HGI re-discovery (987) and orphaned notification suppression (988)

wimpie70:
## Summary

- **Issue #987 **: HGI (`18:`) devices were re-discovered and re-notified every 5-minute checkpoint cycle because `_strip_and_orchestrate` drops them from the schema (correct for ramses_rf), but `_extract_device_ids_from_stripped` then operated on the already-stripped schema, so HGI IDs were never in `schema_device_ids`. Fixed by using `_extract_schema_device_ids` (unstripped) for discovery sync at all three call sites.
- **Issue #988 **: Orphaned device notifications re-appeared every checkpoint cycle after dismissal. Added a per-device schema key `_suppress_not_seen` that controls suppression:
  - `True`: suppress forever, INFO log every `threshold_days` as a battery reminder
  - `N` (int): suppress for N days from `last_seen`, then auto-remove key and re-notify
  - Default from "Keep (dismiss flag)" in `review_device_health`: 7 days
- The key is auto-removed when the device is seen again OR when the int suppress period expires.
- Added `last_orphaned_log` to `DeviceMetadata` to throttle INFO logs to once per `threshold_days`.

#### Test plan
- [x] 1264 pytests pass (10 skipped), +8 new suppress tests + 1 HGI extraction test
- [x] ha_sim_test R77 (HGI not re-discovered): 3/3 checks pass
- [x] ha_sim_test R78 (suppress_not_seen persists + no re-notify): 4/4 checks pass
- [x] Live on `hass`: `18:149488` no longer marked REMOVED, `29:176861` still notifies (not yet dismissed)

### DIFF
--- a/custom_components/ramses_cc/config_flow.py
+++ b/custom_components/ramses_cc/config_flow.py
@@ -1619,15 +1619,16 @@ class RamsesOptionsFlowHandler(BaseRamsesFlow, OptionsFlow):
         # for new devices.  This stashes schema_device_ids so that
         # check_for_new_devices can suppress notifications for devices
         # that are already in the schema but lost their metadata (issue 917).
+        # NOTE: use _extract_schema_device_ids (unstripped) so that HGI
+        # (18:) entries are included — _strip_and_orchestrate drops them
+        # because ramses_rf doesn't need them, but discovery tracking
+        # must know they're in the schema (issue 987).
         config_schema_for_sync = self.options.get(CONF_SCHEMA, {})
         if isinstance(config_schema_for_sync, dict):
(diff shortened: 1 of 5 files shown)


wimpie70:
For the wiki:
_suppress_not_seen will be added in the schema to a device that was reported lost, but dismissed
true: don't notify again
1-n: days to wait before a new notification.
If a device is seen in the mean time, the _suppress_not_seen will be removed.
(this happens also for _suppress_not_seen: true, so we might want to change this behaviour to only remove it for numbered values)

silverailscolo (approved):

Issue #978, opened 2026-08-16
Refactor fan_handler and number platform to use coordinator callbacks rather than hass.bus


PWhite-Eng:
### Summary
Refactor 2411 fan parameter updates between `fan_handler.py` and `number.py` to use direct coordinator/handler callback listeners instead of `hass.bus` event firing (`ramses_cc.fan_param_updated`).

### Background & Architecture
- Currently, `fan_handler.py` fires `hass.bus.async_fire("ramses_cc.fan_param_updated", ...)` and `RamsesNumberParam` entities in `number.py` listen via `hass.bus.async_listen`.
- `EventEntity` (introduced in ramses-rf/ramses_cc#543) is intended for transient user-facing automation events (e.g. RF remote button presses), not internal state propagation for configuration number entities.
- Using direct coordinator listener callbacks or device state dispatch conforms to standard Home Assistant coordinator patterns and eliminates event bus overhead.

### Tasks
- [ ] Implement direct callback subscription mechanism in `fan_handler.py` for 2411 parameter updates.
- [ ] Refactor `RamsesNumberParam` in `custom_components/ramses_cc/number.py` to register callback listeners with the coordinator/handler instead of `hass.bus.async_listen`.
- [ ] Remove `hass.bus.async_fire` calls from `custom_components/ramses_cc/fan_handler.py`.
- [ ] Update test suite in `tests/tests_new/test_fan_handler.py` and `tests/tests_new/test_number.py`.

silverailscolo:
@wimpie70 please add your thoughts when you have the opportunity, what would be required for _extras when we implement this?

wimpie70:
Nothing in ramses_extras needs to change for this refactor itself, since ramses_extras never listened to ramses_cc.fan_param_updated — it tracks parameter entities via standard HA state_changed events.

wimpie70:
However: take note to keep supporting multiple callbacks

wimpie70:
@PWhite-Eng are you doing this ?

PWhite-Eng:
@wimpie70 @silverailscolo 

Yes, I am picking this up now.

Implementation plan:
1. Implement a callback listener subscription mechanism in `fan_handler.py` supporting multiple listeners.
2. Refactor `RamsesNumberParam` in `number.py` to register callback listeners directly with the coordinator/handler.
3. Remove `hass.bus.async_fire` calls for `ramses_cc.fan_param_updated`.
[...]

PR #998, opened 2026-08-18
Fix issue 88: EventEntity fires for all messages when no regex configured

wimpie70:
## Summary
- The `RamsesRegexEvent` entity only fired events when a regex was configured via `CONF_MESSAGE_EVENTS` in advanced features
- If no regex was set, `message_events_regex` was `None`, and the check `if regex and regex.search()` was always `False` — the EventEntity never fired any events
- This meant downstream listeners (like ramses_extras `hvac_fan_card`, which relies on `event.ramses_cc_regex_event` to receive 31DA/10D0 fan data) never received any data, and the card showed no values

## Fix
If no regex is configured (`regex is None`), fire for **all** messages. When a regex IS configured, only fire for matching messages (unchanged).

This is the ramses_cc side of the fix for ramses_extras issue 88 (hvac fan card doesn't show values).

#### Test plan
- [ ] All existing event tests pass (`pytest tests/ -x -q -k event`)
- [ ] Manual: verify `event.ramses_cc_regex_event` fires when no regex configured
- [ ] Manual: verify ramses_extras hvac_fan_card shows values after this fix
- [ ] Manual: verify regex filtering still works when a regex IS configured

Related: https://github.com/wimpie70/ramses_extras/issues/88

### DIFF
--- a/custom_components/ramses_cc/event.py
+++ b/custom_components/ramses_cc/event.py
@@ -252,7 +252,10 @@ class RamsesRegexEvent(RamsesEvent):
             #     )
 
             # filter msg by advanced_config regex, fire an event if a match
-            if regex and regex.search(f"{msg!r}"):
+            # If no regex is configured, fire for all messages (so that
+            # downstream listeners like ramses_extras hvac_fan_card can
+            # receive 31DA/10D0 data without explicit regex configuration).
+            if regex is None or regex.search(f"{msg!r}"):
                 event_data = {
                     "type": RamsesEventType.REGEX,
                     "device_id": msg.src.id,


custom_components/ramses_cc/event.py
```diff
@@ -252,7 +252,10 @@ def async_process_msg(
             #     )
 
             # filter msg by advanced_config regex, fire an event if a match
-            if regex and regex.search(f"{msg!r}"):
+            # If no regex is configured, fire for all messages (so that
+            # downstream listeners like ramses_extras hvac_fan_card can
+            # receive 31DA/10D0 data without explicit regex configuration).
+            if regex is None or regex.search(f"{msg!r}"):
```
silverailscolo, line 258: @wimpie70 I understand it would fix your _extras issue, but our code was designed to _not_ fire unless a regex is entered. That's why it is called a regex event. If you enter a `*`, does it work?

As a user, I often use the event for other purposes and do not want having to script another filter in templates or what. I adjust the regex to define when it fires, or clear it to turn it off.

There must be other ways to listen for _any_ event, just as the logger manages to hear everything.
In HA, are there any built-in ways to plug into "everything" from a given domain.

silverailscolo (changes requested):

custom_components/ramses_cc/event.py
wimpie70: Ah oc, i'll look at this when i'm back at my compu

silverailscolo:
You wrote in https://github.com/wimpie70/ramses_extras/issues/45#issuecomment-3859989516 that for the fan_card to work, users need to enter at least ".*" as regex in _cc.
Yes, it requires some setup (and could be off). But it uses the regex_event as designed.

A better way might be using #978 (issue is on hold until after 0.60 stable).

wimpie70:
good catch, this shortcut had actually some other downsides as well.

This was the Current path (EventEntity — requires .* regex):


packet → add_msg_handler → RamsesRegexEvent.async_process_msg
  → update_data → _trigger_event + async_write_ha_state()
    → HA state machine write (updates recorder, state registry)
    → fires state_changed on HA bus (broadcast to ALL subscribers)
      → browser subscribeEvents('state_changed') receives ALL state_changed events
        → JS filters for entity_id === 'event.ramses_cc_regex_event':

We will now instead (WS command):


packet → add_msg_handler → RamsesMessageStream._handle_msg (already running)
  → _notify_subscribers → WS callback
    → connection.send_message(event_message(msg["id"], data))
      → browser subscribeMessage receives it directly
Per packet: one dict lookup + one WS frame push

Why WS is faster
No async_write_ha_state() — this is the expensive part. It writes to the HA state machine, triggers the recorder (for history), and fires state_changed on the bus. The WS path skips all of that.
No bus broadcast — state_changed goes to every subscriber on the HA bus (every dashboard card, every automation, every integration). The WS path is point-to-point: only the browser connection that subscribed gets the message.
No client-side filtering — the browser currently receives every state_changed event (lights, sensors, switches, everything) and filters in JS for entity_id === 'event.ramses_cc_regex_event'. The WS path only receives ramses messages.
RamsesMessageStream is already running — it already processes every packet for the device_simulator and traffic_collector. Adding one more subscriber is a dict iteration, negligible cost.

Issue #984, opened 2026-08-17
Towards a stable release 0.60.0


silverailscolo:
No new features or refactoring in ramses_rf will be merged until after the release of a stable release 0.60.0, only bug fixes for 0.59.7 and later.

silverailscolo:
Draft summary of changes.
@PWhite-Eng @wimpie70 please add what I missed (and strike out as many lines to keep it focussed ;)

# Ramses RF release 0.60.0 Highlights

## Integration
- configuration schema replaced the known_list + discovery with Passive device scan (On by default)
- any accepted device owned by “me” is automatically placed on the known_list
- the known_list is always active (since 0.59.7)
- handling of new library formats (DTO data transfer objects, refactored parameters etc.)
- implement ConfigEntry.runtime_data
- reduce flash writes: increase SAVE_STATE_INTERVAL to 30m
- fixed inactive 2411 parameter entities
- RamsesError exception hierarchy for useful error feedback instead of “an unknown error occurred”
- new set_polling_interval service
- use intents to generate commands (25% complete)
- 46 bug fixes!

## Library
- ConversationManager (L7 RQ/RP tracking)
- TopologyBuilder + subsystem handlers
- CommandDispatcher (intent → DTO, 50+ actions)
- HvacContradiction detection
- system schema deep_merge fixed
- ramses_tx extracted as a separate package
- added HCC100 cooling support 
- separation of protocol (tx) from domain handling (rf)
- payload dataclasses replacing untyped dicts (80% complete)
- CQRS state projector replaced in-memory dicts/SQLite database: immediately fetch latest value instead of async waiting for reply (99% complete)
- opentherm bridge telemetry
- evohome support: Zone names, temperatures, DHW
- new HVAC topology (“a fan has remote, sensors”), fixes for 2411, filter_remaining
- new PollingManager replaced schedule tasks
- suppress hardware RQ echoes
- improved MQTT stability, autostart and bind transport pause/resume to MQTT connection events
- 58 bug fixes!

## Code quality
- no class mutation, replaced by promotion (just 1 left)
- unified code style (max. 79-character line length, quotes, standardised function and method parameter signatures etc.) 0.59.7
- modernize scheduling subsystem, removed “lock”
- replaced address string slicing with domain properties and DevType enums
- defer string evaluation in logger (using %s instead of f-strings)
- memory reduction using __slots__ and LRU cache for Address and Packet
[...]

PR #1011, opened 2026-08-20
fix: use live entry.options in discovery checkpoints to stop missing_class loop

wimpie70:
## Summary

The coordinator's `self.options` is a `deepcopy` snapshot from `__init__` that becomes stale when the config flow updates the config entry without triggering a reload (e.g. when the user accepts devices or adds `_class` via the `review_discovered` form — the reload is suppressed by the accept flow).

This caused two problems:

1. **`check_missing_class` re-flagged the same devices every 5-minute checkpoint** because it read the stale schema (no `_class`) instead of the live config entry schema (with `_class` just added by the user). The user kept seeing the same 3 TRVs re-prompted in an endless loop.

2. **`sync_learned_topology` overwrote the user's `_class` changes**: it read the stale schema, merged learned topology into it, and wrote the result back via `async_update_entry` — overwriting the `_class` that the user just added. This ensured the `_class` never survived to the next checkpoint even though the user kept clicking "add_class".

### Root cause trace

From the user's log (https://github.com/ramses-rf/ramses_cc/issues/984#issuecomment-5357048378):

```
14:41:12 review_discovered: added _class=TRV for 04:208992
14:41:12 review_discovered: added _class=TRV for 04:034682
14:41:12 review_discovered: added _class=TRV for 04:056675
14:44:22 review_discovered: added _class=TRV for 04:208992  ← same devices!
14:44:22 review_discovered: added _class=TRV for 04:034682
14:44:22 review_discovered: added _class=TRV for 04:056675
14:54:12 review_discovered: added _class=TRV for 04:208992  ← still looping!
```

The user clicked "add_class" three times for the same devices over 13 minutes. Each time, `_class` was written to the config entry, but the next checkpoint read from the stale `self.options` (no `_class`) and re-flagged them. Worse, `sync_learned_topology` then wrote the stale schema back to the config entry, erasing the user's changes.

### Fix

Use `self.entry.options` (the live config entry options) instead of `self.options` (the stale snapshot) in all discovery-related and schema-write-back code paths:

- `_async_discovery_checkpoint` (check_missing_class, check_all_mismatches)
- `_async_start_discovery_scan` (sync_with_schema)
- `_check_rf_contradictions`
- `sync_learned_topology` (read base + all three write-back paths)
- `_sync_remotes_to_schema` (read base + write-back)
- post-reload `check_all_mismatches`
- `async_save_client_state` unload path + `_devices_with_commands` tracking

Tests updated to set both `mock_coordinator.options` and `mock_coordinator.entry.options` for `TestCheckRfContradictions`.

#### Test plan
- [x] All 1360 ramses_cc tests pass
- [x] User verifies the missing_class loop is resolved (devices no longer re-prompted after clicking "add_class")

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -732,7 +732,8 @@ class RamsesCoordinator(DataUpdateCoordinator):
         # (18:) entries are included — _strip_and_orchestrate drops them
         # because ramses_rf doesn't need them, but discovery tracking
         # must know they're in the schema (issue 987).
-        schema = self.options.get(CONF_SCHEMA, {})
+        # Use self.entry.options (live) — see _async_discovery_checkpoint.
+        schema = self.entry.options.get(CONF_SCHEMA, {})
         schema_device_ids = self._extract_schema_device_ids(schema)
         foreign_device_ids = self._extract_foreign_device_ids(schema)
         self.discovery_manager.sync_with_schema(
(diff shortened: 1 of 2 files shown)


wimpie70:
Peter confirmed the discovery loop fixed: https://github.com/ramses-rf/ramses_cc/issues/984#issuecomment-5358626119

silverailscolo (approved):
Good inline annotation to fetch live version

Issue #995, opened 2026-08-17
0.59.7 HVAC: Error setting mode from fan control


silverailscolo:
I see a mix of built-in (Orcon? = symbols, capitalised) options and my custom commands (bulleted, lower case from config schema) in the controller:

<img width="596" height="606" alt="Image" src="https://github.com/user-attachments/assets/aad6c591-8c39-440e-bffe-9cd0aa6a518b" />

When I selected an option from the right hand drop down (5 times), a timeout occurs each time:
```
Logger: homeassistant.components.websocket_api.http.connection
Bron: components/websocket_api/commands.py:336
Integratie: Home Assistant WebSocket API (documentatie, problemen)
Eerst voorgekomen: 05:16:11 (10 gebeurtenissen)
Laatst gelogd: 05:30:35

[548007849392] Error during service call to climate.set_fan_mode: Failed to set fan mode: Timeout waiting for reply to set_fan_mode
[545796859280] Error during service call to climate.set_hvac_mode: Failed to set fan mode: Timeout waiting for reply to set_fan_mode
[545796859280] Error during service call to climate.set_fan_mode: Failed to set fan mode: fan_mode is not valid for scheme 'orcon': laag
[545796859280] Error during service call to climate.set_fan_mode: Failed to set fan mode: fan_mode is not valid for scheme 'orcon': hoog
[545796859280] Error during service call to climate.set_fan_mode: Failed to set fan mode: Timeout waiting for reply to set_fan_mode
```

The HRU does change speed when I use my custom commands, so this seems like an await/feedback problem.

silverailscolo:
System schema:

```
32:022222:
  _class: FAN
  _commands:
    auto: ' I --- 29:123150 29:099029 --:------ 22F1 003 000506'
    automatic: ' I --- 29:123150 29:099029 --:------ 22F1 003 000506'
    boost: ' I --- 29:123150 29:099029 --:------ 22F3 007 00021E04060000'
    high: ' I --- 29:123150 29:099029 --:------ 22F1 003 000406'
    hoog: ' I --- 29:123150 29:099029 --:------ 22F1 003 000406'
    laag: ' I --- 29:123150 29:099029 --:------ 22F1 003 000206'
    llaaag: ' I --- 29:123150 29:099029 --:------ 22F1 003 000106'
    low: ' I --- 29:123150 29:099029 --:------ 22F1 003 000206'
[...]

PR #1014, opened 2026-08-20
fix: add defensive guard for raw packet strings under FAN _commands (#995)

PWhite-Eng:
### The Problem:
When users configure or copy raw RF packet strings directly under `FAN._commands` (instead of `{verb, code, payload}` dict templates), `fan_modes` exposes those custom modes in the Home Assistant UI dropdown. However, selecting a custom mode causes `_is_command_dict()` to return `False`, skipping custom command interception and falling through to native device scheme validation (`self._device.set_fan_mode()`), which crashes with `fan_mode is not valid for scheme 'orcon': <mode>`. Additionally, sending a string command on a FAN remote entity fails with `<fan_id> is not configured for faking`.

### The Fix:
- **`climate.py`**: Added a defensive fallback in `RamsesHvac.async_set_fan_mode()` to detect string commands under `FAN._commands`, parse them via `parse_packet_string()`, and send them directly via `async_send_cmd()` without falling through to scheme validation.
- **`remote.py`**: Updated `RamsesRemote.async_send_command()` faking check so that `HvacVentilator` entities can send raw string packets without requiring `is_faked == True` on the FAN device.
- **`coordinator.py`**: Updated `RamsesCoordinator._migrate_rem_commands_to_fan()` to automatically normalise existing raw string commands on `FAN` entities to structured dict templates.

### Testing Performed:
- Added `test_set_fan_mode_with_fan_raw_string_commands_defensive_guard` to `tests/tests_new/test_climate.py`.
- Added `test_fan_send_command_raw_string_defensive_guard` to `tests/tests_new/test_remote.py`.
- Added `test_fan_string_commands_normalized_to_dicts` to `tests/tests_new/test_coordinator.py`.
- Executed `.venv/bin/ruff format --check .` (75 files formatted).
- Executed `.venv/bin/ruff check --select D custom_components/ramses_cc/climate.py custom_components/ramses_cc/remote.py custom_components/ramses_cc/coordinator.py` (All checks passed).
- Executed `.venv/bin/mypy` (68 source files checked, 0 errors).
- Executed full test suite `.venv/bin/python -u -m pytest -n auto` (1363 passed, 10 skipped, 3 snapshots passed in 20.07s).

### AI Assistance Disclosure:
This contribution was developed with the assistance of Google Gemini 3.7 Flash for code generation and documentation. All logic and implementations were reviewed and verified by the author.

### DIFF
--- a/custom_components/ramses_cc/climate.py
+++ b/custom_components/ramses_cc/climate.py
@@ -1303,32 +1303,44 @@ class RamsesHvac(RamsesEntity, ClimateEntity):
                 # Commands live in schema _commands (handled above via
                 # remotes, which are populated from schema).
 
-            # Check FAN dict templates first (highest priority)
-            if fan_mode in fan_commands and _is_command_dict(
-                fan_commands[fan_mode]
-            ):
+            # Check FAN commands first (highest priority: dict template or
+            # raw string fallback)
+            if fan_mode in fan_commands:
                 cmd_def = fan_commands[fan_mode]
-                packet_str = _build_packet_from_template(
-                    cmd_def, self._device, self.coordinator
-                )
-                _LOGGER.info(
-                    "Intercepted fan_mode '%s'; building from FAN template: "
-                    "%s",
-                    fan_mode,
-                    packet_str,
-                )
-
-                # parse_packet_string parses both CLI and raw packet formats
-                cmd = parse_packet_string(packet_str)
-                if cmd is None:
-                    raise ValueError(
-                        f"Failed to parse packet_str: {packet_str}"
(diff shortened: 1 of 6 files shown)


silverailscolo:
Thanks for the guard fix. The risk of (trying to) parse string commands is that (just like I did), it assumes the src and dst are correct. The beauty of the intent dict is that it does not include either so it can't be wrong.
Could the parsed cmd intent replace the string? That way a user is forced/assisted to migrate to intent commands.
~Just an idea~

@PWhite-Eng Looking closer at the file changes, that's exactly what you wrote, thanks. Merging

silverailscolo (approved):

Issue #976, opened 2026-08-16
0.59.6 - Evohome zone current_temperature appears to using multiple sensors for source


peternash:
Since updating to 0.59.6 the `current_temperature` attributes of my Evohome heating zones have started to fluctuate by a couple of degrees. This is happening on all zones that have more than one device reporting temperature, either multiple TRV's or TRV(s) with an external thermostat.  This includes one zone with a single TRV and an external thermostat.  The zones that contain _only_ a single TRV continue to show a smooth temperature curve.  It would appear that the fluctuations are due to different sensors in the zone being used to report the zone `current_temperature` instead of a single `sensor` device for the zone.  Individual TRV's and the thermostats all report slightly different temperatures within the same zone of course.

I've also noticed that the `binary_sensor.hgi_18_191664_gateway_status` no longer seems to have a `working_schema` attribute.  In earlier releases this attribute used to show a schema that included a `sensor` attribute for each zone showing which device was the authoritative source of zone temperature. The sensor now has a `schema` attribute which doesn't include this data.  On one of my systems I've tried specifying the `sensor` attribute for each zone in the schema (from saved records) but this doesn't change the behaviour.

Example climate zone - one TRV with external thermostat:

<img width="589" height="647" alt="Image" src="https://github.com/user-attachments/assets/b7d49b82-181c-402a-b47b-dc450334dcb6" />

Schema and logs:

[Schema.txt](https://github.com/user-attachments/files/31116850/Schema.txt)
[packet_log.log](https://github.com/user-attachments/files/31116849/packet_log.log)
[home-assistant.log](https://github.com/user-attachments/files/31116848/home-assistant.log)

peternash:
Possibly related to the fix for [#948](https://github.com/ramses-rf/ramses_cc/issues/948) ?

PWhite-Eng:
We investigated the root cause and submitted a fix in ramses-rf/ramses_rf#1070.

### Findings
1. **Unconstrained Child 30C9 Routing**: The temperature projection logic added in 0.59.6 routed all `30C9` packets from child devices where `_parent == zone` directly to `parent.temp_state.temperature`.
2. **Multi-TRV & Thermostat Clobbering**: In zones configured with multiple TRVs or an external room thermostat alongside TRVs, all TRV actuators have `_parent == zone`. Consequently, each TRV's local radiator temperature broadcast was overwriting the zone temperature reading.
[...]

Issue #887, opened 2026-07-31
0.59.1/.2: Zone sensors being set to null on config schema save


bmhughes:
Is this expected behaviour?

When saving this schema:
```yaml
"01:161591":
  stored_hotwater:
    hotwater_valve: "13:101694"
    sensor: "07:045491"
  system:
    appliance_control: "13:062068"
  zones:
    "00":
      actuators:
        - "04:147093"
        - "04:107745"
      class: radiator_valve
      sensor: "04:147093"
    "01":
      actuators:
        - "04:144637"
      class: radiator_valve
      sensor: "04:144637"
    "02":
      actuators:
        - "04:147099"
      class: radiator_valve
      sensor: "04:147099"
    "03":
      actuators:
        - "04:147095"
      class: radiator_valve
      sensor: "04:147095"
```

It's saved as:
```yaml
"01:161591":
  stored_hotwater:
    hotwater_valve: "13:101694"
    sensor: "07:045491"
  system:
    appliance_control: "13:062068"
  zones:
    "00":
      actuators:
        - "04:147093"
        - "04:107745"
      class: radiator_valve
      sensor: null
    "01":
      actuators:
        - "04:144637"
      class: radiator_valve
      sensor: null
    "02":
      actuators:
        - "04:147099"
      class: radiator_valve
      sensor: null
    "03":
      actuators:
        - "04:147095"
      class: radiator_valve
      sensor: null
```

It looks like it's failing the regex match here, but I'm not sure whether that's being used as that seems to be a 'minimal' schema and there is another in `ramses_rf` where the regex would match.
https://github.com/ramses-rf/ramses_cc/blob/8bbe775b6eda468cd731a7e3821624739aab70b4/custom_components/ramses_cc/schemas.py#L202

silverailscolo:
@bmhughes Please install 0.59.2 from HACS Redownload drop down and confirm this is solved.

bmhughes:
@silverailscolo Nope, still happening with 0.59.2

peternash:
[...]

Issue #813, opened 2026-07-14
0.57.12+ schema discovery issues


peternash:
The results of testing schema discovery 0.57.12+ (master of both ramses_cc and ramses_rf)

Strategy:

Cleared caches and schema, enabled passive device discovery
Reviewed discovered devices as prompted, only "accepting" TRVs that had a zone in their discovered data

General observations:

Schema discovery seemed to work well apart from the two room thermostats which were not allocated to a zone and then caused errors when "accepting" them (see below). I had to move them from "Orphans" to the correct zones manually in the schema.

I also had a few instances of a similar error when "accepting" TRVs and I had to remove them from the Orphans list to get the integration to load again.

It took about 30 minutes to capture the zone information for all the TRVs but they started showing much earlier with no associated zone.  ~~It's worth stressing in the documentation to wait until they are showing a zone before "accepting" them.~~ Edit - it looks like the TRVs are moved into the correct zone eventually even it initially "accepted" without a zone.

Notes (timestamps for reference to the attached logs):

12:25

Cleared cache and schema, enabled passive scans

Initial device discovery - most TRVs have no zone

Accepted Evohome controller and OTB
Declined the "foreign" HGI (on my other HA system)

Repeatedly accepted only TRVs with zones

12:33
Example discovery screen 

<img width="586" height="767" alt="Image" src="https://github.com/user-attachments/assets/9dd3d92b-0aec-4a11-add2-a916be07a9a8" />

12:35 

After accepting some TRVs I saw this error.  It's possible I accepted a TRV without a zone by mistake.  This happened four times in total for TRVs (see below).

Error:
```
Failed setup, will retry: Setup failed: for Parent 01:216136 (evohome): Actuator 04:036068 (TRV): <bound method TrvActuator.heat_demand of 04:036068 (TRV)> must be ('BdrSwitch', 'OtbGateway', 'UfhController') (hint: check the configuration schema for conflicting definitions)
```

Removed "04:036068" from Orphans list

12:37

Error:
```
Failed setup, will retry: Setup failed: for Parent 01:216136 (evohome): Actuator 22:012299 (THM): <bound method Temperature.temperature of 22:012299 (THM)> must be ('BdrSwitch', 'OtbGateway', 'UfhController') (hint: check the configuration schema for conflicting definitions)
```

Removed     - "22:012299" from Orphans list

12:40

Error:
```
[...]

PR #1021, opened 2026-08-21
fix: preserve learned representative TRV zone sensors

### LINKED ISSUE #1013: Multi-room zones have no `current_temperature` since 0.59.7
**Describe the bug**

Since 0.59.7, climate entities for my Evohome multi-room zones have no `current_temperature`. The temperature is blank both on the climate control and in the history graph.

This works correctly with 0.59.6. Upgrading to 0.59.7 causes the problem, and it is still present in the 0.59.8 prerelease. Downgrading back to 0.59.6 and restarting Home Assistant restores the temperatures.

The problem appears to specifically affect zones configured as **multi-room**. Zones with multiple TRVs which are not configured as multi-room continue to work.

This may be related to the fix for #976. That issue describes 0.59.6 selecting temperatures from multiple devices within a zone rather than consistently using the authoritative zone temperature. In my case, the affected multi-room zones appear not to have a controller-reported zone temperature at all.

The individual TRVs in the affected zones continue to report their own temperatures correctly; it is only the zone climate entity which has no `current_temperature`.

The packet log appears to support this: the controller's aggregate `30C9` messages omit the affected multi-room zones, while the individual TRVs within those zones continue to transmit `30C9` temperatures.

My suspicion is therefore that 0.59.6 was able to derive/expose a `current_temperature` for these multi-room zones from the individual TRV temperatures, whereas since 0.59.7 there is no fallback when a zone has no controller-reported temperature.

**To Reproduce**

1. Use an Evohome system containing a multi-room zone with multiple TRVs.
2. Install Ramses RF 0.59.6 and restart Home Assistant.
3. Observe that the zone climate entity has a `current_temperature`.
4. Upgrade to Ramses RF 0.59.7 or 0.59.8 prerelease and restart Home Assistant.
5. Observe that the same climate entity has no `current_temperature` - it is blank on the climate control and there is no temperature data on its graph.
6. Observe that the individual TRVs in the affected zone are still reporting their own temperatures.
7. Downgrade to 0.59.6 and restart Home Assistant.
8. The zone `current_temperature` is restored.

In my system the affected zones are:

- Zone `06` - Master Bed
- Zone `07` - Guest Bed
- Zone `08` - Landing

These are the zones configured as multi-room.

For comparison:

- Zone `00` - Living room - works
- Zone `01` - Kitchen - works

Both of these also contain multiple TRVs, but are not configured as multi-room.

For example, Guest Bed is zone `07` and contains three TRVs:

```yaml
'07':
  _name: Guest Bed
  actuators:
    - 04:229039
    - 04:240988
    - 04:240990
  class: radiator_valve
  sensor: null
```

Kitchen also has multiple TRVs but is not configured as multi-room and continues to work:

```yaml
'01':
  _name: Kitchen
  actuators:
    - 04:240802
    - 04:240804
  class: radiator_valve
  sensor: null
```

There is a potentially significant difference visible in the packet log.

The controller's aggregate `30C9` messages contain temperatures for the normal zones, but the affected zones `06`, `07` and `08` are absent. For example:

```text
I --- 01:165167 --:------ 01:165167 30C9 024 0008430107D902086203083004083B05081009085D0A08DC
```

That payload contains zones `00`, `01`, `02`, `03`, `04`, `05`, `09` and `0A`, but not `06`, `07` or `08`.

However, the individual TRVs within the affected zones continue to transmit their own `30C9` temperatures.

For example, Guest Bed (`07`):

```text
I --- 04:229039 --:------ 04:229039 30C9 003 000821
I --- 04:240988 --:------ 04:240988 30C9 003 000840
I --- 04:240990 --:------ 04:240990 30C9 003 00082A
```

This makes me think the absence of zones `06`/`07`/`08` from the controller `30C9` may be normal Evohome behaviour for multi-room zones, rather than missing RF traffic.

The regression would then appear to be that 0.59.6 was able to expose a zone `current_temperature` from the individual device temperatures in this situation, while 0.59.7+ no longer does so.

**Expected behaviour**

Multi-room zone climate entities should expose a `current_temperature`, as they did in 0.59.6.

Where an Evohome multi-room zone does not provide a single controller-level `30C9` temperature, Ramses RF should expose an appropriate temperature based on the individual TRV temperature data rather than returning no `current_temperature`.

**Please include the following information:**

`ramses_cc:` system schema:

```yaml
01:165167:
  _class: CTL
  _owner: me
  stored_hotwater:
    heating_valve: 13:222826
    hotwater_valve: 13:087160
    sensor: 07:042897
  system:
    appliance_control: 13:097523
  zones:
    '00':
      _name: Living room
      actuators:
        - 04:240786
        - 04:241140
      class: radiator_valve
      sensor: null
    '01':
      _name: Kitchen
      actuators:
        - 04:240802
        - 04:240804
      class: radiator_valve
      sensor: null
    '02':
      _name: Bathroom
      actuators:
        - 04:226957
      class: radiator_valve
    '03':
      _name: Hall
      actuators:
        - 04:240806
      class: radiator_valve
      sensor: null
    '04':
      _name: Downstairs WC
      actuators:
        - 04:069739
      class: radiator_valve
      sensor: null
    '05':
      _name: Bedroom 1
      actuators:
        - 04:241144
      class: radiator_valve
      sensor: null
    '06':
      _name: Master Bed
      actuators:
        - 04:240618
        - 04:241142
        - 04:241146
      class: radiator_valve
      sensor: null
    '07':
      _name: Guest Bed
      actuators:
        - 04:229039
        - 04:240988
        - 04:240990
      class: radiator_valve
      sensor: null
    08:
      _name: Landing
      actuators:
        - 04:240992
        - 04:240634
      class: radiator_valve
      sensor: null
    09:
      _name: Bedroom 2
      actuators:
        - 04:240994
      class: radiator_valve
      sensor: null
    0A:
      _name: Dining Room
      actuators:
        - 04:240632
      class: radiator_valve
```

I can provide the complete Home Assistant debug log and packet log if required.

**Screenshots**

Can provide screenshots if useful, but the symptom is simply that the current temperature field is blank for the affected climate entities.

**Additional context**

- Last known working version: **0.59.6**
- Regression present in: **0.59.7**
- Also reproduced with: **0.59.8 prerelease (20 August 2026)**
- Home Assistant was restarted after each upgrade/downgrade.
- Downgrading from 0.59.7/0.59.8 to 0.59.6 restores the temperatures.
- Individual TRV temperature entities continue to work while the zone climate temperature is blank.
- All three affected zones are configured as Evohome **multi-room** zones.
- Non-multi-room zones containing multiple TRVs continue to work.
- The controller does not appear to include the multi-room zones in its aggregate `30C9` temperature messages, although their individual TRVs continue to transmit temperatures.
- This may be related to the changes made following #976.
- Gateway: `18:132024` via MQTT (`rameses_esp` bridge).

wimpie70:
## Summary

- preserve a representative TRV as the zone `sensor` when `ramses_rf` also lists it under `actuators`
- ensure a representative TRV remains an actuator, while still promoting a dedicated `01:`, `22:`, or `34:` thermostat when present
- replace tests that encoded the incorrect assumption that a TRV could never be a zone sensor

## Findings

The live schema attached to https://github.com/ramses-rf/ramses_rf/pull/1096#issuecomment-5360149554 showed that `ramses_rf` initially discovered the representative TRVs correctly. During `sync_learned_topology()`, `ramses_cc` copied those learned sensors and then cleared every `04:` sensor unless it had already been authored in the config entry.

This conflicted with established behavior:

- https://github.com/ramses-rf/ramses_cc/issues/887 documents a TRV legitimately appearing in both `sensor` and `actuators`
- https://github.com/ramses-rf/ramses_rf/pull/1070 uses the designated `zone.sensor` to prevent non-authoritative TRVs from overwriting `current_temperature`
- https://github.com/ramses-rf/ramses_rf/pull/1096 restores discovery of that representative TRV and expects the same device in both roles

The sensor and actuator fields represent independent roles. Keeping the representative TRV in both fields fixes https://github.com/ramses-rf/ramses_cc/issues/1013 without reverting the temperature-source guard from https://github.com/ramses-rf/ramses_cc/issues/976 or the dedicated-thermostat correction associated with https://github.com/ramses-rf/ramses_cc/issues/813.

## Test plan

- [x] `pytest -q` — 1364 passed, 10 skipped
- [x] `mypy custom_components tests`
- [x] `prek run -a`
- [x] `ramses_rf` issue-976 / PR-1096 focused tests — 6 passed
- [x] HA simulator R36, R75, and new R83 — 20 passed
- [x] R83 verifies discovery, topology sync, reload persistence, and authoritative temperature routing
- [x] Full parallel HA simulator run exercised all 69 recipes; issue-1013 coverage passed after hardening its startup retry. R23 and R78 parallel failures passed in isolation; R68 retains its pre-existing active-probe simulator limitation.

## End-to-end coverage

The companion simulator PR is https://github.com/wimpie70/ramses_extras/pull/180.

### DIFF
--- a/custom_components/ramses_cc/schemas.py
+++ b/custom_components/ramses_cc/schemas.py
@@ -1590,19 +1590,6 @@ def sync_learned_topology(
             if isinstance(orig_config_zones, dict):
                 orig_config_zone_keys = set(orig_config_zones.keys())
 
-            # Track original user-authored sensors per zone.  The
-            # issue-813 TRV-demotion logic below must NOT null a sensor
-            # the user explicitly set in the config flow (e.g. a TRV's
-            # built-in sensor that is also listed in actuators).  See
-            # ramses-rf/ramses_cc#887.
-            orig_config_sensors: dict[str, str] = {}
-            if isinstance(orig_config_zones, dict):
-                for z_index, z_val in orig_config_zones.items():
-                    if isinstance(z_val, dict) and isinstance(
-                        z_val.get(SZ_SENSOR), str
-                    ):
-                        orig_config_sensors[z_index] = z_val[SZ_SENSOR]
-
             # 1a. Sync appliance_control
             learned_sys = learned_entry.get(SZ_SYSTEM, {})
             if isinstance(learned_sys, dict):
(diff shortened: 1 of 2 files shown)


wimpie70:
AI assistant: GPT 5.6 Sol Medium

silverailscolo (approved):

PR #1022, opened 2026-08-21
Bump version, _rf dependency

silverailscolo:
Also update ReadMe Actions table and HGI snapshot 

~Will fail until matching ramses_rf is on pypi~


silverailscolo:
@wimpie Some interaction with ramses_rf changes the `tests/tests_new/test_init.py` result. 
Before simply accepting that, can it be explained?

Copilot says:
> ## Root Cause
> 
> The entity state objects now include new fields that weren't previously present:
> 
> 'last_changed': <ANY>
> 'last_reported': <ANY>
> 'effective_polling_interval': dict({...})
> These fields appear to be newly added to the entity state representation, causing the snapshot assertions to fail.

That doesn't clarify _why_ this returns an empty dictionary.
Just browsing the snapshot, this is the HGI StateSnapshot, and given your recent fix indeed should not include any `effective_polling_interval` fields as it did before.
Agreed?

PR #1023, opened 2026-08-21
fix: skip topology sync in review_discovered to prevent reload suppression (issue 1023)

wimpie70:
see https://github.com/ramses-rf/ramses_cc/issues/1013

## Summary

- After a schema clear and device acceptance via the config flow's "Review discovered devices" form, the gateway's `known_list` was not being refreshed, causing the gateway to block all packets from accepted devices and leaving zones without sensors (e.g., TRV temperature not displaying).
- Root cause: `async_save_client_state()` called inside `review_discovered` triggered `sync_learned_topology`'s enriched write-back, which set `_suppress_reload`. The subsequent `_async_save()` reload was then suppressed by the update listener (within the 5-second window), leaving the gateway with a stale empty `known_list`.
- Fix: set `_skip_topology_sync=True` before the pre-reload save in `review_discovered`, preventing the topology sync block from running and setting `_suppress_reload`. The reload from `_async_save()` then proceeds normally, recreating the gateway with the new `known_list` derived from the accepted devices in the schema.

#### Test plan
- [x] Two new unit tests in `test_coordinator.py`:
  - `test_save_client_state_skip_topology_sync_no_suppress_reload` — verifies `_suppress_reload` is NOT set when `_skip_topology_sync=True` (the fix)
  - `test_save_client_state_topology_sync_sets_suppress_reload` — verifies `_suppress_reload` IS set during normal save cycles (counterpart test, ensures no regression in the normal path)
- [x] Full test suite passes (1259 passed, 10 skipped, 3 pre-existing failures deselected)

### DIFF
--- a/custom_components/ramses_cc/config_flow.py
+++ b/custom_components/ramses_cc/config_flow.py
@@ -1991,7 +1991,20 @@ class RamsesOptionsFlowHandler(BaseRamsesFlow, OptionsFlow):
             # 5-minute checkpoint.  After reload, check_for_new_devices
             # sees the devices with no metadata and re-notifies them as
             # NEW (issue 917).
+            #
+            # IMPORTANT: skip topology sync during this save.  Otherwise
+            # sync_learned_topology's enriched write-back sets
+            # _suppress_reload, which suppresses the reload from
+            # _async_save().  Without that reload, the running gateway
+            # keeps its stale (empty) known_list and blocks all packets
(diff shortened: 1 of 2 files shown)


wimpie70:
I"m running ha_sim_test to check on regressions or other issues...full run on 3 containers takes about 25 minutes now...

silverailscolo (approved):

PR #1029, opened 2026-08-23
feat: persistent notification when gateway goes offline

wimpie70:
## Summary

- When no RF packets are received for 5 minutes, a persistent notification is created so the user knows the gateway is offline rather than silently seeing null entity values
- The notification is dismissed automatically when packets resume (handled in the `_on_packet` callback)
- Grace period during startup — no notification until at least one packet has been received

#### Test plan

- [x] `test_gateway_health_no_notification_on_startup` — no false alarm during startup
- [x] `test_gateway_health_notification_when_offline` — notification created after threshold
- [x] `test_gateway_health_no_duplicate_notification` — no re-creation if already notified
- [x] `test_gateway_health_no_notification_when_recent_packets` — no notification when packets are recent
- [x] Full coordinator test suite passes (249 passed)

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -18,6 +18,10 @@ from typing import TYPE_CHECKING, Any, Final, TypeVar
 
 import serial  # type: ignore[import-untyped]
 import voluptuous as vol  # type: ignore[import-untyped, unused-ignore]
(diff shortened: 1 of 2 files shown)


wimpie70:
You probably will tag this with a hold...np

silverailscolo (approved):

Issue #1020, opened 2026-08-20
HGI not-me 'potentially a Foreign gateway'


silverailscolo:
Running 0.59.9-ish current master, this warning shows in the log since the discovery_scan config schema allows multiple HGI's:

```
Logger: ramses_tx.protocol.base
Bron: runner.py:290
Eerst voorgekomen: 23:43:10 (1 gebeurtenis)
Laatst gelogd: 23:43:10

Device 18:154951 is potentially a Foreign gateway, the Active gateway is 18:130140, alternatively, is it a HVAC device?, configure the known_list/block_list as required
```

Might be rephrased to "check the Configuration Schema." but seems just noise.

silverailscolo:
Fixed in https://github.com/ramses-rf/ramses_rf/pull/1099

silverailscolo:
@wimpie70 Still seeing this warning in 0.59.10 (rc-2). At least we must rephrase reference to known_list to schema

wimpie70:
I had to dig a bit deeper:

>The issue 822 scenario ("foreign HGIs communicate with our controller and the controller's responses are addressed to them") only makes sense for an HGI that is actually part of our system — e.g.:

>A second/old HGI still paired to our controller
A multi-gateway setup where two HGIs poll the same controller
In both those cases, the HGI is ours (same _owner), just not yet in the known_list.

>So the current logic has it backwards:

>Foreign HGI (_owner: not-me) → won't talk to our controller → safe to block
Unknown HGI (not in schema at all) → might be our second HGI → don't block, log at INFO
Known HGI (same _owner, in known_list) → no message needed
The ramses_cc coordinator at coordinator.py:1692 explicitly exempts foreign HGIs from the block_list (str(k)[:2] != "18"), based on the issue 822 reasoning. But if the HGI truly has a different _owner, that exemption is wrong — it should go to the block_list like any other foreign device.

By fixing this we would remove the 18: exemption for foreign-owned HGIs, so they get blocked. That would:

>Silence the warning for foreign HGIs (they're blocked, so _is_wanted_addrs returns False before reaching warn_foreign_hgi)
Keep the issue 822 protection for unknown/our-own HGIs (still exempt from blocking)
Keep the ramses_rf INFO-level fix (PR 1107) as a fallback for truly unknown HGIs not in any schema

silverailscolo:
Nice, is gone in 0.59.11

PR #1032, opened 2026-08-23
fix: block foreign-owned HGIs instead of exempting them

wimpie70:
## Summary

- Foreign HGI devices (`18:`) with `_owner` different from the root `_owner` are now added to the `block_list` like any other foreign device.
- Previously, `18:` devices were exempted from the `block_list` based on issue 822 (eavesdropping on controller responses addressed to a foreign HGI). However, a foreign HGI from a different owner does not communicate with our controller — its traffic is addressed to the neighbour's controller.
- The issue 822 exemption only makes sense for an unknown HGI that might be our own second gateway not yet configured. ramses_rf handles that case separately (unknown HGIs not in any list are still allowed through at INFO log level).
- This silences the "potentially a Foreign gateway" warning for foreign-owned HGIs: they are now blocked at the filter level before the warning code is reached.

Companion PR: ramses-rf/ramses_rf#1107 (ramses_rf side: block_list now respected for `18:` devices, unknown HGIs log at INFO)

Ref: issue ramses-rf/ramses_cc#1020

#### Test plan

- [x] `ruff check` — clean
- [x] `mypy --strict` — no new errors (7 pre-existing)
- [x] `pytest tests/tests_new/test_coordinator.py` — 251 passed
- [x] Updated `test_create_client_foreign_hgi_not_in_block_list` → `test_create_client_foreign_hgi_in_block_list` to assert foreign HGIs ARE in the block_list

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -1691,6 +1691,11 @@ class RamsesCoordinator(DataUpdateCoordinator):
                 block_list[str(k)] = {}
         if block_list:
             gateway_kwargs["block_list"] = block_list
+            _LOGGER.debug(
+                "block_list: %s (root_owner=%s)",
+                list(block_list.keys()),
+                root_owner,
+            )
         # Strip commands from traits (ramses_rf doesn't accept them)
         sanitized_known_list = {
             device_id: (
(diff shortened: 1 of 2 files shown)


silverailscolo:
@wimpie70 Is this PR based on #1029 ?
I see some of the same changes (homeassistant.components.persistent_notification imports etc.) as your 0.60.1 "feature" PR.

wimpie70:
Yes, srr, I'll fix it shortly

silverailscolo (approved):

PR #1033, opened 2026-08-23
fix: enable passive scan by default on v2→v3 migration

wimpie70:
## Summary

- The v2→v3 config entry migration now enables `passive_scan` by default in `advanced_features`.
- Pre-0.57 (v2) users typically had `enforce_known_list=False`, so devices not in the known_list were still allowed through. The v3 migration makes `enforce_known_list` always-on, which would silently filter out those devices.
- Enabling passive scan during migration lets the user re-discover those devices via the discovery flow and add them to the schema. The user can disable it from advanced features once they've reviewed everything.

#### Test plan

- [x] `ruff check` — clean
- [x] `pytest tests/tests_new/test_init.py tests/tests_new/test_config_flow.py` — 97 passed, 2 skipped
- [x] Updated `test_async_migrate_entry_v1_to_v3` — asserts `passive_scan: True`
- [x] Updated `test_async_migrate_entry_v2_to_v3` — asserts `passive_scan: True`
- [x] Updated `test_migrate_entry_v2_to_v3` — asserts `passive_scan: True`
- [x] Updated `test_migrate_entry_v2_to_v3_no_known_list` — asserts `passive_scan: True` even without known_list
- [x] Updated `test_migrate_entry_v1_to_v3` — asserts `passive_scan: True`
- [x] Updated `test_chained_config_entry_migration_v1_to_v3` — asserts `passive_scan: True`

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -398,6 +398,27 @@ async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         # Remove deprecated disabled_devices key (replaced by _disabled trait)
         new_options.pop("disabled_devices", None)
 
+        # Enable passive scan by default for upgrading users.
+        # Pre-0.57 (v2) users had enforce_known_list=False in many cases,
+        # so devices that were not in the known_list were still allowed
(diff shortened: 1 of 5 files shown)


silverailscolo (approved):
I understand, during migration. Thanks

Issue #1010, opened 2026-08-20
BDRs go unavailable for long periods


bmhughes:
In a similar vein to #925 I'm seeing BDRs reporting unavailable for large periods of time.

<img width="1832" height="120" alt="Image" src="https://github.com/user-attachments/assets/e173a38b-2958-4b04-88a9-44ac6f2b9033" />

`13:101694` is my DHW BDR91 and I know for a fact that it was on somewhere between 0500-0530 to heat DHW but it doesn't report as Off until 0720.

I can see various relay demand/heat demand messages being receieved but I can't follow from the code what drives the BDR entity state to try and narrow it down. The other two BDR are for two heating zones so I'd expect those to be permanently off. Here's a sample over the last 7 days.

<img width="1832" height="120" alt="Image" src="https://github.com/user-attachments/assets/401a7e75-ada2-487f-b08d-90cebd6471f4" />

I'm currently running the latest 0.59.8 but the snapshot of the last week will cover 0.59.6 onwards.

Schema:
```
_owner: me
"01:161591":
  _alias: Evohome Upstairs Controller
  _name: Evohome Upstairs Controller
  _class: CTL
  _owner: me
  stored_hotwater:
    hotwater_valve: "13:101694"
    sensor: "07:045491"
  system:
    appliance_control: "13:062068"
  zones:
    "00":
      _alias: Evohome Bedroom 1
      _name: Bedroom 1
      actuators:
        - "04:147093"
        - "04:107745"
      class: radiator_valve
      sensor: "04:147093"
    "01":
      _alias: Evohome Bedroom 2
      _name: Bedroom 2
      actuators:
        - "04:144637"
      class: radiator_valve
      sensor: "04:144637"
    "02":
      _alias: Evohome Bedroom 3
      _name: Bedroom 3
      actuators:
        - "04:147099"
      class: radiator_valve
      sensor: "04:147099"
    "03":
      _alias: Evohome Bedroom 4
      _name: Bedroom 4
      actuators:
        - "04:147095"
      class: radiator_valve
      sensor: "04:147095"
"01:258891":
  _alias: Evohome Downstairs Controller
  _name: Evohome Downstairs Controller
  _class: CTL
  _owner: me
  system:
    appliance_control: "13:213343"
  zones:
    "00":
      _alias: Evohome Kitchen
      _name: Kitchen
      actuators:
        - "04:012975"
        - "04:012979"
      class: radiator_valve
      sensor: "04:012979"
    "01":
[...]

PR #1043, opened 2026-08-24
fix: clear async attr cache on packet arrival for real-time state updates (issue 1042)

wimpie70:
## Summary

- Call `clear_async_attr_cache(self)` in `_async_update_and_write_state` before `async_write_ha_state()` so that freshly-received packet data is visible immediately, bypassing the 30-second cooldown in `resolve_async_attr`.

## Root Cause

When a packet arrived and the coordinator sent `SIGNAL_UPDATE` to an entity, `_async_update_and_write_state` called `async_write_ha_state()` without clearing the `resolve_async_attr` cache. This meant async state readers (e.g. `BdrSwitch.active`, `TrvActuator.heat_demand`, `OtbGateway.rel_modulation_level`) returned stale cached values for up to 30 seconds after the underlying state was updated by the ingestion pipeline.

The 30-second cooldown in `resolve_async_attr` exists to prevent command floods from UI re-renders (e.g. `system_mode()` dispatches a 2E04 RQ when unhydrated). But it also prevented fresh data from being fetched when a packet arrived — the entity showed the old relay state until the cooldown expired.

## The Fix

Call the existing `clear_async_attr_cache(self)` helper (already used by `force_update`) in `_async_update_and_write_state` before `async_write_ha_state()`. This:
- Invalidates the cache so the next property access dispatches the async getter fresh
- Bypasses the cooldown for this one state write cycle
- Still protects against UI re-render floods (the cooldown is re-established after the getter completes)
- Fixes all 75+ async state readers in ramses_rf, not just `BdrSwitch.active`

## Why not convert async methods to properties?

ramses_rf has 75+ "falsely async" state-reader methods (e.g. `heat_demand`, `relay_demand`, `temperature`, `co2_level`, `fan_mode`, `bypass_position`, etc.) that just read state attributes without side-effects. Converting them all to properties would be a massive change with high regression risk. This fix addresses the root cause in ramses_cc instead — one line, benefits all entities.

## Context

- Issue: https://github.com/ramses-rf/ramses_cc/issues/1042
- Originally reported by @bmhughes in issue 1041 comments
- `clear_async_attr_cache` was already implemented for `force_update` — this PR reuses it for packet-driven updates

## Test plan

- [x] ramses_cc unit tests: 1376 passed, 10 skipped
- [ ] ha_sim_test with BDR 3EF1 relay state verification
- [ ] Manual verification on `hass` instance with real BDR relays

### DIFF
--- a/custom_components/ramses_cc/entity.py
+++ b/custom_components/ramses_cc/entity.py
@@ -18,7 +18,7 @@ from ramses_rf.devices import Fakeable
 from ramses_rf.entity import Entity as RamsesRFEntity
 
 from .const import DOMAIN, SIGNAL_UPDATE
-from .helpers import resolve_async_attr
+from .helpers import clear_async_attr_cache, resolve_async_attr
 
 if TYPE_CHECKING:
     from .coordinator import RamsesCoordinator
@@ -177,6 +177,13 @@ class RamsesEntity(CoordinatorEntity):
         """Safely write HA state using an async lock.
 
         Prevents event loop saturation from concurrent updates.
+
+        Clears the async attribute cache so that freshly-received packet
+        data is visible immediately, bypassing the 30-second cooldown in
+        resolve_async_attr.  Without this, async state readers (e.g.
+        BdrSwitch.active, TrvActuator.heat_demand) return stale cached
+        values for up to 30 seconds after a packet updates the underlying
+        state (issue 1042).
         """
         if self._update_lock.locked():
             self._dropped_updates += 1
@@ -193,4 +200,5 @@ class RamsesEntity(CoordinatorEntity):
             return
 
         async with self._update_lock:
+            clear_async_attr_cache(self)
             self.async_write_ha_state()


silverailscolo:
@wimpie70 Would you like to have another rc to confirm this change? I can start to prep _rf if so

wimpie70:
yes, if @bmhughes can test this before 0.60.0 would be great.
Still running ha_sim on this...with some issues....

bmhughes:
Sure thing @wimpie70 I'll be able to test it

wimpie70:
nice, I'm still running the sim. Seems the issue I see there is in the test itself, not in the fix on _cc....but I need to be sure...Then I'll tag this ready for review and Egbert can take it to a release

wimpie70:
I checked  but the earlier BDR issues were not related to #1042/#1043 — they were a different layer:

#1010 (fixed by ramses_rf #1095) — BDRs going unavailable due to a too-short 1h heartbeat timeout. Fixed by extending to 24h for BDR/JIM/JST.
#834/#931/#947 — BDRs miscategorised in schema discovery. Fixed in ramses_rf via authoritative domain_id and re-parenting.

#1042/#1043 is a state-staleness bug (async attr cache cooldown), not availability or schema — independent root cause

silverailscolo (approved):

silverailscolo:
> Sure thing @wimpie70 I'll be able to test it

@bmhughes Please install 0.59.12 (rc-5) from HACS dropdown and confirm both issues are fixed. Thanks for reporting

PR #1060, opened 2026-08-26
fix: deprecate legacy known_list from `configuration.yaml` on first upgrade (issue 1055)

### LINKED ISSUE #1055: Known List not removed from `configuration.yaml`
I think we should remove it on upgrading, make a backup file first, like we do for normal known list.

wimpie70:
## Summary

- `async_setup` now detects legacy `known_list`, `block_list`, and `enforce_known_list` in `configuration.yaml`
- Backs up `known_list` and `block_list` to `ramses_cc_backups/` as a YAML file
- Creates a **persistent notification** telling the user to remove the keys manually
- **Does NOT modify `configuration.yaml`** — preserves user comments and formatting

## Context

Phase 4 migrated `known_list` from `configuration.yaml` into the config entry schema (v2→v3 migration in `async_migrate_entry`), but the YAML file was never cleaned up. `block_list` is now derived from `_owner`/`_skipped` traits at runtime. `enforce_known_list` is always-on. Users who had these keys in `configuration.yaml` still have them sitting there, causing HA to log warnings on every restart.

## Approach

Per @silverailscolo's feedback, this PR takes a conservative approach — it does **not** rewrite `configuration.yaml` (which would destroy comments and formatting). Instead:

1. **Backup**: `known_list` and `block_list` are saved to `ramses_cc_backups/backup_<timestamp>_yaml_known_list.yaml` so no data is lost. HA-internal dict types (NodeDictClass) are converted to plain dicts via `json.loads(json.dumps(...))` for clean, portable YAML.
2. **Persistent notification**: A HA persistent notification is created listing all keys to remove
3. **Warning log**: A `WARNING`-level log message is emitted

The user removes the keys manually at their convenience. The backup file allows them to copy/paste values into the schema editor if needed.

Other YAML options (`serial_port`, `ramses_rf`, `scan_interval`, `advanced_features`, `packet_log`, TCS schema structure) are left untouched — only the three legacy keys are flagged.

The first-time import flow (`async_step_import`) still migrates `known_list` traits into the schema for users upgrading from YAML-only setups (no config entry exists yet). This is the only chance to transfer those traits.

Fixes https://github.com/ramses-rf/ramses_cc/issues/1055

#### Test plan

- [x] `pytest tests/tests_new/test_init.py` — 25 passed
- [x] `ruff check` + `ruff format` — clean
- [x] `test_yaml_known_list_cleanup_backs_up_and_notifies` — verifies backup file contains both `known_list` and `block_list`, and persistent notification mentions all three keys
- [x] `test_yaml_known_list_cleanup_noop_without_known_list` — verifies nothing happens when no legacy keys are present
- [x] Manual: verified on ha-sim — persistent notification appears, backup file created with clean YAML (both `known_list` and `block_list`), integration loads normally, `configuration.yaml` not modified

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -81,6 +81,7 @@ from .const import (
     CONF_MQTT_TOPIC,
     CONF_MQTT_USE_HA,
     CONF_PASSIVE_SCAN,
+    CONF_RAMSES_RF,
     CONF_SEND_PACKET,
     DOMAIN,
     STORAGE_KEY,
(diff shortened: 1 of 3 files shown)


silverailscolo:
@wimpie70 Take your time to fine tune this PR. Anything else from the issue list planned? If not, I'll release 0.60.1 with all merged bug fixes later today. Seems stable enough to replace 0.60.0 in HACS.

silverailscolo:
We should deprecate/stop importing the schema/known_list from configuration.yaml soon.

That step is really > 2Y old, and repeated imports cause inefficiency, both in the box and for support.
HA also doesn't like it. Config should be in the config flow. (period)

After the warning, can't you stop importing it again?

wimpie70:
yes, we should stop importing the known list, but maybe we still want to be able to import some 'hidden' settings ?

silverailscolo:
> yes, we should stop importing the known list, but maybe we still want to be able to import some 'hidden' settings ?

In the CLI, you _can_ use a configuration file, but that is separate from HA config. HA Actions provide a fine way to tune the integration.
E.g. the Remove Device action works great! Nobody knows its value yet.

wimpie70:
Tested with a fake configuration.yaml. Known List, Block List, Enforce_known_list are flagged, the other YAML options are left untouched.
Still imported on first update from the old system.

wimpie70:
No tests ?

ah..issues on GH...

silverailscolo:
Yep https://www.githubstatus.com

silverailscolo (approved):

silverailscolo:
Closing and reopening to restart tests

PR #1076, opened 2026-08-28
Config schema YAML (spelling)

silverailscolo:
Consistent console hint `true` lower case spelling (python bool: True)


silverailscolo:
@wimpie70 The code uses true and True, false and False. 
What is your take on config YAML capitalisation? Streamline?

wimpie70:
I'm used to lowercase from all other programming languages, so yeah, I do prefer lower, but can read starting capital as well....If that's a thing we can do lower....

PR #1082, opened 2026-08-29
fix(discovery): handle integer _suppress_not_seen in check_for_lost_devices

wimpie70:
## Summary

- `check_for_lost_devices` only checked `_suppress_not_seen is True` (identity with `True`), but the 'keep' action in `review_device_health` sets `_suppress_not_seen = 7` (integer, suppress for 7 days)
- Integer values were ignored, causing devices to be immediately re-flagged as lost after the config entry reload triggered by the 'keep' submission
- This mirrors the logic already present in `check_orphaned_devices`, which handles both `True` (suppress forever) and `N` int (suppress for N days from `last_seen`, then re-notify)

## Root cause

The 'keep' action for lost devices in `async_step_review_device_health` sets:
```python
dev_entry["_suppress_not_seen"] = 7  # integer, not True
```

But `check_for_lost_devices` checked:
```python
if schema_entry.get("_suppress_not_seen") is True:  # only matches True, not 7
    continue
```

So the suppress was ignored and the device was re-flagged as lost on the next reload.

## Fix

Updated `check_for_lost_devices` to handle both `True` and integer values, with expired-suppress detection for the integer mode — same logic as `check_orphaned_devices`.

## Test plan

- [x] R50 ha_sim_test passes (device health tracking — orphaned/lost devices)
- [x] `pytest tests/tests_new/test_ha_compat.py` passes
- [x] No unexpected ERROR/WARNING logs in ha_sim_test

### DIFF
--- a/custom_components/ramses_cc/discovery.py
+++ b/custom_components/ramses_cc/discovery.py
@@ -2370,13 +2370,52 @@ class DiscoveryManager:
                 continue
 
             # Respect _suppress_not_seen from schema (issue 988)
+            # Accepted values: True (suppress forever) or N int (suppress
+            # for N days from last_seen, then re-notify).  This mirrors
+            # the logic in check_orphaned_devices.
             if schema is not None:
                 schema_entry = schema.get(device_id)
-                if (
-                    isinstance(schema_entry, dict)
-                    and schema_entry.get("_suppress_not_seen") is True
-                ):
-                    continue
+                if isinstance(schema_entry, dict):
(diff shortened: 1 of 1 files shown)


silverailscolo (approved):

wimpie70:
I'll resolve them

silverailscolo:
Thanks, ready for _cc 0.60.2 ?

wimpie70:
yes

Issue #1116, opened 2026-09-01
0.60.3: Orcon Fan on controller not showing its state (ramses_extras unaffected)


cablegunmaster:
**Describe the bug**
The orcon Fan entity (controller, right of image) does not show the state of the fan.
The Ramses Extra is showing it correctly.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to your entity on the ~dashboard~ controller, disable "Auto - Extras Off" by click on it and make sure its surrounding light is gone.
2. Click on anything other than Auto (any command)
3. See nothing happens on the FAN entity (right side) and the dropdown just goes back to blank

The actual Orcon fan on the ~dashboard~ controller keeps its state on "Auto".
The ramses_extras GUI shows the correct setting selected.
The dropdown underneath the unit goes to a "Dot" with nothing selected behind it, the dropdown does no retain its selection.
On ramses extras it keeps showing "Unit auto" and Offline, even though it works with sending commands.

After reselecting multiple times, the unit on the ramses_extras card shows _offline_.

Provide code if you can.

The action itself is reacting, it's just not on the ~dashboard~ climate entity.

**Expected behaviour**
The fan entity shows the last action dropdown send to it, and updates the word Auto with its current state its selected from. and the offline should not be shown on the ramses extra, because its reacting to the state changed.

**Please include the following information:**
 - `ramses_cc:` your integration **config** (copy it from HA > Integrations > Ramses RF > Configure cog wheel > System Schema and Known Devices > System schema)
 
```
_owner: cablegunmaster
18:050704:
  _class: HGI
  _owner: cablegunmaster
32:136873:
  _bound: 37:050704
  _class: FAN
  _commands:
    _comment: Commands on FAN (Phase 3b) — target this entity for automations
    auto:
      code: 22F1
      payload: '000404'
      verb: I
    away:
      code: 22F1
      payload: '000004'
      verb: I
    boost:
      code: 22F1
      payload: '000604'
      verb: I
    'false':
      code: 22F1
      payload: '000004'
      verb: I
    high:
      code: 22F1
      payload: '000304'
      verb: I
    low:
      code: 22F1
      payload: '000104'
      verb: I
    medium:
      code: 22F1
      payload: '000204'
      verb: I
    timer_10mins:
      code: 22F3
      payload: 00020A03040000
      verb: I
    timer_15mins:
      code: 22F3
      payload: 00020F03040000
      verb: I
    timer_30mins:
      code: 22F3
[...]

PR #1117, opened 2026-09-01
fix(climate): optimistic `fan_mode` update + filter `fan_mode`s to `22F1`

wimpie70:
## Summary

Two bugs in `climate.py` caused the Orcon fan dropdown to revert to blank after selecting a mode (issue https://github.com/ramses-rf/ramses_cc/issues/1116):

1. **No optimistic state update after `set_fan_mode`**: Ventilators don't ack 22F1 (fire-and-forget), so `fan_info` stays stale until the next 31D9/31DA broadcast. The dropdown reverted instantly because `fan_mode` still read the old value. Fix: set `_last_known_fan_info = fan_mode` before `async_write_ha_state()` in all three send paths (FAN dict template, REM packet string, native fallback).

2. **All `_commands` injected into `fan_modes` dropdown**: 22F3 timers, 22F7 bypass, 2411 config, and 10D0/31DA info commands appeared in the dropdown but can never be represented as a persistent `fan_mode` state. Fix: new `_is_mode_command()` helper in `remote.py` filters to 22F1 entries only.

### Note

A third issue remains: the 31DA parser reports `fan_info` as `"speed 1, low"` / `"speed 2, medium"` / `"speed 3, high"`, which don't match the dropdown names `"low"` / `"medium"` / `"high"`. > see #1118

#### Test plan

- [x] 17 unit tests added (10 for `_is_mode_command`, 4 for optimistic update, 3 for `fan_modes` filtering)
- [x] `ruff check` — passed
- [x] `mypy` — passed (0 issues)
- [x] `pytest` (full suite) — 1482 passed, 15 skipped, 0 failures
- [x] Manual test on hass: select mode from dropdown, verify it sticks


.github/workflows/post-cov.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         if: steps.download.outcome == 'success'
         # can only run on pr_target repository originally, now runs
         # safely via artifact decoupling
-        uses: MishaKav/pytest-coverage-comment@v1.11.2
+        uses: MishaKav/pytest-coverage-comment@v1.11.0
```
silverailscolo: This is a mistake from earlier branching, please remove

custom_components/ramses_cc/climate.py
```diff
@@ -1514,6 +1516,11 @@ async def async_set_fan_mode(self, fan_mode: str) -> None:
                     await self._device._gateway.async_send_raw_command(
                         cmd, num_repeats=2, priority=Priority.HIGH
                     )
+                    # Optimistic update: ventilators don't ack 22F1, so
```
silverailscolo: This is by design. See comment on issue.

silverailscolo (changes requested):

silverailscolo:
Moved to #1118

wimpie70:
Both review items are addressed in 62d0ca32 and the preceding commit (c7e72608):

1. **`post-cov.yml` downgrade** — reverted, the file now matches master (`@v1.11.2` restored). This was an accidental carry-over from earlier branching, as you noted.

2. **Optimistic state update** — removed. The `async_write_ha_state()` calls after the 22F1 send have been dropped in all three send paths, with comments noting the fire-and-forget behaviour is by design (issue 1116). State feedback comes from the 31DA broadcast, as you said.

Regarding the `_normalise_fan_info_for_selector()` function (added in c7e72608): per your feedback on ramses_rf PR 1183, the descriptive `fan_info` from ramses_rf is **preserved as-is** in `extra_state_attributes["fan_info"]`. Normalisation is applied **only** to the `fan_mode` selector property so the dropdown matches an entry in `fan_modes`. Unknown descriptive values pass through unchanged. This keeps the manufacturer's instruction-book names visible to the user while letting the selector work.

The 22F1-only filtering of `fan_modes` (excluding 22F3 timers, 22F7 bypass, 2411 config, etc.) is unchanged from the original PR.

Would appreciate a re-review when you have a moment.

silverailscolo (approved):
Thanks

PR #1135, opened 2026-09-02
Await faked CO2 level updates

maleadt:
- [x] AI Coding Assistance disclosure: used for 75%

ramses_cc 0.60.4 pins ramses-rf 0.60.4, where CO2 updates use the asynchronous `set_co2_level()` API.

Call and await that method instead of assigning co2_level directly so the Home Assistant service completes the RF announcement before returning.

### DIFF
--- a/custom_components/ramses_cc/sensor.py
+++ b/custom_components/ramses_cc/sensor.py
@@ -222,8 +222,7 @@ class RamsesSensor(RamsesEntity, SensorEntity):
 
     # the following methods are integration-specific service calls
 
-    @callback
-    def async_put_co2_level(self, co2_level: int) -> None:
+    async def async_put_co2_level(self, co2_level: int) -> None:
         """Cast the CO2 level (if faked).
 
         :param co2_level: The CO2 concentration in parts per million (ppm).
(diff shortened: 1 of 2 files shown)


custom_components/ramses_cc/sensor.py
```diff
@@ -238,8 +237,8 @@ def async_put_co2_level(self, co2_level: int) -> None:
             raise TypeError(f"Cannot set CO2 level on {device}")
         # TODO: Until here
 
-        # setter will raise an exception if device is not faked
-        device.co2_level = co2_level  # would accept None
+        # set_co2_level() will raise an exception if device is not faked
```
silverailscolo, line 240: @PWhite-Eng does this call change align with your recent refactoring, or should we keep fire-and-forget calls, e.g. to prevent blocking?

PWhite-Eng: Yes, this change aligns directly with both our recent command bus refactoring and Home Assistant architecture standards. It should definitely be merged.

A few key points for context:

1. **Fixes a broken call**: In `ramses_rf`, `CarbonDioxide.set_co2_level()` is an async method; there is no property setter. The legacy assignment `device.co2_level = co2_level` was merely overwriting the object attribute in memory and never transmitted an RF packet.
2. **Non-blocking dispatch**: `PUT_CO2_LEVEL` dispatches an `I` (announcement) packet (`Code._1298`). Because announcements do not expect a response, `dispatcher.send()` returns as soon as the frame is queued/transmitted over the serial link without awaiting a remote reply. As an `async def` coroutine, it yields cleanly and will not block the event loop.
3. **Consistency across sensor platforms**: This brings `async_put_co2_level` into parity with `async_put_room_temp` and `async_put_dhw_temp`, both of which are already `async def` and await `device.set_temperature()`.
4. **Error propagation**: Awaiting the call ensures errors (e.g. `DeviceNotFaked`) bubble up cleanly to Home Assistant automations.

silverailscolo (approved):

PR #1148, opened 2026-09-04
feat: option to disable gateway-offline notification

### LINKED ISSUE #1123: Feat: Option to disable "no packets" notification
Hello,

I only have one thing on my network, a PIV fan in the loft.

I've set the Gateway timeout to 60 minutes  already (the maximum) but I get a notification every hour telling my my gateway has an issue - it doesn't.

Can we get an option to entirely disable this notification?

wimpie70:
## Summary

- Adds a "Notify when gateway is offline" boolean toggle under **Advanced features** (default: enabled, preserving existing behaviour)
- When disabled, `_check_gateway_health()` returns early — no persistent notification is created or dismissed
- Addresses users with very quiet networks (e.g. a single PIV fan) where the gateway routinely exceeds the timeout between transmissions, causing a recurring "no packets received" notification every cycle

Closes https://github.com/ramses-rf/ramses_cc/issues/1123

#### Test plan

- [x] `ruff check` — clean
- [x] `mypy` — no issues
- [x] `pytest tests/tests_new/test_coordinator.py` — 263 passed
- [x] `pytest tests/tests_new/test_config_flow.py` — 91 passed
- [x] New test `test_gateway_health_disabled_no_notification` verifies no notification/dismiss is called when the option is `False`
- [ ] Manual test on ha-sim: verify the toggle appears under Advanced features and suppresses the notification


wimpie70:
The toggle can be found under Advanced features.
This is different from the _suppress_not_seen = true  (or x days) option since this it about the gateway protection and returns the check early

silverailscolo (approved):

PR #1151, opened 2026-09-04
Expose Orcon ventilation demand service

maleadt:
An Orcon HRC in auto mode responds to transient 31E0 ventilation demand. Publishing 1298 CO2 telemetry alone does not change fan output, while switching fan modes would turn continuous demand control into a sequence of manual overrides.

Add a put_ventilation_demand entity service for faked CO2 sensors. It resolves the bound Orcon fan from the integration schema, validates the target and faked-device capability, converts the public 0-100 percent value to the 0.0-1.0 ramses_rf representation, and delegates transmission to set_ventilation_demand.

With this, I can influence the `auto` mode and have stepless fan control without having to switch or reconfigure modes :tada:

Validated locally on my HRC 350.

PR template:

- [x] AI Coding Assistance disclosure: not used/used for 75%


custom_components/ramses_cc/services.yaml
```diff
@@ -811,6 +811,28 @@ put_co2_level:
           mode: slider
 
 
+put_ventilation_demand:
```
silverailscolo, line 814: Add the service (name + description) also on the ReadMe Services list (if it is public)

silverailscolo (changes requested):

silverailscolo (approved):

PR #1166, opened 2026-09-06
fix: clear _skipped when adding _class in review_discovered

wimpie70:
## Summary

When a device appeared in the missing_class section of review_discovered
(in schema but no `_class`, scan engine has a `likely_type`) and the user
chose "Add _class", the `_skipped` flag was never cleared. The device got
a `_class` but stayed deferred — no entities were created, commands were
blocked.

Fix: clear `_skipped` when adding `_class`. This effectively accepts the
device in one step. If a new msg code later suggests a different class,
`check_class_mismatches` will catch it via the mismatch review path.

## Verification

- All 1679 tests pass.
- Pre-commit hooks pass (ruff, mypy, codespell, etc.).

#### Test plan

- [x] `pytest tests/` passes (1679 passed, 15 skipped)
- [x] ruff + mypy pass
- [x] Manual: add _class to a skipped device → device is accepted with entities

### DIFF
--- a/custom_components/ramses_cc/discovery.py
+++ b/custom_components/ramses_cc/discovery.py
@@ -278,6 +278,10 @@ class DiscoveryManager:
         # Populated by sync_with_schema().
         self._schema_no_owner_ids: set[str] = set()
         self._foreign_device_ids: set[str] = set()
+        # Devices in schema with _skipped: True — the user deferred them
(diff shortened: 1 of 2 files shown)


wimpie70:
Superseded by PR 1170 which combines all Phase 1 fixes (skipped re-surface, primary HGI registration, MQTT bind, notification dismissal) into a single PR.

PR #1170, opened 2026-09-06
fix: Phase 1 multi-HGI MQTT pooling — bind, skipped, primary HGI, notifications

wimpie70:
## Summary

Combined Phase 1 fixes for multi-HGI MQTT pooling (issue 1119):

1. **Register primary HGI in schema at startup** — clean-schema startup
   enriches `_owner: me` with the configured HGI (`18:130236`) so the
   pool bridge has an accepted HGI even without a cached device entry.

2. **`_schema_skipped_ids` implementation** — populate the set in
   `sync_with_schema()` and use it to keep skipped devices as NEW
   (not auto-accepted) and reset ACCEPTED→NEW when `_skipped` is added
   (issue 1136).

3. **Bind protocol immediately on pool creation** — the pool bridge
   now calls `protocol.connection_made()` as soon as the
   `PooledTransport` is created, instead of waiting for an HGI to come
   online.  This eliminates the 60s bind timeout on clean-schema
   startup when no HGI is online yet.  HGIs come online later via LWT.

4. **Wait for HA MQTT client before subscribing** — call
   `mqtt.async_wait_for_mqtt_client()` before `async_subscribe()` so
   retained LWT messages are delivered promptly.

5. **Dismiss stale discovery notifications on schema wipe** — when the
   schema is wiped, persistent notifications from before the wipe are
   dismissed so the user doesn't click a notification that leads to an
   empty review form (devices haven't been re-discovered by the scan
   yet).

#### Test plan

- [x] Restart with clean schema → primary HGI registered in <1s
- [x] Second HGI discovered as ownerless candidate, flagged for review
- [x] FAN/REM discovered and accepted after RF packets arrive
- [x] No bind timeout on clean-schema startup
- [x] No stale notification after schema wipe
- [x] `_skipped` devices reappear in review_discovered
- [x] Adding `_class` clears `_skipped`
- [x] Full ha_sim_test on 3 containers

### DIFF
--- a/custom_components/ramses_cc/coordinator.py
+++ b/custom_components/ramses_cc/coordinator.py
@@ -821,6 +821,29 @@ class RamsesCoordinator(DataUpdateCoordinator):
 
         schema = copy.deepcopy(self.entry.options.get(CONF_SCHEMA, {}))
         schema_changed = False
+
+        # Ensure the primary HGI is in the schema with _class: HGI and
+        # _owner: root_owner.  With a clean schema (e.g. after the user
+        # cleared it for testing), the primary HGI won't be in the schema
+        # until sync_learned_topology runs (30-min cycle).  Without this,
+        # enforce_known_list blocks all commands because the known_list
+        # is derived from the schema and the primary HGI is missing.
+        primary_hgi = self._get_primary_hgi_id()
+        if (
+            primary_hgi
+            and primary_hgi.startswith(HGI_PREFIX)
+            and primary_hgi not in schema
+        ):
+            root_owner = schema.get(SZ_OWNER)
+            schema[primary_hgi] = {"_class": "HGI"}
+            if root_owner:
+                schema[primary_hgi][SZ_TR_OWNER] = root_owner
+            schema_changed = True
+            _LOGGER.info(
+                "Registered primary HGI %s in schema (was missing)",
+                primary_hgi,
+            )
+
         for dev_id, entry in schema.items():
             if (
                 dev_id.startswith(HGI_PREFIX)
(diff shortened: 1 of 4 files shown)


wimpie70:
running ha-sim on the latest changes

running for > 50 minutes now on 3 parallel containers... so far only service timeouts due to the parallel test (reloads are heavy on the other tests), no real issues yet

silverailscolo (approved):

wimpie70:
this does need some pr's that didn't make it to current rf version, but it won't break

silverailscolo:
Now would it work with 2 MQTT dongles?

Preparing the release notes, it struck me that in software there is just 1 HGI, but the RF dongle devices are `_class: HGI` and we call the feature "multi-HGI". Better introduce it with the ideal names or it will take off, being "multi-dongle"?, "multi-bridge"??

wimpie70:
On my system it works with 2 MQTT dongels. They are both in the pool and receive both msgs. The HGI with the best RSSI per device is choosen to send a msg towards. All the extra protection, failing connections etc. all works for what I've come up with to test (including the requirements in the plan + more). Single HGI (MQTT, USB still work as well)...wouldn't know why it wouldn't work. But still intro it with caution...it's new...

You can even add another topic and by using a bridge on eg mosquitto you can use HGI's from another location. But that's theory, not tested.

Maybe introduce it as 'multi port'.
Port is actually used in the UI (Primary port, additions ports, etc.) introduced by this plan (sorry, didn't feedback on this, can still be changed in next updates if you want it different). 
Dongle, Gateway, HGI, bridge...all doesn't fit everything. HGI and gateway are used internal with different meanings.

HGI = "Honeywell Gateway Interface" — originally the HGI80 USB dongle hardware
ramses_esp = an ESP32 running firmware that bridges RF to MQTT — it's not a "dongle", it's a WiFi bridge
Both present as 18:xxxxxx devices with _class: HGI
Then we will have Zigbee as well...if we can have it.
Gateway is more accurate, but the pool routes to 1 gateway...

Issue #1171, opened 2026-09-06
0.60.5 Multi-port pool config


silverailscolo:
After updating to 0.60.5 and inspecting the Ramses RF > Config > Manage Gateway Pool pane, I can't remove a USB evofw3 dongle from the list ("uncheck to remove from pool").
When I return to the same pane, it is still blue/selected.
In the console is this:

```
Logger: homeassistant
Bron: runner.py:230
Eerst voorgekomen: 20:54:31 (1 gebeurtenis)
Laatst gelogd: 20:54:31

Error doing job: Task was destroyed but it is pending! (task: <Task pending name='PortProtocol._tx_worker()' coro=<PortProtocol._tx_worker() running at /usr/local/lib/python3.14/site-packages/ramses_tx/protocol/core.py:429> wait_for=<Future pending cb=[Task.task_wakeup()]>>)
```

Note: On the Manage HGI Pool tab, the mqtt usr:pwd is masked for the ports in the Pool Members check box list, but it is not masked at the top (primary port).
Just a small detail

wimpie70:
oops, posted testing results here, on the wrong issue...I'll have a look at this issue soon.

silverailscolo:
@wimpie70 One more strange thing happening:

- my primary HGI (picked at random, assigned to MQTT but actually on USB at the time) shows incomplete MQTT port (correct mqtt://*** but no topic).
- my second HGI is real MQTT ramses_esp dongle. It has the full mqtt:// string with topic in the Pool checkbox list.
- unplug Primary evofw3 from USB, so I'm sure it is Off.
- HGI Status bin_sensor stays OK !! Would expect Problem.
- no errors in system log
- Restart HA
- Still Primary HGI, HGI Status immediately goes to OK. Would expect Problem.

Perhaps this is caused by the boot wait, or cached packets?

wimpie70:
the primary mqtt is not without topic, but you see it as such. And this is a display bug...
What actually happens: it uses a shared topic prefix with wildcard subscriptions to detect all HGI's under that prefix.

Here's how it works:

Topic structure (prefix is RAMSES/GATEWAY by default, configurable via CONF_MQTT_TOPIC):

Purpose	Topic
RX packets	{prefix}/{hgi_id}/rx
TX packets	{prefix}/{hgi_id}/tx
Commands	{prefix}/{hgi_id}/cmd/cmd
Command results	{prefix}/{hgi_id}/cmd/result
[...]

PR #1178, opened 2026-09-07
pool config step improvements + fix bugs — primary HGI removal, credential masking (issue 1171)

wimpie70:
## Summary

Fixes three bugs reported in https://github.com/ramses-rf/ramses_cc/issues/1171:

1. **Primary HGI can't be removed from pool but looks removable** — The primary HGI was shown as a checkbox in the Manage Gateway Pool pane, but unchecking it was silently ignored (the code skips the primary). Now the primary HGI is excluded from the removable checkbox list entirely, so users only see checkboxes for HGIs they can actually remove.

2. **MQTT credentials not masked at top of pane** — The `primary_port` description placeholder showed the raw `mqtt://user:pass@host:port` URL. Now credentials are masked with `***:***@host:port`, matching the checkbox labels.

3. **`Task was destroyed but it is pending`** — `PortProtocol._tx_worker()` task was cancelled but not awaited during `engine.stop()`. Fixed in ramses_rf PR (see below).

#### Test plan
- [x] `tests/tests_new/test_config_flow.py` — 113 passed, 2 skipped
- [x] Full ramses_cc suite — 1748 passed, 15 skipped
- [x] Updated `test_options_flow_manage_pool_remove_schema_member` to match new primary-excluded behavior

### DIFF
--- a/custom_components/ramses_cc/__init__.py
+++ b/custom_components/ramses_cc/__init__.py
@@ -81,8 +81,11 @@ from .const import (
     CONF_MQTT_TOPIC,
     CONF_MQTT_USE_HA,
     CONF_PASSIVE_SCAN,
+    CONF_SCHEMA,
     CONF_SEND_PACKET,
+    DEFAULT_HGI_ID,
     DOMAIN,
+    HGI_PREFIX,
     STORAGE_KEY,
     STORAGE_VERSION,
     SVC_ACCEPT_DISCOVERED_DEVICE,
@@ -94,8 +97,10 @@ from .const import (
     SVC_GET_DISCOVERED_DEVICES,
     SVC_REMOVE_DEVICE,
     SVC_REMOVE_DISCOVERED_DEVICE,
+    SZ_OWNER,
     SZ_PORT_NAME,
     SZ_SERIAL_PORT,
+    SZ_TR_OWNER,
 )
 from .coordinator import RamsesCoordinator
 from .schemas import (
@@ -617,6 +622,24 @@ def _healed_serial_port_options(
         mqtt_hints_present or mqtt_entries_present
     ):
         return None
+
+    # Don't heal if there are no accepted HGIs in the schema — the
+    # user explicitly cleared the pool (issue 1171).  Healing would
+    # undo the clear and re-default to mqtt_ha.
+    schema = options.get(CONF_SCHEMA)
+    if isinstance(schema, dict):
+        root_owner = schema.get(SZ_OWNER)
+        has_accepted_hgi = any(
+            dev_id.startswith(HGI_PREFIX)
+            and isinstance(entry, dict)
+            and entry.get("_class", "").upper() == "HGI"
+            and entry.get(SZ_TR_OWNER) == root_owner
+            and not entry.get("_removed_from_pool")
+            and dev_id != DEFAULT_HGI_ID
+            for dev_id, entry in schema.items()
+        )
+        if not has_accepted_hgi:
+            return None
 
     new_options = {**options}
     new_options[SZ_SERIAL_PORT] = {SZ_PORT_NAME: "mqtt_ha"}
--- a/custom_components/ramses_cc/config_flow.py
+++ b/custom_components/ramses_cc/config_flow.py
@@ -1983,6 +1983,7 @@ class RamsesOptionsFlowHandler(BaseRamsesFlow, OptionsFlow):
         for dev_id, entry in schema.items():
             if (
                 dev_id.startswith(HGI_PREFIX)
+                and dev_id != DEFAULT_HGI_ID
                 and isinstance(entry, dict)
                 and entry.get("_class", "").upper() == "HGI"
                 and entry.get(SZ_TR_OWNER) == root_owner
--- a/custom_components/ramses_cc/config_flow.py
+++ b/custom_components/ramses_cc/config_flow.py
@@ -2045,6 +2045,17 @@ class RamsesOptionsFlowHandler(BaseRamsesFlow, OptionsFlow):
                             )
                     except (ValueError, AttributeError):
                         pass
+                elif display_url == "mqtt_ha" or self.options.get(
+                    CONF_MQTT_USE_HA
+                ):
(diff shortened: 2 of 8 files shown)


silverailscolo:
Do I understand correctly that to remove/replace the primary HGI device (port), I should reconfigure the port (on first tab)?

Difficult for MQTT. Maybe just unplug the primary ramses_esp and restart HA?

wimpie70:
I would like to go to a situation without a primary port and let them be equal in the pool. For mqtt we need a starting point...
But this can be done but is more a phase 2 refactor.
 here's the analysis, I can adapt the plan if you agree:

For MQTT:

- Replace "Choose Serial Port" with a "Broker Config" step (broker URL, topic, credentials)
- Pool pane becomes the only HGI management — add first HGI, add more, remove any
- All HGIs are peers — outbound routing uses RSSI, not "primary"
- No special enrichment — any HGI the user accepts gets _owner: me

For USB:

- No port step at all — USB HGIs are auto-detected from traffic
- User just adds them to the pool when discovered
- Multiple USB ports work the same way

What's blocking in Phase 1:

- SZ_SERIAL_PORT is deeply embedded — transport type, broker URL, HGI ID all live there
- The pool bridge uses the primary URL to build explicit per-HGI MQTT URLs for secondaries
- Config entry version/migration would need updating
- The config flow has 5+ steps built around the primary port concept

The clean path for Phase 2:

1. Split SZ_SERIAL_PORT into separate config: transport_type (mqtt/usb/zigbee) + broker_config (URL, topic, credentials)
2. Pool pane manages all HGIs — no distinction between primary and secondary
3. First accepted HGI becomes the default for outbound until RSSI data arrives
4. Schema ownership is the only membership authority — no "primary" concept

This is the right direction. The "primary" was a Phase 1 shortcut — it let us reuse the existing serial port config flow without redesigning it. For Phase 2, eliminating it would make the UX much cleaner, especially for MQTT where all HGIs are truly peers.

wimpie70:
Also, I want to get some feedback on the new pool to check and fix problems that may be found on different systems...

silverailscolo:
> Also, I want to get some feedback on the new pool to check and fix problems that may be found on different systems...

Sure. We need that, from many users.

1. Rename Gateway Pool tab to **HGI Pool Management.** There is only ever 1 gateway, this is perhaps its pool, but in the pool are HGIs (RF interfaces). Add the Connection/Primary Port to the tab too. (You wouldn't want separate changes to the Port without reviewing the HGI Pool).

2. My 'issue' occurred as the pool randomly picked my FTDI port as Primary, although it doesn't support MQTT at all.
Until a mixed USB/MQTT pool is supported, users need to be able to:
  - turn off the whole pool feature (always on?), or
  - remove _any_ HGI from the pool

3. The Pool tab also suggested adding the USB-connected modbus bridge (mine, not _owned by 'me'!). It only says: "not yet supported". Let's not try.
The Pool checkbox list should skip it, filtering by _owner.

wimpie70:
thanks, yes I agree this is confusing. These should have been gated too on the release.
 I'll create a PR for this.
 I think phase 2 won't take that long to finish. All required tests have been successful, and the connection issues that were blocking have been addressed...

wimpie70:
@silverailscolo 
I'm doing some other adjustements to the HGI Pool Management Step. Maybe you can have a look if this goes into the right direction.
When phase2/3 are fully implemented, this can replace the Connection / Port step (the first), but now it is still needed for serial/zigbee only setup.
I've introduced a _removed_from_pool to keep HGI's in the schema, but excluded from the pool. They can be re-added thru the pool management.

wimpie70:
I'll continue tomorrow testing and juggling esp's around

wimpie70:
## Update — pool config bugs fixed (issue 1171)

All issues from the review and live hardware testing have been addressed on this branch. Full test matrix passes (USB-only, USB+MQTT blocked, 1 MQTT, 2 MQTT + dedup, remove both with confirmation).

### Fixes in this branch

**Pool membership and removal**
- Allow removing primary HGI with auto-promotion of another accepted HGI
- Defer demotion until confirmation; add `_removed_from_pool` trait
- Confirm before clearing the last HGI; clear primary port to start fresh
- Detect last-HGI removal even without `CONF_MQTT_HGI_ID`
- Skip `_removed_from_pool` HGIs in coordinator and `sync_learned_topology`
- Don't flag `_removed_from_pool` HGIs as discovery candidates
- Don't heal `serial_port` back to `mqtt_ha` when no accepted HGIs remain
- Exclude `DEFAULT_HGI_ID` (18:000730) from pool membership

**Config flow UX**
- Split pool add into HA MQTT device ID / full MQTT URL
- List removed HGIs as explicit "re-add" choices
- Hide re-add options when primary is serial (Phase 1 gating)
- Block MQTT pool additions when primary is serial
- Filter `__none__` sentinel from saved `additional_ports`
- Allow adding MQTT HGI when there is no primary (becomes new primary)
- Show primary as a regular pool entry with change/remove option
- Merge Connection/Port and HGI Pool Management into one pane

**Display**
- Mask MQTT credentials in primary port display and all pool member labels
- Show `/RAMSES/GATEWAY` topic for primary even when URL has no path
- Show topic prefix for `mqtt_ha` primary in pool UI
- Fix MQTT URL display and auto-promotion URL construction

**Coordinator / pool bridge**
- Use `_get_primary_hgi_id()` fallback in pool bridge setup
- Handle `mqtt_ha` sentinel in primary detection
- Skip removed HGIs in primary fallback

**CI**
- Fix Ruff F841 (unused `ADD_NEW`) and I001 (import order)
- Add tests for `manage_pool_mqtt_url` error paths, full URL blocked, primary label display, additional port labels
- Coverage for `config_flow.py` now at 95.02% (threshold: 95%)

### Known limitations (Phase 2)

- Gateway status binary sensor tracks the pool bridge's overall `is_active`, not per-HGI connectivity — tracked in Phase 2/3 follow-up
- LWT is the sole source-of-truth for MQTT child availability (no heartbeat expiry yet)

silverailscolo (approved):

Issue #985, opened 2026-08-17
Feat: Control a FAN from HomeKit


silverailscolo:
Adding a FAN entity thru HomeKit Bridge works without HA warnings or errors since 0.59.7, but controlling my FAN from within HomeKit doesn't.

```
Logger: ramses_rf.pipeline.conversation
Bron: runner.py:290
Eerst voorgekomen: 16 augustus 2026 om 22:32:01 (7 gebeurtenissen)
Laatst gelogd: 16 augustus 2026 om 22:37:08

Conversation for 32:139370:22F1:12f822e2-0a71-4840-b88f-343934eeab4e timed out after 3 retries.
Conversation for 32:139370:22F1:e4e782a9-4115-45b9-8d15-369e8c49dd30 timed out after 3 retries.
Conversation for 32:139370:22F1:cce52eef-88a9-41f4-8229-bd1b51bf69ad timed out after 3 retries.
Conversation for 32:139370:22F1:61df44d4-d6eb-4822-a811-b57fe8087eb3 timed out after 3 retries.
Conversation for 32:139370:22F1:e493a325-e754-45dc-9716-a2805b05a00c timed out after 3 retries.
```

```
Logger: homeassistant.components.homekit.accessories
Bron: components/homekit/accessories.py:787
Integratie: HomeKit Bridge (documentatie, problemen)
Eerst voorgekomen: 16 augustus 2026 om 22:32:01 (7 gebeurtenissen)
Laatst gelogd: 16 augustus 2026 om 22:37:08

climate.32_139370: climate.set_fan_mode failed (Failed to set fan mode: Timeout waiting for reply to set_fan_mode); re-syncing HomeKit state
climate.32_139370: climate.set_hvac_mode failed (Failed to set fan mode: Timeout waiting for reply to set_fan_mode); re-syncing HomeKit state
```

_Originally posted by @AbeltjeNL in https://github.com/ramses-rf/ramses_cc/issues/500#issuecomment-5309600754_

wimpie70:
Investigation results — the timeout reported here has the same root cause as issue 995 (closed, fixed in ramses_rf 0.59.8), and there is a separate secondary issue worth noting.

## 1. The timeout (`Conversation for … timed out after 3 retries`)

[...]

PR #1202, opened 2026-09-14
fix(climate): raise on unrecognized command format instead of silent fall-through

wimpie70:
## Summary
- Fix the secondary bug identified in issue #985: when a `fan_mode` is found in the FAN's `_commands` but its value is neither a dict template (`{verb, code, payload}`) nor a raw packet string, the code silently set `packet_str=None` and fell through to the ramses_rf native `set_fan_mode`.
- That fallback rejects custom command names (e.g. Dutch aliases like `laag`/`hoog`) with a confusing `fan_mode is not valid for scheme 'orcon': laag` error.
- Now raise a `ValueError` (wrapped as `HomeAssistantError` by the existing `except Exception` handler) with a clear message about the unrecognized command format.

## Context
Issue 985 has two parts:
1. **The timeout** (`Conversation for … timed out after 3 retries`) — already fixed in ramses_rf commit `8839ceb` (`wait_for_reply=False`, fire-and-forget). Shipped in ramses_rf 0.59.8+; ramses_cc 0.60.6 pins `ramses-rf==0.60.6` which includes the fix.
2. **Silent fall-through on unrecognized command format** — this PR.

#### Test plan
- [x] New test `test_hvac_set_fan_mode_unrecognized_format_raises` verifies the error is raised, the command is NOT sent, and the ramses_rf native `set_fan_mode` is NOT called
- [x] All 118 climate tests pass
- [x] `ruff check` — clean
- [x] `mypy` — clean
- [x] pre-commit hooks pass


wimpie70:
cleaning up some issues...like https://github.com/ramses-rf/ramses_cc/issues/985

silverailscolo (approved):

Issue #1113, opened 2026-08-31
feature: add HVAC timer mode commands


wimpie70:
We are missing the High15 etc ~mode~ boost_timer commands on the strategies and downstream support

wimpie70:
## Analysis: timer commands should be `boost_timer`, not `mode`

Looking at the Orcon schema and the existing strategy infrastructure, the `High15`-style commands are **22F3** (ventilation control / timed boost), not 22F1 (persistent fan mode). They should be classified as `boost_timer`, which is a `type` slot `HvacStrategyBase` already reserves.

### Decoding the Orcon 22F3 timer commands

All four are 7-byte 22F3 payloads, verb `I` (spoofed from the bound REM, same as the 22F1 mode entries):

| Command | Payload | Header | Flags | Mins | fan_mode_byte | fb_mode | fb_fan_mode | trailing |
|---|---|---|---|---|---|---|---|---|
| `high_15` | `00 12 0F 03 04 04 04` | 00 | 0x12 | 0x0F=**15** | 03=**high** | 04 | 04 | 04 |
| `high_30` | `00 12 1E 03 04 04 04` | 00 | 0x12 | 0x1E=**30** | 03=high | 04 | 04 | 04 |
| `high_60` | `00 12 3C 03 04 04 04` | 00 | 0x12 | 0x3C=**60** | 03=high | 04 | 04 | 04 |
| `med_60`  | `00 12 3C 02 04 04 04` | 00 | 0x12 | 0x3C=60 | 02=**medium** | 04 | 04 | 04 |

The `fan_mode_byte` map (from `to_dict` in `ramses_rf/payloads/hvac.py`) is `{0: off/away, 1: low, 2: medium, 3: high, 4: high/away}` — so `03` = high, `02` = medium. Matches the labels.

The naming `high_15` / `med_60` is **mode + duration** — the canonical 22F3 "boost_timer" semantic. These are unambiguously **timers, not modes**.

### Why `boost_timer`, not `mode`

The two opcodes are fundamentally different:

- **22F1** = persistent fan *mode* (`away`, `low`, `medium`, `high`, `boost`, `auto`...). State change, no duration. Maps cleanly to HA `fan_mode`.
- **22F3** = ventilation control / *timed boost*. Payload carries `minutes`, `fan_mode_byte`, `fallback_mode_byte`, `fallback_fan_mode_byte` — i.e. "do mode X for N mins, then fall back to mode Y".

[...]

PR #1203, opened 2026-09-14
feat(remote): send_command uses strategy builtin_commands (issue 1113)

wimpie70:
## Summary

Implements https://github.com/ramses-rf/ramses_cc/issues/1113 — makes strategy-provided timed boost commands (`22F3`) sendable on FAN entities **without schema `_commands` entries**.

Timed boost commands are temporary actions (fan runs at a speed for N minutes, then reverts) — `send_command` is the right abstraction, not `preset_mode`. This supersedes the earlier preset_mode draft on this branch.

**Precedence in `async_send_command`:**

```
schema _commands  →  strategy builtin_commands  →  set_fan_mode (modes + aliases)
```

- `RemoteEntity._lookup_builtin_command()` resolves the FAN strategy's `builtin_commands` dict templates, resolving `_boost_aliases` (`laag_15` → `low_15`, `hoog_30` → `high_30`, ...) — aliases always accepted, same rule as `fan_mode_to_hex`.
- Builtin commands send via the existing `_build_packet_from_template` path (src = bound REM / HGI, dst = FAN).
- Builtins are **never persisted** to schema `_commands`.
- `extra_state_attributes`: builtin commands merged into the displayed `commands` attribute for FAN entities — schema entries win on name conflict.
- `async_delete_command`: the strategy-provided warning now also covers builtin names and boost aliases.

Works through both `remote.send_command` and the `ramses_cc.send_command` entity service.

## Depends on

- https://github.com/ramses-rf/ramses_rf/pull/1219 — strategy `_builtin_commands` + `_boost_aliases` (Orcon, Vasco, ClimaRad, Itho)
- Backward-compatible: if the installed ramses_rf has no builtin commands, nothing changes.

#### Test plan

- [x] 8 new tests in `test_remote.py`: Orcon/Vasco payload, Itho 3-byte, Dutch alias resolution, schema override precedence, no schema persistence, `set_fan_mode` fallback intact, attribute merge, delete warning
- [x] 1855 tests pass; ruff / mypy / pre-commit clean
- [ ] End-to-end on ha-sim once ramses-rf 1219 lands


wimpie70:
Here's my hesitation on this PR and why I needed to check this a bit more: Now the Timed Boost commands can be set by the set_preset_mode, but I think my LLM made a wrong turn there...(The one I used before isn't free anymore, so I am checking some others...)
They should just be used by the send_command actions instead...

silverailscolo (approved):

PR #1209, opened 2026-09-18
fix(mqtt): stop sentinel HGI 18:000730 leaking into pool and schema

wimpie70:
## Summary

`18:000730` (`HGI_DEV_ADDR`) is ramses_rf's internal placeholder source address — not a physical gateway. Several unguarded paths allowed it to become a configured, accepted, or discovered MQTT pool member. Once it became a pool child, outbound frames were published to `GATEWAY/18:000730/tx` — a topic no real gateway subscribes to, so commands routed there vanished (observed in a user MQTT dump on the HA forum).

The same gaps allowed non-HGI device IDs (e.g. a `01:` controller) into the **Gateway Device ID** option, which silently created a dead pool child.

### Changes

- **coordinator**: exclude the sentinel from `_get_primary_hgi_id()` schema fallbacks, `_get_accepted_hgi_ids()`, the `pool_hgi_ids` → scan/schema writers, and `mqtt://` URL-derived HGI lists
- **discovery**: `_MqttHgiDiscoveryCallback.on_unknown_hgi()` ignores sentinel topics — they can no longer create schema candidates
- **mqtt_pool_bridge**: strip the sentinel from `configured_hgi_ids` / `accepted_hgi_ids` at init and ignore RX/LWT on its topic — defense in depth so a sentinel child can never exist regardless of caller
- **config_flow**: validate **Gateway Device ID** against `_HGI_ID_RE` (`mqtt_hgi_id_invalid`), never inject the sentinel into the schema, and reject it in the `manage_pool_mqtt` / `manage_pool_mqtt_url` steps (it passed `_HGI_ID_RE` before and got `_owner` set)

The sentinel default in the Gateway Device ID field is still allowed (it is the "unset" value) — it is just never stored into the schema as an HGI. A sentinel already persisted in a user schema is now inert rather than removed.

#### Test plan

- [x] `pytest tests/tests_new` — 1857 passed (11 new regression tests covering every filtered path)
- [x] `prek run -a` — ruff, mypy, codespell all clean
- [x] ha_sim_test R105/R111/R112/R119 — 43/43 pass against the branch in ha-sim
- [x] New R119 section verifies e2e: retained `GATEWAY_SIM/18:000730` online + rx → **absent** from persisted schema; control `18:009999` → present as ownerless candidate. Fails on unpatched code, passes on this branch

Companion test change (ramses_extras R119 recipe) is on the open ramses_extras PR.

Refs issue 1171.


silverailscolo:
Good catch. I received a sentinel HGI once, and then got confused which _extra_ physical device got into the pool...

wimpie70:
We did have some guards, but not enough. This should fix part of Chris his problems on https://community.home-assistant.io/t/honeywell-ch-dhw-via-rf-evohome-sundial-hometronics-chronotherm/151584/5152?

silverailscolo (approved):

PR #1220, opened 2026-09-21
Migrate HVAC strategy lookups to public ramses_rf accessor

### LINKED ISSUE #1137: Expose a public strategy accessor on DeviceBase
Future improvement: 
(found when reviewing https://github.com/ramses-rf/ramses_rf/pull/1187?notification_referrer_id=NT_kwHODOAaC9oAJVJlcG9zaXRvcnk7MjIzOTEyNzk5O0lzc3VlOzUzMjU1NjQyMDc#issuecomment-5513646232)

_resolve_fan_strategy does:

```python
fan._strategy
fan._scheme
```

Both are private. Other call sites in the codebase already go through _get_configured_strategy() (which is also private, but at least encapsulates the logic). Reaching into two separate private attributes from a different device class is a smell.

HvacCarbonDioxideSensor._resolve_fan_strategy accesses fan._strategy and fan._scheme directly. Add a public fan.strategy property (or extend _get_configured_strategy() to accept a device argument) so cross-device strategy resolution doesn't break encapsulation

wimpie70:
## Summary

Fixes https://github.com/ramses-rf/ramses_cc/issues/1137 — migrates `ramses_cc` off private `ramses_rf` strategy internals now that `DeviceBase` exposes public accessors (https://github.com/ramses-rf/ramses_rf/pull/1228).

- **`helpers.py`**: new `configured_hvac_strategy()` — prefers public `get_configured_strategy()`, falls back to legacy `_get_configured_strategy()` for older ramses_rf releases; isinstance-guards the result
- **`helpers.py`**: new `strategy_mode_aliases()`/`strategy_boost_aliases()` — prefer the public `aliases`/`boost_aliases` properties, fall back to `_aliases`/`_boost_aliases` on older ramses_rf
- **`remote.py`**: three `_get_configured_strategy` getattr sites → helper; `_aliases`/`_boost_aliases` reads → alias helpers
- **`climate.py`**: deletes `_get_device_strategy()` and its private `_STRATEGY_BY_SCHEME` import — `_scheme` lookup now happens inside ramses_rf, and explicit `set_strategy()` overrides are honoured (previously ignored by the fan-mode path); `_aliases` reads → `strategy_mode_aliases()`
- **model-awareness**: since `get_configured_strategy()` now defaults `model` to the device's cached 10E0 model, the sync `fan_modes` property picks up `OrconHrc350Strategy` automatically once a 10E0 has been seen — no separate ingestion-time caching needed
- **tests**: mocks target `get_configured_strategy` instead of `_scheme` / `_get_configured_strategy`

**Dependency note:** uses `getattr` fallback, so it works with both old and new ramses_rf.

#### Test plan
- [x] `pytest tests/` — 1996 passed, 15 skipped (against ramses_rf feature branch installed editable)
- [x] `ruff check` + `prek run -a` (incl. mypy custom_components) — clean
- [x] ha_sim_test on ha-sim (new rf code via bind mount + new cc code): R18 R43 R65-68 R79 R90 R91 R93-97 R101 R125 — all pass, no unexpected log errors
- [x] hass dev instance restart with new cc code — clean startup, no errors


silverailscolo:
CI didn't fail, despite the upstream PR not yet accessible.

Safeguards in place? In that case you might adjust your last comment above.

custom_components/ramses_cc/helpers.py
```diff
@@ -444,3 +445,51 @@ def dto_to_dict(val: Any) -> Any:
...
+    ``_get_configured_strategy()`` so older ramses_rf releases keep
+    working until the manifest pin is bumped.
+
+    :param device: The ramses_rf device (e.g., an HvacVentilator).
+    :return: The configured strategy object, or None.
+    """
+    getter = getattr(device, "get_configured_strategy", None) or getattr(
+        device, "_get_configured_strategy", None
+    )
+    if not callable(getter):
```
silverailscolo, line 464: Safeguards for older _rf libs

silverailscolo (approved):

wimpie70:
yes, it's contradicting...fallback and ci fails

we should clear those fallbacks on a proper cleanup PR some day. There is more old code and there are TODO's

Issue #1216, opened 2026-09-20
Feature: last_message / last_mode_sent entities for REM/DIS (and reconsider fan_rate)


wimpie70:
Follow-up from an investigation into why REM/DIS devices expose `sensor.*_fan_mode` / `sensor.*_fan_rate`.

## Background

`HvacRemote` (REM) / `HvacDisplayRemote` (DIS) expose `fan_mode`/`fan_rate`, so ramses_cc creates these sensors for any such device (`hasattr(device, 'fan_mode')`; `sensor.py` `SZ_FAN_MODE`/`SZ_FAN_RATE` descriptions). Because the state projector treats the source device as a hardware twin, a REM's `fan_mode` entity effectively shows the last 22F1-mode command the remote itself transmitted — a "last mode sent", not a ventilator state.

## Problems

- `fan_mode` only reflects mode-value packets (22F1/22F4/31D9 semantics). Other commands the device sends — e.g. a 22F3 timed boost, or 2411/10D0 parameter get/set — are invisible.
- A generic "last command" sensor has a catch: param chatter (2411 etc.) would overwrite and hide the last mode command. So both are useful: `last_command` (whatever was sent last) and `last_mode_sent` (last mode command).
- `fan_rate` on REM/DIS is meaningless — remotes have no ventilator and only mirror a byte from their own command. Consider restricting the `SZ_FAN_RATE` sensor description to `ramses_rf_class=HvacVentilator`. This is a breaking change for existing installs (orphaned entities), so stage it: `entity_registry_enabled_default=False` first, or a registry cleanup.

## Proposal

- `sensor.<dev>_last_command`: TIMESTAMP device class, state = `dtm` of the last message sourced by the device; attributes = verb, code, decoded payload, dst. Data is already available via `device.entity_state` (StateHeader → Message map); likely implementable in ramses_cc only.
- Keep/reframe `fan_mode` on REM/DIS as the "last mode sent" — possibly rename the entity or clarify in docs that it tracks the remote's own transmitted mode.
- Restrict `fan_rate` to FAN-class devices.

## Notes

- `binary_sensor.<dev>_status` already exposes `last_seen` + staleness/RSSI attributes; `last_command` complements it with the message content.
- Related wrinkle found during the investigation: devices that send both 1298 and 22F1 (CO2 remotes with buttons) can be eavesdrop-classed as either CO2 or REM depending on first-seen packet, leaving orphaned `rem_*` entities when the class flips.

[...]

PR #1227, opened 2026-09-21
feat(sensor): last_command sensor + fan_rate/fan_mode fixes for REM/DIS

wimpie70:
## Summary

Implements https://github.com/ramses-rf/ramses_cc/issues/1216 (all three proposals):

- **`sensor.<dev>_last_message_sent`** — diagnostic sensor. State = the decoded payload dict of the last message *sourced* by the device, rendered as a string (truncated to the 255-char state limit); attributes = `sent` (ISO timestamp), `verb`, `code`, `dst`. Unlike trait sensors it reflects *every* verb, so messages invisible to `fan_mode` (e.g. 22F3 timed boost, 2411/10D0 param get/set) now show up.
  - **Disabled by default on HGIs**: an HGI transmits on every poll cycle, so its sensor would churn the recorder DB on flash-constrained installs (`entity_registry_enabled_default=False` for `HgiGateway`; users can enable per entity). Device-class sensors stay enabled. HA offers no integration-side recorder-exclude flag — the registry toggle is the native per-entity off-switch; activity/history follow the same recorder data either way.
- **`fan_rate` restricted to `HvacVentilator`** — a REM/DIS remote has no ventilator; its `fan_rate` only mirrored a byte from its own last command (the modbus `requested_rate` per silverailscolo's comment), which can differ from the fan's actual rate when e.g. a CO2 sensor demands more. FAN is the single source of truth for actual rate.
- **`fan_mode` on REM/DIS renamed to "Last mode sent"** — same key/unique_id (`<id>-fan_mode`), so existing entities keep their entity_id; only the display name changes. On FAN it stays "Fan mode". Both descriptions expose a `last_sent` attribute (the device's `last_fan_mode_dtm`), skipped on ramses_rf versions without it.

Depends on https://github.com/ramses-rf/ramses_rf/pull/1230 (public `device.last_msg` and `device.last_fan_mode_dtm`, plus the scheme-aware 22F1 mode-name fix for Orcon remotes that send `mode_max=04`). On older ramses_rf the `hasattr` gates simply skip the entities/attributes — no errors.

#### Test plan
- [x] 9 sensor tests: description wiring, class restrictions, payload-string state, `sent`/`last_sent` attrs, naive-dtm handling, no-message and None-payload cases, 255-char truncation, HGI disabled-by-default
- [x] `tests_new` suite: 1901 passed, 15 skipped; snapshots regenerated (HGI entities absent — disabled)
- [x] Coverage gate: 100% of modules >=95%
- [x] `ruff` / `mypy` / `prek -a`: clean
- [x] Snapshot flake handled: `last_msg` state masked + `sent`/`last_sent` attrs excluded (synthesized 0010 echo embeds a live timestamp)
- [x] Verified live on `hass`: `sensor.rem_*_last_message_sent` shows `I/22F1` payloads from real remote presses


wimpie70:
you may want to re-review this

silverailscolo (approved):

